PR #5, opened 2023-09-15
superchain: contract implementations

tynes:
**Description**

Adds a concept of the contract implementations. yaml files that include the semantic version to address exist for each network as well as a global file for implementations that are deployed with create2. The network specific config is merged into the global config, so that contract addresses specific to a network will override the globals. This will be the canonical list of contract addresses for implementations going forward and will be used by the bundler tool to build upgrade transactions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,8 @@ The superchain configs are made available in minimal form, to embed in OP-Stack 
 Full deployment artifacts and genesis-states can be derived from the minimal form
 using the reference [`op-chain-ops`] tooling.
 
+The `semver.yaml` file represents the semantic versioning lockfile for the smart contracts in the superchain.
+
 ## Go Module
 
 Superchain configs can be imported as Go-module:
--- a/README.md
+++ b/README.md
@@ -11,7 +11,8 @@ The superchain configs are made available in minimal form, to embed in OP-Stack 
 Full deployment artifacts and genesis-states can be derived from the minimal form
 using the reference [`op-chain-ops`] tooling.
 
-The `semver.yaml` file represents the semantic versioning lockfile for the smart contracts in the superchain.
+The `semver.yaml` file represents the semantic versioning lockfile for the all of the smart contracts in the superchain.
+It is meant to be used when building transactions that upgrade the implementations set in the proxies.
 
 ## Go Module
 
--- a/superchain/implementations/networks/mainnet.yaml
+++ b/superchain/implementations/networks/mainnet.yaml
@@ -1,0 +1,7 @@
+l1_cross_domain_messenger:
+l1_erc721_bridge:
+l1_standard_bridge:
+l2_output_oracle:
+optimism_mintable_erc20_factory:
+optimism_portal:
+system_config:
--- a/superchain/superchain.go
+++ b/superchain/superchain.go
@@ -59,6 +59,7 @@ type ChainConfig struct {
 	Chain string `yaml:"-"`
 }
 
+// AddressList represents the set of network specific contracts for a given network.
 type AddressList struct {
 	AddressManager                    Address `json:"AddressManager"`
 	L1CrossDomainMessengerProxy       Address `json:"L1CrossDomainMessengerProxy"`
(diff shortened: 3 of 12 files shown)


superchain/superchain.go
```diff
@@ -129,7 +211,23 @@ var Addresses = map[uint64]*AddressList{}
 
 var GenesisSystemConfigs = map[uint64]*GenesisSystemConfig{}
 
+var Implementations = map[uint64]ContractImplementations{}
+
+// networkToChainID maps a network name to a chain ID. Only the superchain targets
+// are required to be in this mapping
+var networkToChainID = map[string]uint64{
```
protolambda, line 218: The L1 chain ID is already included in the superchain targets info (see `Superchains` global map)

superchain/implementations/networks/mainnet.yaml
```diff
@@ -0,0 +1,7 @@
+l1_cross_domain_messenger:
+l1_erc721_bridge:
+l1_standard_bridge:
+l2_output_oracle:
+optimism_mintable_erc20_factory:
+optimism_portal:
+system_config:
```
protolambda, line 7: Should these be set to the older currently deployed addresses?

superchain/superchain.go
```diff
@@ -61,6 +70,121 @@ type AddressList struct {
...
+			return fmt.Errorf("invalid semver %s for field %s", str, val.Type().Field(i).Name)
+		}
+	}
+	return nil
+}
+
+// NewContractImplementations returns a new empty ContractImplementations.
+// Use this constructor to ensure that none of struct fields are nil.
+// The path argument is relative to the embedded fs that contains the implementations.
+func NewContractImplementations(path string) (ContractImplementations, error) {
```
protolambda, line 126: Can we make this a private function, and hide the path from the public interface, and instead just use the superchain name? The internal embedded FS with configuration data should be an implementation detail, something which we may want to change later.

superchain/superchain.go
```diff
@@ -129,7 +253,26 @@ var Addresses = map[uint64]*AddressList{}
 
 var GenesisSystemConfigs = map[uint64]*GenesisSystemConfig{}
 
+// Implementations represents a global mapping of contract implementations
+// to chain by chain id.
+var Implementations = map[uint64]ContractImplementations{}
+
+// SuperchainSemver represents a global mapping of contract name to desired semver version.
+var SuperchainSemver = ContractVersions{}
```
protolambda, line 261:
```suggestion
var SuperchainSemver ContractVersions
```
nit: it's a struct, not a map, it's fine to leave it as variable declaration

README.md
```diff
@@ -11,6 +11,8 @@ The superchain configs are made available in minimal form, to embed in OP-Stack
 Full deployment artifacts and genesis-states can be derived from the minimal form
 using the reference [`op-chain-ops`] tooling.
 
+The `semver.yaml` file represents the semantic versioning lockfile for the smart contracts in the superchain.
```
protolambda, line 14: Would reword this, since it's a single file, for all superchain targets at the same time.

superchain/implementations/networks/mainnet.yaml
tynes: I left them empty because the current addresses are technically not superchain compatible. Open to adding them, ideally we can have tooling that can grab from storage each of the impl addresses and generate these files given the proxy addresses

superchain/superchain.go
tynes: Sure this makes sense, good call at not leaking that abstraction

superchain/superchain.go
```diff
@@ -61,6 +71,251 @@ type AddressList struct {
...
+func (a AddressSet) Get(key string) Address {
+	if strings.HasPrefix(key, "v") {
+		if addr, ok := a[key]; ok {
+			return addr
+		}
+		if addr, ok := a[strings.TrimPrefix(key, "v")]; ok {
+			return addr
+		}
+	}
+	return a[key]
```
refcell, line 121: [`strings.TrimPrefix`](https://pkg.go.dev/strings#TrimPrefix) "returns s without the provided leading prefix string. If s doesn't start with prefix, s is returned unchanged."

So I think we can do the following.
```suggestion
	if !strings.HasPrefix(key, "v") {
		key = "v" + key
	}
	if addr, ok := a[strings.TrimPrefix(key, "v")]; ok {
		return addr
	}
	return a[key]
```

superchain/superchain.go
```diff
@@ -61,6 +71,251 @@ type AddressList struct {
...
+	}
+	return implementations, nil
+}
+
+// resolve returns a VersionedContract that matches the passed in semver version
+// given a set of addresses.
+func resolve(set AddressSet, version string) (VersionedContract, error) {
+	if !strings.HasPrefix(version, "v") {
+		version = "v" + version
+	}
```
refcell, line 171: If we define a semver type we could create a `Canonicalize()` method on it that automagically adds the prefix.

refcell: (It would allow us to have this check be one lined)
```golang
version = version.Canonicalize()
```

refcell: And if we really want a string param, can do
```golang
semver = Semver(version).Canonicalize()
```

refcell (approved):
Few small nits, otherwise LGTM

superchain/superchain.go
```diff
@@ -18,6 +21,12 @@ var superchainFS embed.FS
 //go:embed extra/addresses extra/bytecodes extra/genesis extra/genesis-system-configs
 var extraFS embed.FS
 
+//go:embed implementations/networks implementations
+var implementationsFS embed.FS
```
refcell, line 25: Also, go embeds files in directories [recursively](https://pkg.go.dev/embed), so this can just be `implementations`.
Though, note it will exclude dotfiles and `_` prefixed files. If we want to include this, we can use the `all:` directive prefix.
```suggestion
//go:embed implementations
var implementationsFS embed.FS
```

superchain/superchain.go
```diff
@@ -18,6 +21,12 @@ var superchainFS embed.FS
 //go:embed extra/addresses extra/bytecodes extra/genesis extra/genesis-system-configs
 var extraFS embed.FS
```
refcell, line 22: I think we can also update this since go embeds subtree files recursively in directories (see below).
```suggestion
//go:embed extra
var extraFS embed.FS
```

tynes: I wanted to skip the `README.md` in this case

tynes:
Thank you for the review @refcell, I have addressed the comments

superchain/superchain.go
refcell: Ah that makes sense, good call

PR #20, opened 2023-10-05
goerli-dev: update internal devnet addresses

tynes:
**Description**

Makes the addresses match the deployments in `internal-devnet`
https://github.com/ethereum-optimism/optimism/tree/develop/packages/contracts-bedrock/deployments/internal-devnet


protolambda (changes requested):
The actual latest devnet does not match the deployments in the monorepo, I believe this change may be wrong, since the addresses this changes it to match that of the monorepo deployments dir. See rollup-config output of devnet op-node RPC: the system config and deposit-contract address do not match the monorepo deployments.

tynes:
@protolambda I reverted the changes to the addresses but we still want the change that adds the new version of the system config

PR #23, opened 2023-10-18
Set mainnet and sepolia superchain-target ProtocolVersions contract address

protolambda:
**Description**

Update Mainnet and Sepolia superchain-target configs with protocol-versions contract addresses.


tynes (approved):
Looks good to me, confirmed both addresses are correct

PR #28, opened 2023-10-26
Add Canyon time & activate Canyon on devnets

trianglesphere:
**Description**

This adds the canyon time to the chain config & activates it on devnets.
Devnet activation time is 1698436800 which is Fri Oct 27 13:00:00 PDT 2023.

**Tests**

Manually tested that this was getting pulled into op-geth and op-node with a local go mod replace.

**Additional Context**

I do want to switch over to defining the canyon time in the superchain file, but it was easier to define it in each chain file for the moment. For consumers, there is no difference.

### DIFF
--- a/superchain/configs/goerli-dev-0/base-devnet-0.yaml
+++ b/superchain/configs/goerli-dev-0/base-devnet-0.yaml
@@ -7,6 +7,8 @@ explorer: ""
 system_config_addr: "0x4f775d578e3ab8ce81f5ec065050938dbd5fb8c2"
 batch_inbox_addr: "0x1176307100000000000000000000000000000000"
 
+canyon_time: 1698436800
+
 genesis:
   l1:
     hash: "0x02578eb559ac846feaeb0a8fdf1f19de9965fc59e1a38ebba9ff93ed4e21c99a"
--- a/superchain/configs/goerli-dev-0/base-devnet-0.yaml
+++ b/superchain/configs/goerli-dev-0/base-devnet-0.yaml
@@ -7,8 +7,6 @@ explorer: ""
 system_config_addr: "0x4f775d578e3ab8ce81f5ec065050938dbd5fb8c2"
 batch_inbox_addr: "0x1176307100000000000000000000000000000000"
 
-canyon_time: 1698436800
-
 genesis:
   l1:
     hash: "0x02578eb559ac846feaeb0a8fdf1f19de9965fc59e1a38ebba9ff93ed4e21c99a"
--- a/superchain/superchain_test.go
+++ b/superchain/superchain_test.go
@@ -285,14 +285,19 @@ func TestContractBytecodes(t *testing.T) {
 // TestCanyonTimestampOnBlockBoundary asserts that Canyon will activate on a block's timestamp.
 // This is critical because the create2Deployer only activates on a block's timestamp.
 func TestCanyonTimestampOnBlockBoundary(t *testing.T) {
-	for id, config := range OPChains {
-		if config.CanyonTime == nil {
-			continue // skip if canyon is not set
+	for superchainName, superchainConfig := range Superchains {
+		if superchainConfig.Config.CanyonTime == nil {
+			continue
 		}
-		canyonOffset := *config.CanyonTime - config.Genesis.L2Time
-		if canyonOffset%2 != 0 {
-			t.Fatalf("Canyon time for %v is not on the block time. canyon time: %v. L2 start time: %v, block time: %v",
-				id, *config.CanyonTime, config.Genesis.L2Time, 2)
+		ct := *superchainConfig.Config.CanyonTime
+		for _, id := range superchainConfig.ChainIDs {
+			chainCfg := OPChains[id]
+			canyonOffset := ct - chainCfg.Genesis.L2Time
+			// Block time is hardcoded in op-node/rollup/superchain.go
+			if canyonOffset%2 != 0 {
+				t.Fatalf("Canyon time on superchain %v for %v is not on the block time. canyon time: %v. L2 start time: %v, block time: %v",
+					superchainName, id, ct, chainCfg.Genesis.L2Time, 2)
+			}
 		}
 	}
 }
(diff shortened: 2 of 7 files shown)


tynes:
Can we add a test that shows the timestamp is divisible by the blocktime, since activation is based on the exact timestamp being hit?

tynes:
interesting that there is no regolith time, which it should be 0 anyways

superchain/configs/goerli-dev-0/base-devnet-0.yaml
```diff
@@ -7,6 +7,8 @@ explorer: ""
 system_config_addr: "0x4f775d578e3ab8ce81f5ec065050938dbd5fb8c2"
 batch_inbox_addr: "0x1176307100000000000000000000000000000000"
 
+canyon_time: 1698436800
```
protolambda, line 10: Why do we have to repeat the exact same timestamp in 4 configs of the same superchain target? Can we not make the configuration part of the `superchain.yaml`? (We can add an override option to the chain, but ideally we keep superchain member chain on the exact same hardfork schedule to avoid confusing inconsistencies)

protolambda (changes requested):

superchain/configs/goerli-dev-0/base-devnet-0.yaml
protolambda: In op-node and op-geth the superchain-config is readily available when constructing the resulting rollup config and EL chain-config, so the superchain attribute can be pulled in that way, without duplication.

trianglesphere:
> interesting that there is no regolith time, which it should be 0 anyways

Regolith set inside op-node/op-geth on a per-chain ID basis.

https://github.com/ethereum-optimism/optimism/blob/5a80dc914ba5b7723457e49dad6b22c2de9fb51c/op-node/rollup/superchain.go#L63-L75

https://github.com/ethereum-optimism/op-geth/blob/dda1a26683a8b297f534b57ff22a4fd8b98d217a/params/superchain.go#L88-L114

superchain/configs/goerli-dev-0/base-devnet-0.yaml
trianglesphere: Specifically addressed in the PR description.

> I do want to switch over to defining the canyon time in the superchain file, but it was easier to define it in each chain file for the moment. For consumers, there is no difference.

sebastianst: Why was it easier? I'm curious, I thought that by default we want to add activation times into the `superchain.yaml`.

superchain/superchain_test.go
```diff
@@ -281,3 +281,18 @@ func TestContractBytecodes(t *testing.T) {
...
+
+// TestCanyonTimestampOnBlockBoundary asserts that Canyon will activate on a block's timestamp.
+// This is critical because the create2Deployer only activates on a block's timestamp.
+func TestCanyonTimestampOnBlockBoundary(t *testing.T) {
+	for id, config := range OPChains {
+		if config.CanyonTime == nil {
+			continue // skip if canyon is not set
+		}
+		canyonOffset := *config.CanyonTime - config.Genesis.L2Time
+		if canyonOffset%2 != 0 {
```
sebastianst, line 293: nit: can we add a `const BlockTime` to the `superchain` package and use it instead of magic `2`s?

superchain/configs/goerli-dev-0/base-devnet-0.yaml
protolambda: I don't see why we cannot just change it to the superchain config already, I disagree with the PR description on what's easier there.

trianglesphere: I went with this approach b/c I tend to think of HF as per chain (i.e. look up a HF by chain ID). Looking more closely, I see that when we actually pull the chain config per chain, we also pull the SuperchainConfig. I could load the value directly from that. I was initially thinking that I'd have to set the value in the per-chain config from the superchain config at init time.

In general, there's no documentation on what structs are important/what they do which made it hard to find how to do it with the supchain-config

superchain/superchain_test.go
trianglesphere: would be nice. We also hardcode it in op-node which is why I went with the hardcode here (b/c there was nowhere to import it from)

tynes: We should update the config in `op-node` so that if blocktime is `nil` it defaults to a constant value (`2s`), then this constant can be imported everywhere. This reduces required config and can be done for the other config values that have superchain required defaults

tynes:
Generally looks good to me

superchain/superchain_test.go
sebastianst: upstream const that can be imported everywhere would definitely be even better! but I thought keeping this change localized to the `superchain-registry` repo and add it locally would still be better than continue hardcoding consts.

superchain/superchain_test.go
```diff
@@ -281,3 +281,23 @@ func TestContractBytecodes(t *testing.T) {
...
+	for superchainName, superchainConfig := range Superchains {
+		if superchainConfig.Config.CanyonTime == nil {
+			continue
+		}
+		ct := *superchainConfig.Config.CanyonTime
+		for _, id := range superchainConfig.ChainIDs {
+			chainCfg := OPChains[id]
+			canyonOffset := ct - chainCfg.Genesis.L2Time
+			// Block time is hardcoded in op-node/rollup/superchain.go
+			if canyonOffset%2 != 0 {
```
protolambda, line 297: Not every chain genesis time may align perfectly with the hardfork time of every other chain in the same superchain target (although, if every chain has a standard genesis based on L1 time starting block time, it might, because the L1 time alignment).

protolambda (approved):

superchain/superchain_test.go
trianglesphere: That's true and a problem for canyon. Hopefully it works for canyon b/c everyone chose a even start time. Maybe we want to be more explicit about this in the future.

PR #29, opened 2023-10-30
Activate Canyon on Sepolia & Goerli testnets

trianglesphere:
**Description**

Unix timestamp is 1699981200 which is Tue Nov 14 09:00:00 PST 2023 or Tue Nov 14 17:00:00 UTC 2023.

### DIFF
--- a/superchain/configs/goerli/superchain.yaml
+++ b/superchain/configs/goerli/superchain.yaml
@@ -5,3 +5,4 @@ l1:
   explorer: https://goerli.etherscan.io
 
 protocol_versions_addr: "0x0C24F5098774aA366827D667494e9F889f7cFc08"
+canyon_time: 1699981200
--- a/superchain/configs/sepolia/superchain.yaml
+++ b/superchain/configs/sepolia/superchain.yaml
@@ -5,3 +5,4 @@ l1:
   explorer: https://sepolia.etherscan.io
 
 protocol_versions_addr: "0x79ADD5713B383DAa0a138d3C4780C7A1804a8090"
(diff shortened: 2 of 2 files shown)


superchain/configs/goerli/superchain.yaml
```diff
@@ -5,3 +5,4 @@ l1:
   explorer: https://goerli.etherscan.io
 
 protocol_versions_addr: "0x0C24F5098774aA366827D667494e9F889f7cFc08"
+canyon_time: 1699981200
```
trianglesphere, line 8: This need to end in an 8 b/c zora is on a 10s block time

superchain/superchain_test.go
```diff
@@ -294,7 +294,12 @@ func TestCanyonTimestampOnBlockBoundary(t *testing.T) {
 			chainCfg := OPChains[id]
 			canyonOffset := ct - chainCfg.Genesis.L2Time
 			// Block time is hardcoded in op-node/rollup/superchain.go
-			if canyonOffset%2 != 0 {
+			blockTime := 2
+			// Zora Base has a 10s block time.
```
sebastianst, line 298:
```suggestion
			// Zora Goerli has a 10s block time.
```

sebastianst (changes requested):

superchain/superchain_test.go
trianglesphere: Nice catch. Going to be removing Zora Goerli the registry so no longer relevant.

sebastianst (approved):

PR #32, opened 2023-11-16
feat: codegen combined json files

smartcontracts:
**Description**
Adds a script to codegen chain ID and address json files.

### DIFF
--- a/scripts/codegen.js
+++ b/scripts/codegen.js
@@ -1,0 +1,81 @@
+const fs = require('fs')
+const path = require('path')
+
+/**
+ * Generates the chainids.json file that sits at the root of the
+ * superchain/configs folder. Useful to have a combined JSON file for all of
+ * the various chain IDs to avoid dynamic imports.
+ */
+const chainids = () => {
+  const result = {}
+  const folder = path.resolve(__dirname, '../superchain/configs')
+  const subfolders = fs.readdirSync(folder, { withFileTypes: true })
+  for (const subfolder of subfolders) {
+    if (!subfolder.isDirectory()) {
+      continue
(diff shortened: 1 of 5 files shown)


scripts/codegen.js
```diff
@@ -0,0 +1,81 @@
...
+  const subfolders = fs.readdirSync(folder, { withFileTypes: true })
+  for (const subfolder of subfolders) {
+    if (!subfolder.isDirectory()) {
+      continue
+    }
+
+    const subpath = path.resolve(folder, subfolder.name)
+    const files = fs.readdirSync(subpath)
+    for (const file of files) {
+      if (!path.extname(file) === 'json') {
```
roninjin10, line 61:
```suggestion
      if (path.extname(file) !== 'json') {
```
This looks like a bug

scripts/codegen.js
```diff
@@ -0,0 +1,81 @@
...
+  const subfolders = fs.readdirSync(folder, { withFileTypes: true })
+  for (const subfolder of subfolders) {
+    if (!subfolder.isDirectory()) {
+      continue
+    }
+
+    const subpath = path.resolve(folder, subfolder.name)
+    const files = fs.readdirSync(subpath)
+    for (const file of files) {
+      if (!path.extname(file) === 'yaml') {
```
roninjin10, line 21:
```suggestion
      if (path.extname(file) !== 'yaml') {
```
This looks like a bug

scripts/codegen.js
smartcontracts: yep

roninjin10 (approved):

PR #42, opened 2023-12-12
Enforce address checksums in configs

protolambda:
**Description**

Enforce address checksums by making correct checksums a requirement when loading the config data with the Go bindings, which are verified in CI to load.

**Tests**

Verified the checksum util code with the test-cases of ERC-55.


maurelian (commented):
There is an error in CI. Perhaps a flake or race condition?

Otherwise the new test code LGTM overall, and I verified locally that incorrectly checksummed addresses are caught by `go test`. 

However AFAICT these test utils are not running in CI, or am I missing something?

protolambda:
Rebased on latest `main`

protolambda:
> There is an error in CI. Perhaps a flake or race condition?

Not sure, it was 3 weeks old, rebased and should be good now.

> However AFAICT these test utils are not running in CI, or am I missing something?

The Go tests load all the embedded configs (e.g. `TestGenesis` loads the genesis of all files).
As part of the yaml decoding it unmarshals into typed Go `Address` values. The checksum is enforced as part of that. If the checksum is invalid, the address parsing will fail. So, by covering all files in tests, we automatically cover all address encoding of all files.

mds1:
Reassigning to @geoknee to take this over the line. Let's also make sure there's a CI check that runs on PRs

geoknee:
Happy to carry this forward, but just pausing in case there is an opportunity to cut scope.

My reasoning is as follows: I'm not totally clear on the reasons for enforcing address checksums in this repository. 

I suppose the upsides are:
* We may catch a mistyped address on a contributors PR 
* ...but these addresses are moved around by a machine, so typos are very unlikely.
* ... and the addresses are _also_ highly likely to be involved in some deeper validation test, so any typo would make that test fail anyway

The downsides are:
* a risk of adding friction to contributors, who may need to worry about checksums when they didn't before
* slightly more complexity in this repo
* increased efforts for either contributors, maintainers, or both

With the enforcement in place, in order to make tests pass in general, we need to:

* go back and "fix" a bunch of non-checksummed addresses, by reapplying the checksum rules. This presumably bypasses the point of the check in the first place.
* fix the CONTRIBUTING.md guide to warn contributors to only add checksummed addresses, or alter the contributing tooling to do that automatically. Either way requires a bit of work, and again is most likely bypassing the point of the check if it is a case of blindly transforming an unchecksummed address into a checksummed one. As of now the monorepo files such as `rollup.json` are not checksummed, so putting enforcement in will certainly break the current contribution flow which scrapes addresses from there.

mds1:
One other upside is that some tooling requires checksummed addresses, for example if you paste an address from the registry into a solidity contract. The friction is a good point though, so my suggestion would be to have our `add-chain.sh` script convert to checksum addresses.

geoknee:
I think it depends on what our aims are. If we want to use checksums as a way to catch typo errors on data coming in to the repo, taking a lowercase string and transforming it into a checksum format doesn't make sense. If we just want to trust the data coming in, but ensure it is in a friendly format for copypasting into other tooling, then it makes a bit more sense.

mds1:
Yea the rest of our checks should catch typo errors (e.g. a call to a typo'd address will not return the expected data), so the main value of checksums IMO is for consumers of the data

protolambda:
Are we really going to bikeshed on the need of checksums? It's a net improvement to security. Let's get this functionality merged and maintain quality configs.

BlocksOnAChain:
@geoknee @mds1 what are the next steps for this one, I had it our Node board for a longer period and wanted to follow up on next steps - are we merging, doing more work to it or leaving it open with a comment?
CC: @protolambda for visibility

geoknee:
This PR is stale and has merge conflicts, so I am continuing the work here on #266. I can add it to my workload for this cycle.

PR #44, opened 2023-12-14
Add lura/mode/orderly mainnet superchain configs

kahuang:
**Description**

A clear and concise description of the features you're adding in this pull request.

**Tests**

Please describe any tests you've added. If you've added no tests, or left important behavior untested, please explain why not.

**Additional context**

Add any other context about the problem you're solving.

**Metadata**

- Fixes #[Link to Issue]

PR #45, opened 2023-12-18
Make forge script rpc limit explicit.

zchn:
We ran into our RPC limit in https://github.com/ethereum-optimism/superchain-registry/pull/44. Devinfra team bumped it. But I'm also adding additional options to the CI configuration so next time we have the flags ready to be tweaked.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,7 +69,10 @@ jobs:
           name: check security configs
           command: |
             forge build
-            forge script CheckSecurityConfigs --fork-url="https://ci-mainnet-l1.optimism.io"
+            # Note: If RPC is being rate-limited, consider reducing
+            # --compute-units-per-second or using --fork-retries and
+            # --fork-retry-backoff to stay under the limit.
+            forge script CheckSecurityConfigs --fork-url="https://ci-mainnet-l1.optimism.io" --compute-units-per-second=320
 
 workflows:
   main:


.circleci/config.yml
```diff
@@ -69,7 +69,10 @@ jobs:
           name: check security configs
           command: |
             forge build
-            forge script CheckSecurityConfigs --rpc-url="https://ci-mainnet-l1.optimism.io"
+            # Note: If RPC is being rate-limited, consider reducing
+            # --compute-units-per-second or using --fork-retries and
+            # --fork-retry-backoff to stay under the limit.
+            forge script CheckSecurityConfigs --fork-url="https://ci-mainnet-l1.optimism.io" --compute-units-per-second=320
```
mds1, line 75: It's possible `--rpc-url` and `--fork-url` behave the same, but in the monorepo scripts (and IME) `--rpc-url` is the flag used. So just wondering why that was changed?
```suggestion
            forge script CheckSecurityConfigs --rpc-url="https://ci-mainnet-l1.optimism.io" --compute-units-per-second=320
```

mds1: Got it, yea if you're sure `--rpc-url` and `--fork-url` are equivalent then I'm ok with that. My main concern was that `--rpc-url` was intended for `forge script` and `--fork-url` was for `forge test`, so I wasn't sure if they were exactly interchangeable and the CLI help was just incorrect (which it has been in the past). So as long as we know all the `--fork-*` flags work in the script environment, this works for me

mds1 (approved):

PR #46, opened 2023-12-19
add new fields required by Ecotone upgrade

roberto-bayardo:
**Description**

Adds two new parameters required by the Ecotone (formerly Eclipse) upgrade to the GenesisSystemConfig.

### DIFF
--- a/superchain/superchain.go
+++ b/superchain/superchain.go
@@ -322,9 +322,14 @@ func (c ContractImplementations) Copy() ContractImplementations {
 
 type GenesisSystemConfig struct {
 	BatcherAddr Address `json:"batcherAddr"`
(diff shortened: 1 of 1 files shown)


tynes:
I am not sure if we want to merge this because we may want to hardcode the values so that L1 and the L2 view never get out of sync. The current implementation of the smart contracts in https://github.com/ethereum-optimism/optimism/pull/8723 migrates the legacy scalar to the base fee scalar and sets the blob base fee scalar to 0 so we could just hardcode those values. We need to consider the cases of new chains and upgrading chains which adds some complexity to the smart contract implementations

tynes:
I am feeling confident that we do not need this PR based on the params being deterministic

PR #53, opened 2024-01-16
Feat/add contracts address manager proxy admin

0xfuturistic:
addresses https://github.com/ethereum-optimism/optimism/pull/8859#issuecomment-1881663953

**important**: it considers contracts without semver as contracts with a semver version of `0.0.0`, because it sets `0.0.0` as the placeholder for the two new contracts added to `ContractVersions` (`AddressManager` & `ProxyAdmin`).


tynes:
Hmm these contracts are not versioned so i don't think that we should add them here

PR #43, opened 2023-12-13
Add Mode mainnet

ssallam:
**Description**

Adding Mode mainnet to superchain registry.

Addresses and rollup can be verified here:
https://github.com/mode-network/chain-deployments/blob/main/mainnet/addresses.json
https://github.com/mode-network/chain-deployments/blob/main/mainnet/rollup.json

Used this command to generate bytecodes and genesis:

go run ./op-chain-ops/cmd/registry-data --l2-genesis=../chain-deployments/blob/main/mainnet/genesis-l2.json --bytecodes-dir=$SUPERCHAIN_REPO/superchain/extra/bytecodes --output=$SUPERCHAIN_REPO/superchain/extra/genesis/$SUPERCHAIN_TARGET/$CHAIN_NAME.json.gz

**Tests**

None

PR #59, opened 2024-01-20
fix: update mode mainnet rpc and sequencer urls

agileurbanite:
**Description**

The current mainnet rpc and sequencer urls are incorrect and would like it updated to:

```
public_rpc: https://mainnet.mode.network
sequencer_rpc: https://mainnet-sequencer.mode.network
```

**Tests**

N/A


agileurbanite:
@trianglesphere would appreciate reviews of this PR as the registry has incorrect url that was merged in #44 but our initial PR #43 had the correct urls. This PR addresses the fix.  Thank you for your help!

EvanJRichard:
Spoke to @trianglesphere, verified with `cast`, these values indeed match #43, LGTM :+1: merging shortly

PR #60, opened 2024-01-22
WIP begin work on checking semver of deployed contracts on each chain. 

geoknee:
NB: this is inspired by [`op-version-check`](https://github.com/ethereum-optimism/optimism/pull/8859) over in the monorepo. As yet it does not import any of that code nor replicate all of the functionality. Raising this to get early feedback before ploughing on.


superchain/superchain-version_test.go
```diff
@@ -0,0 +1,61 @@
+package superchain_test
```
geoknee, line 1: Note new package to allow us to import from the monorepo without incurring a dependency cycle (nothing will import this code).

superchain/superchain-version_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+// TestContractVersionsCheck will fail if the superchain semver file
+// is not read correctly.
+func TestContractVersionsCheck(t *testing.T) {
+	if err := superchain.SuperchainSemver.SanityCheck(); err != nil {
+		t.Fatal(err)
+	}
+
+	desiredSemver := superchain.SuperchainSemver.OptimismPortal
+	// TODO complete all members of SuperchainSemver
```
geoknee, line 29: This can be achieved either through manually enumerating all of the contracts, or by using reflection to iterate through the members of the struct. Both patterns are in use elsewhere.

superchain/superchain-version_test.go
```diff
@@ -0,0 +1,61 @@
...
+	// TODO resolve missing contracts
+	// TODO resolve extraneous contracts
+	for _, chain := range superchain.OPChains {
+		client, err := ethclient.Dial(superchain.Superchains[chain.Superchain].Config.L1.PublicRPC)
+		checkErr(t, err)
+		actualSemver, err := getVersion(context.Background(), common.Address(superchain.Addresses[chain.ChainID].OptimismPortalProxy), client) // TODO resolve difference in names
```
geoknee, line 35: The desired semver is populated from `semver.yaml`, and gets mapped to `OptimismPortal`. But the actual semver is pulled from a contract with a name with the `Proxy` suffix:
https://github.com/ethereum-optimism/superchain-registry/blob/68a2c8bb2ce15ccaabb051979d383d50d0d332c8/superchain/extra/addresses/goerli/base.json#L8
And similarly for other contracts.

superchain/superchain-version_test.go
```diff
@@ -0,0 +1,61 @@
...
+// is not read correctly.
+func TestContractVersionsCheck(t *testing.T) {
+	if err := superchain.SuperchainSemver.SanityCheck(); err != nil {
+		t.Fatal(err)
+	}
+
+	desiredSemver := superchain.SuperchainSemver.OptimismPortal
+	// TODO complete all members of SuperchainSemver
+	// TODO resolve missing contracts
+	// TODO resolve extraneous contracts
```
geoknee, line 31: I notice there is not a 1-1 mapping between the set of contract names with declared desired semver, and the set of contract names for each chain. Also some chains have contracts which others lack.

superchain/superchain-version_test.go
```diff
@@ -0,0 +1,61 @@
...
+	for _, chain := range superchain.OPChains {
+		client, err := ethclient.Dial(superchain.Superchains[chain.Superchain].Config.L1.PublicRPC)
+		checkErr(t, err)
+		actualSemver, err := getVersion(context.Background(), common.Address(superchain.Addresses[chain.ChainID].OptimismPortalProxy), client) // TODO resolve difference in names
+		checkErr(t, err)
+		if desiredSemver != actualSemver {
+			t.Fatalf("OptimismPortalProxy should have version %v but has version %v (%s on %s)", desiredSemver, actualSemver, chain.Name, chain.Superchain)
```
geoknee, line 38: It's not yet clear to me if we should require an exact match or allow for the desired semver to declare a range of acceptable versions https://devhints.io/semver .

superchain/superchain-version_test.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+	}
+
+}
+
+// getVersion will get the version of a contract at a given address, if it exposes a version() method.
+func getVersion(ctx context.Context, addr common.Address, client *ethclient.Client) (string, error) {
+	result, err := client.CallContract(context.Background(), ethereum.CallMsg{
```
geoknee, line 46: It would be nice to import from the monorepo so we can use preexisting functions for this. However, I found that there is no published version with the `ISemVer` bindings (and associated utils) available. So I rolled my own utility for now just to get the ball rolling.

superchain/superchain-version_test.go
EvanJRichard: Because we already have plans to add another contract at the end of this work, I'm leaning towards reflection, then we wouldn't have to come back and edit this test when a new contract gets added - right?

superchain/superchain-version_test.go
EvanJRichard: I think for now we should do `>=` [whatever OP Mainnet has] - this will allow chains like Base, which have newer versions than OP Mainnet, to pass this test. Eventually we'll want to do strict equality - but at that point, we'll hopefully be doing strict bytecode equality checking anyways

superchain/superchain-version_test.go
EvanJRichard: I think that makes sense. I was briefly concerned "what if OP Mainnet, which we want to use as our source of truth, doesn't match `semver.yaml`?" But it should never mismatch, and if it does, we'll detect it here in this tests.

superchain/superchain-version_test.go
geoknee: Yes. We might want to take a pass through other bits of code that are using the more brittle / handcrafted approach and update them, too.

superchain/superchain-version_test.go
geoknee: Agreed. But my real concern here was the mismatch in naming `OptimismPortal` vs `OptimismPortalProxy`. Should these be the same, or not?

geoknee: As per discussion today, contributors are asked to provide proxy addresses. We therefore need to call `version` on those proxy contracts (which will pass the call through to the implementation contract), to read the semantic version we wish to inspect / check. So we will just add the `Proxy` suffix manually in this test.

superchain/superchain-version_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+	checkAllOPChainsSatisfySemver := func(contractName string) {
+		desiredSemver := reflect.ValueOf(superchain.SuperchainSemver).FieldByName(contractName).String()
+
+		// ASSUMPTION: we will check the version of the implementation via the declared proxy contract
+		proxyContractName := contractName + "Proxy"
+
+		for _, chain := range superchain.OPChains {
+			rpcEndpoint := superchain.Superchains[chain.Superchain].Config.L1.PublicRPC
+			t.Logf("Dialing %s...", rpcEndpoint)
```
geoknee, line 47: Right now we are blocked on RPC endpoints returning `400 Bad Request: {"error": "request has an invalid host info","id":1}`. I guess we need an API key to make this test reliable?

We may need to inject this manually at test time?

EvanJRichard: Discussed on Slack: we'll use Alchemy or something to unblock for now, then figure out the better permanent solution

geoknee: I think the problem actually was mostly with the Goerli endpoint (I guess because it is deprecated). For now we can work apparently work around it using a different endpoint (tenderly) plus some retry logic so we don't anger their rate limiter.

geoknee:
As of commit [57610bd](https://github.com/ethereum-optimism/superchain-registry/pull/60/commits/57610bd9721fda7594cdb8c2c5c21c2944af4fe8), this test failing with the following output:
```
=== RUN   TestContractVersionsCheck
    superchain-version_test.go:64:    mainnet:Mode                :L1CrossDomainMessenger         has version 1.4.1 but should have version 1.7.0
    superchain-version_test.go:64:    mainnet:Mode                :L1ERC721Bridge                 has version 1.1.2 but should have version 1.4.0
    superchain-version_test.go:64:    mainnet:Mode                :L1StandardBridge               has version 1.1.1 but should have version 1.4.0
    superchain-version_test.go:64:    mainnet:Mode                :L2OutputOracle                 has version 1.3.1 but should have version 1.6.0
    superchain-version_test.go:64:    mainnet:Mode                :OptimismMintableERC20Factory   has version 1.1.1 but should have version 1.6.0
    superchain-version_test.go:64:    mainnet:Mode                :OptimismPortal                 has version 1.7.2 but should have version 1.10.0
    superchain-version_test.go:56:    mainnet:Mode                :SystemConfigProxy              has version UNSPECIFIED, but should have version 1.10.0
    superchain-version_test.go:64:    mainnet:OP-Mainnet          :L1CrossDomainMessenger         has version 1.4.0 but should have version 1.7.0
    superchain-version_test.go:64:    mainnet:OP-Mainnet          :L1ERC721Bridge                 has version 1.0.0 but should have version 1.4.0
    superchain-version_test.go:64:    mainnet:OP-Mainnet          :L1StandardBridge               has version 1.1.0 but should have version 1.4.0
    superchain-version_test.go:64:    mainnet:OP-Mainnet          :L2OutputOracle                 has version 1.3.0 but should have version 1.6.0
    superchain-version_test.go:64:    mainnet:OP-Mainnet          :OptimismMintableERC20Factory   has version 1.1.0 but should have version 1.6.0
    superchain-version_test.go:64:    mainnet:OP-Mainnet          :OptimismPortal                 has version 1.6.0 but should have version 1.10.0
    superchain-version_test.go:56:    mainnet:OP-Mainnet          :SystemConfigProxy              has version UNSPECIFIED, but should have version 1.10.0
    superchain-version_test.go:64:    mainnet:PGN                 :L1CrossDomainMessenger         has version 1.4.0 but should have version 1.7.0
    superchain-version_test.go:64:    mainnet:PGN                 :L1ERC721Bridge                 has version 1.1.1 but should have version 1.4.0
    superchain-version_test.go:64:    mainnet:PGN                 :L1StandardBridge               has version 1.1.0 but should have version 1.4.0
    superchain-version_test.go:64:    mainnet:PGN                 :L2OutputOracle                 has version 1.3.0 but should have version 1.6.0
    superchain-version_test.go:64:    mainnet:PGN                 :OptimismMintableERC20Factory   has version 1.1.0 but should have version 1.6.0
    superchain-version_test.go:64:    mainnet:PGN                 :OptimismPortal                 has version 1.6.0 but should have version 1.10.0
    superchain-version_test.go:56:    mainnet:PGN                 :SystemConfigProxy              has version UNSPECIFIED, but should have version 1.10.0
    superchain-version_test.go:64:    mainnet:Base                :L1CrossDomainMessenger         has version 1.4.0 but should have version 1.7.0
    superchain-version_test.go:64:    mainnet:Base                :L1ERC721Bridge                 has version 1.1.1 but should have version 1.4.0
    superchain-version_test.go:64:    mainnet:Base                :L1StandardBridge               has version 1.1.0 but should have version 1.4.0
    superchain-version_test.go:64:    mainnet:Base                :L2OutputOracle                 has version 1.3.0 but should have version 1.6.0
    superchain-version_test.go:64:    mainnet:Base                :OptimismMintableERC20Factory   has version 1.1.0 but should have version 1.6.0
    superchain-version_test.go:64:    mainnet:Base                :OptimismPortal                 has version 1.7.0 but should have version 1.10.0
    superchain-version_test.go:56:    mainnet:Base                :SystemConfigProxy              has version UNSPECIFIED, but should have version 1.10.0
    superchain-version_test.go:64:    mainnet:Lyra                :L1CrossDomainMessenger         has version 1.4.1 but should have version 1.7.0
    superchain-version_test.go:64:    mainnet:Lyra                :L1ERC721Bridge                 has version 1.1.2 but should have version 1.4.0
    superchain-version_test.go:64:    mainnet:Lyra                :L1StandardBridge               has version 1.1.1 but should have version 1.4.0
    superchain-version_test.go:64:    mainnet:Lyra                :L2OutputOracle                 has version 1.3.1 but should have version 1.6.0
    superchain-version_test.go:64:    mainnet:Lyra                :OptimismMintableERC20Factory   has version 1.1.1 but should have version 1.6.0
    superchain-version_test.go:64:    mainnet:Lyra                :OptimismPortal                 has version 1.7.2 but should have version 1.10.0
    superchain-version_test.go:56:    mainnet:Lyra                :SystemConfigProxy              has version UNSPECIFIED, but should have version 1.10.0
    superchain-version_test.go:64:    mainnet:Orderly Network     :L1CrossDomainMessenger         has version 1.4.1 but should have version 1.7.0
    superchain-version_test.go:64:    mainnet:Orderly Network     :L1ERC721Bridge                 has version 1.1.2 but should have version 1.4.0
    superchain-version_test.go:64:    mainnet:Orderly Network     :L1StandardBridge               has version 1.1.1 but should have version 1.4.0
    superchain-version_test.go:64:    mainnet:Orderly Network     :L2OutputOracle                 has version 1.3.1 but should have version 1.6.0
    superchain-version_test.go:64:    mainnet:Orderly Network     :OptimismMintableERC20Factory   has version 1.1.1 but should have version 1.6.0
    superchain-version_test.go:64:    mainnet:Orderly Network     :OptimismPortal                 has version 1.7.2 but should have version 1.10.0
    superchain-version_test.go:56:    mainnet:Orderly Network     :SystemConfigProxy              has version UNSPECIFIED, but should have version 1.10.0
    superchain-version_test.go:64:    mainnet:Zora                :L1CrossDomainMessenger         has version 1.4.0 but should have version 1.7.0
    superchain-version_test.go:64:    mainnet:Zora                :L1ERC721Bridge                 has version 1.1.1 but should have version 1.4.0
    superchain-version_test.go:64:    mainnet:Zora                :L1StandardBridge               has version 1.1.0 but should have version 1.4.0
    superchain-version_test.go:64:    mainnet:Zora                :L2OutputOracle                 has version 1.3.0 but should have version 1.6.0
    superchain-version_test.go:64:    mainnet:Zora                :OptimismMintableERC20Factory   has version 1.1.0 but should have version 1.6.0
    superchain-version_test.go:64:    mainnet:Zora                :OptimismPortal                 has version 1.6.0 but should have version 1.10.0
    superchain-version_test.go:56:    mainnet:Zora                :SystemConfigProxy              has version UNSPECIFIED, but should have version 1.10.0
--- FAIL: TestContractVersionsCheck (17.11s)
FAIL
FAIL	github.com/ethereum-optimism/superchain-registry/superchain	17.441s
FAIL
```

I think we need to check our assumption that `semver.yaml` has the source of truth about contract versions. I know that one idea is that we just use OP mainnet as the source of truth: but still it seems like a potential footgun to have `semver.yaml` hanging around unless it is clear what it means (and that we are confident it will stay up to date etc).

superchain/configs/goerli-dev-0/superchain.yaml
```diff
@@ -1,7 +1,7 @@
 name: Goerli Dev 0
 l1:
   chain_id: 5
-  public_rpc: https://ethereum-goerli-rpc.allthatnode.com
+  public_rpc: wss://goerli.gateway.tenderly.co
```
geoknee, line 4: Might be something to pinch off this PR on to a separate one, if we think it makes sense. I think the existing RPC endpoint is deprecated.

superchain/superchain-version_test.go
```diff
@@ -0,0 +1,109 @@
...
+			}
+			if desiredSemver != actualSemver {
+				t.Errorf("%10s:%-20s:%-30s has version %s but should have version %s", chain.Superchain, chain.Name, contractName, actualSemver, desiredSemver)
+			}
+		}
+	}
+
+	for _, chain := range superchain.OPChains {
+		if strings.ToLower(chain.Superchain) != "mainnet" {
+			continue // concentrate on mainnet superchain for now
```
geoknee, line 71: Flagging this. I intend for this to be a temporary thing while we work some other questions out. 
```suggestion
			continue // TODO Check all superchains
```

superchain/go.mod
```diff
@@ -1,12 +1,54 @@
...
-
+go 1.21
 require (
-	golang.org/x/crypto v0.13.0
-	golang.org/x/exp v0.0.0-20230905200255-921286631fa9
-	golang.org/x/mod v0.12.0
+	github.com/ethereum/go-ethereum v1.13.10
+	golang.org/x/crypto v0.18.0
+	golang.org/x/exp v0.0.0-20231110203233-9a3e6036ecaa
+	golang.org/x/mod v0.14.0
```
protolambda, line 8: So far we have avoided depending on any Implementation. Especially upstream geth. Since the registry gets pulled in by the monorepo, which depends on a different conflicting version of geth.

If we build tooling to check advanced things in this repo, it would be better to put it as CLI tool in the optimism monorepo, and then call the CLI from the CI here.

protolambda (changes requested):
I believe this introduces circular dependencies between the configs and config consuming code (op-geth, and monorepo).

protolambda:
The package separation thing might work, but still counts on go module dependencies to resolve without conflicts, and we only detect issues when updating it in op-geth/monorepo, rather than here. Ideally we put this in its own CLI in the monorepo and don't create any heavy go.mod here with potential cycles. Or alternatively, maybe a separate Go module in this repo, that pulls in the local superchain registry go module as well as what might be required from the monorepo or op-geth, to detect any Go module conflicts between the them. The latter might actually work quite well for purposes of encapsulating tests and getting compiler feedback on breaking changes.

superchain/go.mod
geoknee: Thanks for the feedback @protolambda! I hadn't considered the conflicting geth versions. I think we could remove the dependency on geth. We are only making a very simple on chain query, and that could be done with a hand rolled JSON-RPC request.

geoknee: For full context:

My instinct is to steer away from using CLI tools from the monorepo in this repo. That's for two reasons:
1. Running compiled Go code as an executable, and then parsing the output (probably with more Go code) in order to inspect / assert on it seems somewhat inelegant and brittle (no type checking, for example).
2. This would bake in a cyclic dependency -- not one the Go compiler is going to complain about, but still somewhat of an antipattern.

protolambda:
The Go module separation can be achieved with a `go mod replace github.com/ethereum-optimism/superchain-registry/superchain => ../superchain`, or with a `go.work` which defines how local go modules can be combined. Then local editing is seamless, no continuous go.sum updates.

geoknee:
> 271ae42

I tried this out here [271ae42](https://github.com/ethereum-optimism/superchain-registry/pull/60/commits/271ae4219185273f3256dcb44bb48a543287966b) . I agree it can be a pretty nice solution, allowing us to import heavier libraries into the validation code without polluting the `superchain` module (and therefore without polluting the `monorepo`).

geoknee:
In favour of #62

PR #62, opened 2024-01-25
Add a `validate` package including new tests for validating contract versions.

geoknee:
**Description**

* New **module** for validating superchain configuration
  * This approach suggested by @protolambda to avoid the `monorepo` inheriting dependencies from the `superchain` module. 
  * We can use heavier libraries, and even pull stuff in from the `monorepo` in this module

**Tests**

`TestContractVersions` which:
* Scrapes contract names from `semver.yaml`
* appends "Proxy" to those names
* Calls out to an RPC provider to get the on-chain version of each of those contracts on OP Mainnet
* Loops through all OPChains and checks _their_ on chain versions are greater than or equal to the above

**Additional context**

This test duplicates some of the functionality of `op-version-check` command line tool from the monorepo. That tool could be removed. 

**Metadata**

- Fixes https://github.com/ethereum-optimism/client-pod/issues/485


validation/superchain-version_test.go
```diff
@@ -0,0 +1,135 @@
...
+	checkErr(t, err)
+
+	t.Logf("Desired Semver from %s/%s : \n %+v", SOURCE_OF_TRUTH_SUPERCHAIN, superchain.OPChains[SOURCE_OF_TRUTH_CHAINID].Name, string(logOutput))
+
+	checkOPChainSatisfiesSemver := func(chain *superchain.ChainConfig) {
+
+		for _, field := range semverFields {
+			if field.Name == "SystemConfig" {
+				continue // TODO this is specified in semver.yaml but not in an address list.
+			}
```
geoknee, line 70: This is less than ideal. There's a mismatch in these types which we need to resolve:
https://github.com/ethereum-optimism/superchain-registry/blob/baa49d9c2242562a446254b74c7ebd2d2b5b14ee/superchain/superchain.go#L191-L202

(this is the schema for `semver.yaml`)
and

https://github.com/ethereum-optimism/superchain-registry/blob/baa49d9c2242562a446254b74c7ebd2d2b5b14ee/superchain/superchain.go#L62-L72

(this is what each contributed chain is [asked to provide](https://github.com/ethereum-optimism/superchain-registry/blob/bce4dc12678c705f7c7b0930094d9a21ab0fc674/CONTRIBUTING.md#extras)).

geoknee:
Pairing with @mslipper today, I'll record some TODOs here:
- [ ] semver check should be a strict equality check
- [ ] consider an exclude list so we can concentrate on new chains being added if some existing ones would fail the checks
- [ ] avoid reflection, let's just bake a list of contracts into the test
- [ ] we should be doing a sanity check that chainId and rpc endpoints are consistent.

.circleci/config.yml
```diff
@@ -20,6 +20,22 @@ jobs:
...
+    executor:
+      name: go/default  # is based on cimg/go
+      tag: '1.20'
+    steps:
+      - checkout:
+          path: ~/superchain-registry
+      - go/load-cache
+      - go/mod-download
+      - go/save-cache
+      - run: golangci-lint run -c=.golangci.yml
```
mslipper, line 34: I think we should extract the linting into its own job that lints every project at once. Makes it easier to identify failures since lints won't block the tests from running.

validation/superchain-version_test.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	return s[0].(string), nil
+}
+
+// getVersionWithRetries is a wrapper for getVersion
+// which, on error, will wait 5 seconds an retry up to 3 times.
+func getVersionWithRetries(ctx context.Context, addr common.Address, client *ethclient.Client) (string, error) {
```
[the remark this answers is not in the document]
mslipper: `op-service` has a `retry` package with some exponential backoff methods you can use here. Would rather use the lib if we can.

validation/superchain-version_test.go
```diff
@@ -0,0 +1,135 @@
...
+			checkOPChainSatisfiesSemver(chain)
+		}
+	}
+}
+
+// getVersion will get the version of a contract at a given address, if it exposes a version() method.
+func getVersion(ctx context.Context, addr common.Address, client *ethclient.Client) (string, error) {
+	result, err := client.CallContract(context.Background(), ethereum.CallMsg{
+		To:   &addr,
+		Data: common.FromHex("0x54fd4d50")}, // this is the function selector for "version"
```
mslipper, line 105: [nit] Pull into a `const`

validation/superchain-version_test.go
geoknee: The top data type seems to be the canonical one with respect to contract names (which have a version() method, at least). These does seem to imply we want to update `CONTRIBUTING.md` to ask contributors for an address for `SystemConfig`, which it doesn't currently do.

validation/superchain-version_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+	isSemverAcceptable := func(desired, actual string) bool {
+		return desired == actual
+	}
+
+	checkOPChainSatisfiesSemver := func(chain *ChainConfig) {
+		rpcEndpoint := Superchains[chain.Superchain].Config.L1.PublicRPC
+
+		if rpcEndpoint == "" {
+			t.Errorf("%s has MISSING RPC endpoint", chain.Superchain)
```
mslipper, line 54: [nit] We use [testify](https://github.com/stretchr/testify) for all of our assertions - can you please convert these to use the `require` package from Testify?

validation/superchain-version_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+		if err != nil {
+			t.Errorf("could not dial rpc endpoint %s: %v", rpcEndpoint, err)
+			return
+		}
+
+		semverFields := reflect.VisibleFields(reflect.TypeOf(SuperchainSemver[chain.Superchain]))
+
+		for _, field := range semverFields {
+			desiredSemver := reflect.Indirect(reflect.ValueOf(SuperchainSemver[chain.Superchain])).FieldByName(field.Name).String()
```
mslipper, line 68: I find reflection to difficult to read and reason about. If we absolutely have to use it, let's hide the complexity behind utility getter methods on the structs containing the versions/addresses. Something like:

```go
// within superchain.go
func (c ContractVersions) VersionFor(name string) (string, error) {}

func (a AddressList) AddressFor(name string) (string, error) {}

// in the test

var requiredContracts = []string { "OptimismPortal", ... }
```

This way we avoid proliferating the use of `reflect` elsewhere in the codebase, and expose a cleaner more readable API to future developers.

mslipper (changes requested):

.circleci/config.yml
geoknee: [4a43a41](https://github.com/ethereum-optimism/superchain-registry/pull/62/commits/4a43a41ef4b1234f4d896eebd6bd63eb96dbab8d)

geoknee: I also tidied up the testing part of the CI pipeline [eb63f1a](https://github.com/ethereum-optimism/superchain-registry/pull/62/commits/eb63f1a565ef87d00317052c72398cb238f7398b). This will mean the two module's tests won't run concurrently, but IMO this is acceptable.

validation/superchain-version_test.go
geoknee: In the end I pulled in the `op-bindings` library [489a386](https://github.com/ethereum-optimism/superchain-registry/pull/62/commits/489a386e91b21fb9bc550b2137b77c717d17661b) instead.

validation/superchain-version_test.go
geoknee: [549d7d6](https://github.com/ethereum-optimism/superchain-registry/pull/62/commits/549d7d64b8065c1a049cb7f4d43ba8c15b793117)

validation/superchain-version_test.go
geoknee: I went with your suggestion, only I didn't even use `reflect` in the implementations [9f78655](https://github.com/ethereum-optimism/superchain-registry/pull/62/commits/9f78655bdf19aa6a25542e48f663bb478d65a095).

I think there is maybe still a question about when to check for a missing address (at init time or at runtime). For now,  the new getters will return an error if the struct has a zero/default value for a field.

validation/superchain-version_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+	return version, nil
+
+}
+
+// getVersionWithRetries is a wrapper for getVersion
+// which retries up to 10 times with exponential backoff.
+func getVersionWithRetries(ctx context.Context, addr common.Address, client *ethclient.Client) (string, error) {
+	const maxAttempts = 10
+	return retry.Do(context.Background(), maxAttempts, retry.Exponential(), func() (string, error) {
```
mslipper, line 119: [nit] Pass in `ctx` from above to allow this to be cancelled with the parent context

validation/superchain-version_test.go
```diff
@@ -0,0 +1,122 @@
...
+		return getVersion(context.Background(), addr, client)
```
mslipper, line 120: [nit] Ditto above comment

mslipper (approved):
LGTM, added some final nits

validation/superchain-version_test.go
geoknee: [94b84e7](https://github.com/ethereum-optimism/superchain-registry/pull/62/commits/94b84e74fc6d0516da5214369cb48631212da86a)

validation/superchain-version_test.go
geoknee: [94b84e7](https://github.com/ethereum-optimism/superchain-registry/pull/62/commits/94b84e74fc6d0516da5214369cb48631212da86a)

PR #68, opened 2024-01-29
ci: attempt to fix go modules caching behaviour

geoknee:
**Description**
The existing workflow has errors because it tries to use `go.sum` from the top level as a cache key. This PR changes it to use `go.work`, which I believe is more appropriate for a Go workspace (we now have multiple modules at play). 

**Tests**

manual


mslipper:
If the numbers above are correct let's strip the caching out for now.

.circleci/config.yml
```diff
@@ -21,9 +21,13 @@ jobs:
     steps:
       - checkout:
           path: ~/superchain-registry
-      - go/load-cache
+      - restore_cache:
+          key: v1-{{checksum "go.work"}}
       - go/mod-download
-      - go/save-cache
+      - save_cache:
+          key: v1-{{checksum "go.work"}}
```
geoknee: Actually I don't think `go.work` is a good choice as a cache key. I had probably meant to use `go.work.sum`, but I don't believe that is a good choice either.

No matter for now, as the short term plan is to rip out caching.

mslipper (approved):

README.md
```diff
@@ -26,6 +26,10 @@ See [`op-chain-ops`] for config tooling and
 [`op-chain-ops`]: https://github.com/ethereum-optimism/optimism/tree/develop/op-chain-ops
 [`op-bindings`]: https://github.com/ethereum-optimism/optimism/tree/develop/op-bindings
 
+
+## Validation Go Module
+A second module exists in this repo whose purpose is to validate the config exported by the `superchain` module. It is a separate module to avoid import cycles and polluting downstream dependencies with things like `go-ethereum` (which is use in the validation tests). The modules are tracked by a top level `go.work` file. The associated `go.work.sum` file is gitignored and not important to typical workflows, which should mirror those of the [CI configuration](.circleci/config.yml).
```
sebastianst, line 31:
```suggestion
A second module exists in this repo whose purpose is to validate the config exported by the `superchain` module. It is a separate module to avoid import cycles and polluting downstream dependencies with things like `go-ethereum` (which is used in the validation tests). The modules are tracked by a top level `go.work` file. The associated `go.work.sum` file is gitignored and not important to typical workflows, which should mirror those of the [CI configuration](.circleci/config.yml).
```

sebastianst (approved):
LGTM thanks for cleaning that up! Found one typo.

Would also be nice to add a short pointer to the `CONTRIBUTING.md` guide for adding new chains.

protolambda (approved):

PR #69, opened 2024-01-29
CheckSecurityConfigs: support testnets.

zchn:
This PR adds support for performing `CheckSecurityConfigs` on testnets by
1. making the directory and network to perform the checks on input parameters.
2. fixing sepolia and goerli addresses
3. making the FINALIZATION_PERIOD_SECONDS check only required for mainnet

### DIFF
--- a/scripts/CheckSecurityConfigs.s.sol
+++ b/scripts/CheckSecurityConfigs.s.sol
@@ -39,15 +39,24 @@ contract CheckSecurityConfigs is Script {
      */
     function run() public {
         string memory network;
-        if (block.chainid == 1) network = "mainnet";
-        else if (block.chainid == 11155111) network = "sepolia";
-        else if (block.chainid == 5) network = "goerli";
-        else revert(string.concat("Unsupported chain ID: ", vm.toString(block.chainid),
-                                  ". Please call runOnDir(string,bool) directly."));
+        if (block.chainid == 1) {
+            network = "mainnet";
+        } else if (block.chainid == 11155111) {
+            network = "sepolia";
+        } else if (block.chainid == 5) {
+            network = "goerli";
+        } else {
+            revert(
+                string.concat(
+                    "Unsupported chain ID: ",
+                    vm.toString(block.chainid),
+                    ". Please call runOnDir(string,bool) directly."
+                )
+            );
+        }
         string memory jsonDir = string.concat("superchain/extra/addresses/", network);
         runOnDir(jsonDir, block.chainid == 1);
     }
-
 
     function runOnDir(string memory jsonDir, bool isMainnet) public {
         VmSafe.DirEntry[] memory addressesJsonEntries = vm.readDir(jsonDir);


scripts/CheckSecurityConfigs.s.sol
```diff
@@ -37,24 +37,24 @@ contract CheckSecurityConfigs is Script {
     /**
      * @notice The entrypoint function.
      */
-    function run() external {
-        VmSafe.DirEntry[] memory addressesJsonEntries = vm.readDir("superchain/extra/addresses/mainnet");
+    function run(string memory jsonDir, string memory network) external {
```
mds1, line 40: The network should match the provided RPC URL's chain ID, and we currently do not verify that. Since the RPC URL is a required command line arg, we could remove both of these input args and key off `block.chainid`, which will be the value returned by the RPC URL, to enforce that automatically. This also improves UX by reducing the number of args needed. For example:
```solidity
string memory jsonDir;
string memory network;
if (block.chainid == 1) {
  jsonDir = "superchain/extra/addresses/mainnet";
  network = "mainnet";
} else if (block.chainid == 11155111) {
  jsonDir = "superchain/extra/addresses/sepolia";
  network = "sepolia";
} else {
  string memory chainId = vm.toString(block.chainid);
  revert(string.concat("Unsupported chain ID: ", chainId));
}
```

If we do need to support arbitrary `jsonDir` values, we can have another run method to support that, but this way the common case of `superchain/extra/addresses/<network>` becomes safer and simpler

superchain/extra/addresses/goerli/base.json
```diff
@@ -6,5 +6,10 @@
   "L2OutputOracleProxy": "0x2A35891ff30313CcFa6CE88dcf3858bb075A2298",
   "OptimismMintableERC20FactoryProxy": "0xa88530E2DD811363cA3Ef479dBab3C0BF73d90b1",
   "OptimismPortalProxy": "0xe93c8cD0D409341205A592f8c4Ac1A5fe5585cfA",
-  "ProxyAdmin": "0xbc0Fc544736b7d610D9b05F31B182C8154BEf336"
-}
\ No newline at end of file
+  "ProxyAdmin": "0xbc0Fc544736b7d610D9b05F31B182C8154BEf336",
+  "SystemConfigProxy": "0xb15eea247eCE011C68a614e4a77AD648ff495bc1",
+  "ProxyAdminOwner": "0x444b8C3E4eA49cE15A93D96AfA83D421F6049524",
+  "SystemConfigOwner": "0x4C35Ca57616E0d5fD808574772f632D8dA4eadCa",
```
mds1, line 12: The goal of this file to serve as a reference and source of truth, which we verify in the `CheckSecurityConfigs` script? I ask because as of https://github.com/ethereum-optimism/optimism/pull/6555, we can now only define a chain's `SystemConfig` and then discover everything else. Wondering if we can leverage that in this repo (once everyone is on MCP L1 contracts)

superchain/extra/addresses/addresses.json
mds1, line 1: What's the relation between this file and `superchain/extra/address/<l1-chain>/<l2-chain>.json`? It seems they hold the same data

scripts/CheckSecurityConfigs.s.sol
mds1: Thanks! Here's a version that's slightly simpler to maintain, of course either is ok:

```solidity
string memory network;
if (block.chainid == 1) network = "mainnet";
else if (block.chainid == 11155111) network = "sepolia";
else revert(string.concat("Unsupported chain ID: ", vm.toString(block.chainid));

string memory jsonDir = string.concat("superchain/extra/addresses/", network);
```

scripts/CheckSecurityConfigs.s.sol
```diff
@@ -90,12 +111,14 @@ contract CheckSecurityConfigs is Script {
...
     }
 
-    function checkL2OutputOracleProxy(ProtocolAddresses memory addresses) internal {
+    function checkL2OutputOracleProxy(ProtocolAddresses memory addresses, bool isMainnet) internal {
         console2.log("Checking L2OutputOracleProxy %s", addresses.L2OutputOracleProxy);
         isAdminOf(addresses.ProxyAdmin, addresses.L2OutputOracleProxy);
         checkAddressIsExpected(addresses.Challenger, addresses.L2OutputOracleProxy, "CHALLENGER()");
-        // Reusing the logic in checkAddressIsExpected below for simplicity.
-        checkAddressIsExpected(address(7 days), addresses.L2OutputOracleProxy, "FINALIZATION_PERIOD_SECONDS()");
+        if (isMainnet) {
```
mds1, line 118: We could remove this `bool isMainnet` from all functions and just use `block.chainid == 1` here
```suggestion
        if (block.chainid == 1) {
```

mds1: That's fair, thanks for explaining!

mds1 (approved):
Looks great!

PR #70, opened 2024-01-29
sepolia-dev-0: Add chain oplabs-devnet-0 and set Ecotone activation time

sebastianst:
Fixes https://github.com/ethereum-optimism/protocol-quest/issues/119


trianglesphere (approved):
@zchn LGTM

validation/go.mod
```diff
@@ -5,8 +5,10 @@ go 1.21
 replace github.com/ethereum-optimism/superchain-registry/superchain => ../superchain
 
 require (
+	github.com/ethereum-optimism/optimism v1.4.2
```
protolambda, line 8: Dependency on monorepo here is fine, since this module doesn't get pulled into the monorepo. Still a bit odd how the `go.mod` was outdated. Is the `validation` module not checked in Go CI?

superchain/extra/genesis-system-configs/sepolia-dev-0/oplabs-devnet-0.json
```diff
@@ -0,0 +1,6 @@
+{
+  "batcherAddr": "0x19cc7073150d9f5888f09e0e9016d2a39667df14",
+  "overhead": "0x00000000000000000000000000000000000000000000000000000000000000bc",
+  "scalar": "0x00000000000000000000000000000000000000000000000000000000000a6fe0",
```
protolambda, line 4: Different from Base, but shouldn't be a problem, and matches the rollup-config contents when I query the sequencer RPC.

protolambda (approved):
The Go changes and `configs` dir look good to me. The genesis system config looks good. And the go.mod / style changes/fixes of the separate `validation` module looks good (although that should really have been caught/fixed in prior CI runs outside of this PR).

validation/go.mod
sebastianst: Apparently not, I really just saved the file and my dev env did the rest here. I then ran `go mod tidy` in both modules to make sure everything's up to date.

superchain/extra/genesis-system-configs/sepolia-dev-0/oplabs-devnet-0.json
sebastianst: Agree, and we'll both change to the same values after Ecotone activation.

PR #72, opened 2024-01-31
test: Add TestGenesisHash (protocol parameter verification)

geoknee:
**Description**

* Adds a new test to validate the genesis block hash against the provided genesis config
* This test generates a new genesis block from (the provided chainconfig + some hardcoded values) and checks it matches the declared genesis hash. Thus (trusting the declared hash is correct) if the check passes we know the config is "correct" according to 
  - the [function](https://github.com/ethereum-optimism/op-geth/blob/daade41d463b4ff332c6ed955603e47dcd25528b/core/superchain.go#L13) which builds the block. 
  - the config read from `superchain` (e.g. genesis block nonce and many other fields)
  - the config baked into the function above (e.g. [`OptimismConfig`](https://github.com/ethereum-optimism/op-geth/blob/d35dd231805c4c05bef761c3403f2ee2a71dc275/params/superchain.go#L79-L83)) 

**Additional context**

* Inspired by [Josh's `op-genesis` tool. ](https://github.com/trianglesphere/op-genesis/blob/d280ff4d6d26529b33c68c4dc9d760d4a069d5fb/genesis_test.go#L175C1-L187C2)

* I think there is an opportunity to move a lot or all of the code looped in from `op-geth` into `superchain`. If we want to do that, will require some coordination between the two repos. Dependency relationships: 
  - `validation` => `op-geth`  (remove)
  -  `validation`  => `superchain` (keep)
  -  `op-geth`  => `superchain` (keep)

**Metadata**

- Towards https://github.com/ethereum-optimism/client-pod/issues/486 (Task 1)


validation/superchain-genesis_test.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	declaredGenesisHash := chainConfig.Genesis.L2.Hash
+
+	// In this step, we call a function from op-geth which utilises
+	// superchain.OPChains, superchain.LoadGenesis, superchain.LoadContractBytecode
+	// to reconstruct a core.Genesis and compute its block hash.
+	// We can then compare that against the declared genesis block hash.
+	// TODO core.LoadOPStackGenesis actually already performs the check we are about to do
+	// TODO core.LoadOPStackGenesis could be moved to superchain
```
geoknee, line 26: Placeholder for further work I would like to do, possibly on this PR, after getting some feedback.

validation/superchain-genesis_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	require.Equal(t, common.Hash(declaredGenesisHash), computedGenesisHash, "chain %d: Genesis block hash must match computed value", chainID)
+	t.Logf("chain %d: Genesis block hash passed validation", chainID)
+}
+
+func TestGenesisHash(t *testing.T) {
+	isExcluded := map[uint64]bool{
+		10:  true,
+		420: true,
+	}
```
geoknee, line 40: I did this instead of copying the special cases here

https://github.com/ethereum-optimism/op-geth/blob/daade41d463b4ff332c6ed955603e47dcd25528b/core/superchain.go#L83-L94

Open to suggestions on alternatives, though.

validation/superchain-genesis_test.go
mslipper: Let's start with getting the tests working, then we can refactor.

mslipper (approved):
As we discussed during our 1:1, we should ideally move the genesis generation loading code + hardcoded genesis hashes into the Registry.

geoknee:
> As we discussed during our 1:1, we should ideally move the genesis generation loading code + hardcoded genesis hashes into the Registry.

Agreed. I'll queue that up for future work.

PR #73, opened 2024-01-31
feat: add `ProtocolVersions` to list of contract addresses

geoknee:
**Description**

* Adds another `contract name: version` entry to the list of declared contract versions (per superchain)
* Adds a test which verifies that the superchain-wide contract `protocol_versions` (address declared in `superchain.yaml`) has the version declared above.

The value of this check is that:

1. The package now exposes the `ProtocolVersions` version under `SuperchainSemver` (along with existing contract versions). 
2. any new superchain being added to the repo will be checked to ensure that the following lines are consistent:

https://github.com/ethereum-optimism/superchain-registry/blob/afa2460347b5278ea2e1b642e915879f57cbfddf/superchain/configs/mainnet/superchain.yaml#L7

https://github.com/ethereum-optimism/superchain-registry/blob/b5cde4101cd1952a04636ea9d1a63771071bb703/superchain/configs/mainnet/semver.yaml#L10

**Additional context**

* I pulled the versions for existing superchains from etherscan
* I have written the test in a way which anticipates more superchain wide contracts being added in future (e.g. `SuperchainConfig`

**Metadata**

- Towards https://github.com/ethereum-optimism/client-pod/issues/485

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,7 +29,7 @@ EOF
 Superchain-wide configuration, like the `ProtocolVersions` contract address, should be configured here when available.
 
 ### Approved contract versions
-Each superchain target should have a `semver.yaml` file in the same directory declaing the approved contract semantic versions for that superchain, e.g: 
+Each superchain target should have a `semver.yaml` file in the same directory declaring the approved contract semantic versions for that superchain, e.g: 
 ```yaml
 l1_cross_domain_messenger: 1.4.0
 l1_erc721_bridge: 1.0.0
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,6 +28,21 @@ EOF
 ```
 Superchain-wide configuration, like the `ProtocolVersions` contract address, should be configured here when available.
 
+### Approved contract versions
+Each superchain target should have a `semver.yaml` file in the same directory declaing the approved contract semantic versions for that superchain, e.g: 
+```yaml
+l1_cross_domain_messenger: 1.4.0
+l1_erc721_bridge: 1.0.0
+l1_standard_bridge: 1.1.0
+l2_output_oracle: 1.3.0
+optimism_mintable_erc20_factory: 1.1.0
+optimism_portal: 1.6.0
+system_config: 1.3.0
+
+# superchain-wide contracts
+protocol_versions: 1.0.0
+```
+
 ### `implementations`
 
 Per superchain a set of canonical implementation deployments, per semver version, is tracked.
--- a/superchain/superchain.go
+++ b/superchain/superchain.go
@@ -255,6 +255,8 @@ func (c ContractVersions) VersionFor(contractName string) (string, error) {
 		version = c.OptimismPortal
 	case "SystemConfig":
 		version = c.SystemConfig
+	case "ProtocolVersions":
+		version = c.ProtocolVersions
 	default:
 		return "", errors.New("no such contract name")
 	}
--- a/superchain/superchain.go
+++ b/superchain/superchain.go
@@ -232,6 +232,8 @@ type ContractVersions struct {
 	OptimismMintableERC20Factory string `yaml:"optimism_mintable_erc20_factory"`
 	OptimismPortal               string `yaml:"optimism_portal"`
 	SystemConfig                 string `yaml:"system_config"`
+	// Superchain-wide contracts:
+	ProtocolVersions string `yaml:"protocol_versions"`
 }
 
 // VersionFor returns the version for the supplied contract name, if it exits
--- a/validation/superchain-version_test.go
+++ b/validation/superchain-version_test.go
@@ -20,35 +20,40 @@ var isSemverAcceptable = func(desired, actual string) bool {
 	return desired == actual
 }
 
+func checkSemverForContract(t *testing.T, contractName string, contractAddress *Address, client *ethclient.Client, desiredSemver string) {
+
+	actualSemver, err := getVersionWithRetries(context.Background(), common.Address(*contractAddress), client)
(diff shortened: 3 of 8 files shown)


superchain/extra/addresses/mainnet/op.json
```diff
@@ -11,5 +11,6 @@
   "ProxyAdminOwner": "0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A",
   "SystemConfigOwner": "0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A",
   "Guardian": "0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A",
-  "Challenger": "0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A"
+  "Challenger": "0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A",
+  "ProtocolVersionsProxy":"0x8062AbC286f5e7D9428a0Ccb9AbD71e50d93b935"
```
mslipper, line 15: The actual address for both the protocol versions proxy and the superchain config proxy should go into the sueprchain.yaml file for each superchain target since these contracts apply to every chain that belongs to the target.

validation/superchain-version_test.go
```diff
@@ -16,9 +16,52 @@ import (
...
+	return desired == actual
+}
+
+// TestSuperchainWideContractVersions will check that
+//   - for each superchain
+//   - for each declared superchain-wide contract "Foo" : version entry in the corresponding superchain's semver.yaml
+//   - the superchain has a contract Foo deployed (at an address declared in superchain.yaml) at the same version
+//
+// Actual semvers are
+// read from the L1 chain RPC provider for the superchain in question.
```
geoknee, line 29: I'm not sure if we really want to keep this godoc here (same for the other test in this file). 

* This function will not be imported by another package, so no "API" to document. 
* We risk it getting out of sync with the implementation.
* The code should be clear enough to document itself.
* godoc references underlying source data files (when in truth it operates only on Go types)

superchain/superchain.go
```diff
@@ -232,6 +232,8 @@ type ContractVersions struct {
 	OptimismMintableERC20Factory string `yaml:"optimism_mintable_erc20_factory"`
 	OptimismPortal               string `yaml:"optimism_portal"`
 	SystemConfig                 string `yaml:"system_config"`
+	// Superchain-wide contracts:
+	ProtocolVersions string `yaml:"protocol_versions"`
```
mslipper, line 236: Where does this get populated from? Have we already implemented the glue code to populate this from the `superchain.yaml` files?

validation/superchain-version_test.go
```diff
@@ -16,9 +16,52 @@ import (
...
+func TestSuperchainWideContractVersions(t *testing.T) {
+	for name, superchain := range Superchains {
+		rpcEndpoint := superchain.Config.L1.PublicRPC
+		require.NotEmpty(t, rpcEndpoint)
+
+		client, err := ethclient.Dial(rpcEndpoint)
+		require.NoErrorf(t, err, "could not dial rpc endpoint %s", rpcEndpoint)
+
+		contractNames := []string{
```
mslipper, line 38: Do we have to manually specify the contract names like this? I wonder if there's a way for us to combine this test + the one below it, since they both achieve roughly the same thing.

mslipper (changes requested):
Left some comments with open questions.

superchain/superchain.go
geoknee: This is the "declared version", and gets populated from the `semver.yaml` files, leaning on existing code which unmarshals into this struct (and new line in those files I have populated on this PR). 

The corresponding address is already being loaded from the `superchain.yaml` file and accessible from the data exported from the `superchain` Go module. The new test I added "knows" to grab that address to make the rpc request and get the "actual" version.

Does that make sense?

protolambda (commented):
The `semver.yaml` should be described in the superchain-target configuration section in `CONTRIBUTING.md`. Can you document the functionality?

validation/superchain-version_test.go
geoknee: [4b3a478](https://github.com/ethereum-optimism/superchain-registry/pull/73/commits/4b3a4784f1aef72188492c1657c618c2f95f23b5)

validation/superchain-version_test.go
geoknee: [1e52cf0](https://github.com/ethereum-optimism/superchain-registry/pull/73/commits/1e52cf0079b085ff71601425a3fafad83410db58)

geoknee:
> The `semver.yaml` should be described in the superchain-target configuration section in `CONTRIBUTING.md`. Can you document the functionality?

Good point [d16d554](https://github.com/ethereum-optimism/superchain-registry/pull/73/commits/d16d554b8ffef580b0b034916b0f5c555c860309) .

superchain/superchain.go
mslipper: sgtm

mslipper (approved):

maurelian (approved):

CONTRIBUTING.md
```diff
@@ -28,6 +28,21 @@ EOF
 ```
 Superchain-wide configuration, like the `ProtocolVersions` contract address, should be configured here when available.
 
+### Approved contract versions
+Each superchain target should have a `semver.yaml` file in the same directory declaing the approved contract semantic versions for that superchain, e.g:
```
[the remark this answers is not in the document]
geoknee: [1b93e84](https://github.com/ethereum-optimism/superchain-registry/pull/73/commits/1b93e84ddce5529ec6a5c66d84f960043dfcd31e)

PR #81, opened 2024-02-06
codeowners: Let go-reviewers own superchain configs and validation

sebastianst:


protolambda (approved):
Code-owners change LGTM.

The Go mod change seems unrelated, but valid. `go mod tidy` and a diff-check should be added in CI, like we have in the monorepo.

PR #83, opened 2024-02-07
validation: check `SuperchainConfig` version

geoknee:
This is a superchain-wide contract.

Some targets do not have it deployed or do not specify an address. Therefore we cannot specify anything in the semver.yaml file. Therefore the semver.yaml would fail a sanity check which currently enforces that no field is empty.

My usual approach of skipping certain superchains / chains in the test won't work here without some deeper changes, because there failing check runs when the package is built/imported. 

Options: 
1. ~Store the `SuperchainConfig` version in a different place and allow it to be null/empty~
2. Relax the requirements that specified versions cannot be empty. We can allow empty versions at init time, and check them at validation time. 
3. ~Use a magic value instead of `""` for unspecified chains~

Closes https://github.com/ethereum-optimism/client-pod/issues/485

### DIFF
--- a/superchain/superchain.go
+++ b/superchain/superchain.go
@@ -235,6 +235,7 @@ type ContractVersions struct {
 	SystemConfig                 string `yaml:"system_config"`
 	// Superchain-wide contracts:
 	ProtocolVersions string `yaml:"protocol_versions"`
+	SuperchainConfig string `yaml:"superchain_config,omitempty"`
 }
 
 // VersionFor returns the version for the supplied contract name, if it exits
@@ -258,6 +259,8 @@ func (c ContractVersions) VersionFor(contractName string) (string, error) {
 		version = c.SystemConfig
 	case "ProtocolVersions":
 		version = c.ProtocolVersions
+	case "SuperchainConfig":
+		version = c.SuperchainConfig
 	default:
 		return "", errors.New("no such contract name")
 	}
--- a/superchain/superchain.go
+++ b/superchain/superchain.go
@@ -281,7 +281,7 @@ func (c ContractVersions) Check() error {
 			return fmt.Errorf("invalid type for field %s", val.Type().Field(i).Name)
 		}
 		if str == "" {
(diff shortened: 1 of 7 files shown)


superchain/superchain.go
```diff
@@ -278,7 +281,7 @@ func (c ContractVersions) Check() error {
 			return fmt.Errorf("invalid type for field %s", val.Type().Field(i).Name)
 		}
 		if str == "" {
-			return fmt.Errorf("empty version for field %s", val.Type().Field(i).Name)
+			continue // we allow empty strings and rely on tests to assert (or except) a nonempty version
```
geoknee, line 284: Here be the most contentious change. See PR description for justification.

validation/superchain-version_test.go
```diff
@@ -32,6 +33,21 @@ func TestSuperchainWideContractVersions(t *testing.T) {
 		desiredSemver, err := SuperchainSemver[superchain.Superchain].VersionFor("ProtocolVersions")
 		require.NoError(t, err)
 		checkSemverForContract(t, "ProtocolVersions", superchain.Config.ProtocolVersionsAddr, client, desiredSemver)
+
+		isExcludedFromSuperchainConfigCheck := map[string]bool{
+			"Goerli":       true, // no version specified
+			"Goerli Dev 0": true, // no version specified
+			"Mainnet":      true, // no version specified
```
geoknee, line 40: Any superchain not in this exclusion list will fail the check if they fail to specify a SuperchainConfig semver.

maurelian (approved):

maurelian (approved):

superchain/superchain.go
Inphi, line 284: Makes sense. Though we'll have to be careful in the future as we version new superchain contracts. I wonder if this behavior can be configured by the caller? Say something like `Check(skipEmptyVersions)`.

Inphi (approved):
Looks good. Just have one comment on forward compatibility.

PR #84, opened 2024-02-08
wip: start performing bytecode checks

geoknee:
Towards https://github.com/ethereum-optimism/client-pod/issues/488


validation/expected-bytecode.go
```diff
@@ -0,0 +1,9 @@
...
...03821115611dd357611dd3611cfb565b01949350505050565b7f4e487b7100000000000000000000000000000000000000000000000000000000600052600160045260246000fd5b600082821015611e1d57611e1d611cfb565b500390565b73ffffffffffffffffffffffffffffffffffffffff8616815284602082015267ffffffffffffffff84166040820152821515606082015260a060808201526000611e6f60a0830184611a38565b979650505050505050565b60007fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff8203611eab57611eab611cfb565b5060010190565b600082611ec157611ec1611d5a565b500490565b7f4e487b7100000000000000000000000000000000000000000000000000000000600052604160045260246000fd5b600082611f0457611f04611d5a565b500690565b60008219821115611f1c57611f1c611cfb565b500190565b7f4e487b7100000000000000000000000000000000000000000000000000000000600052603260045260246000fd5b600060208284031215611f6257600080fd5b8151611a958161191d565b600073ffffffffffffffffffffffffffffffffffffffff808716835280861660208401525060806040830152611fa66080830185611a38565b905082606083015295945050505050565b868152600073ffffffffffffffffffffffffffffffffffffffff808816602084015280871660408401525084606083015283608083015260c060a083015261200260c0830184611a38565b9897505050505050505056fea164736f6c634300080f000a",
```
geoknee, line 8: This will add up to quite a lot of data (approx 100KB per superchain target). For this reason, I haven't yet populated the remaining fields, and have housed this in the `validation` module instead of the `superchain` module (so we don't weigh down downstream packages). 

Alternatives to this approach:
1. Load this bytecode dynamically over RPC when the test runs (that's all I did to get this data anyway)
2. Store a hash of the bytecode and use that for comparison
3. Import this [bytecode from the monorepo ](https://github.com/ethereum-optimism/optimism/blob/a22c0f9bb678323e448de997c9db655a3b93a4ad/packages/contracts-bedrock/deployments/mainnet/L1CrossDomainMessenger.json#L425)(I don't think this works for all of the targets).

PR #90, opened 2024-02-12
sepolia: update implementation addresses following MCP L1 upgrade

0xfuturistic:
Update implementation addresses following MCP L1 upgrade on Sepolia


superchain/implementations/networks/sepolia.yaml
mds1, line 1: Is this folder consumed by any tooling or is this just for reference?

superchain/implementations/networks/sepolia.yaml
```diff
@@ -1,14 +1,14 @@
 l1_cross_domain_messenger:
-  1.5.1: 0xDAF83496D1E2bF53EB60F29Ca141189118db1B0A
+  2.3.0: 0xD3494713A5cfaD3F5359379DfA074E2Ac8C6Fd65
```
mds1, line 2: Prior to MCP, was this file documenting only OP Sepolia implementations addresses? And post MCP it now documents the implementation addresses for all standard chains?

superchain/implementations/networks/sepolia.yaml
0xfuturistic: I believe just for reference, but tagging @tynes to make sure

superchain/implementations/networks/sepolia.yaml
mds1: I see in https://github.com/ethereum-optimism/superchain-registry/blob/main/CONTRIBUTING.md#implementations that it's supposed to be a canonical set of implementation deployments. I don't know what exactly that meant pre-MCP, but it does make sense now

0xfuturistic: should we resolve this conversation then?

tynes: This repo was created with the intention of MCP but the code red bug resulted in a contracts release before MCP. In theory all contract addresses in this repo should support MCP, there are downstream risks on how it is used if that is not the case, it would just require a bit more attention to the way the tooling is used if pre-MCP contracts are included in this registry

superchain/implementations/networks/sepolia.yaml
tynes: This is consumed by tooling - the [semver.yaml])https://github.com/ethereum-optimism/superchain-registry/blob/fdd0d21715eea9fdfd48eac191e23d1249c59899/superchain/configs/mainnet/semver.yaml) matches against these implementations

superchain/implementations/networks/sepolia.yaml
tynes: The idea is that we can put the create2 addresses used on all chains in [here](https://github.com/ethereum-optimism/superchain-registry/blob/fdd0d21715eea9fdfd48eac191e23d1249c59899/superchain/implementations/implementations.yaml) and overrides in the network specific directories

mds1: That makes sense to me, thank you!

superchain/implementations/networks/sepolia.yaml
mds1: Ah I see that now, thanks https://github.com/ethereum-optimism/superchain-registry/blob/fdd0d21715eea9fdfd48eac191e23d1249c59899/superchain/init.go#L27-L30

mds1 (approved):

tynes (approved):

PR #91, opened 2024-02-12
mainnet: Set Ecotone activation time

sebastianst:
unix timestamp: 1710781201
Mon Mar 18 17:00:01 UTC 2024

Fixes https://github.com/ethereum-optimism/protocol-quest/issues/80


trianglesphere (commented):
Time LGTM, my only concern is merging this in so early

trianglesphere (approved):

PR #92, opened 2024-02-13
ci: add nightly workflow for `golang-test` and `check-security-configs` jobs

### LINKED ISSUE #89: Run `CheckSecurityConfigs` on a scheduled job
From https://github.com/ethereum-optimism/superchain-registry/issues/87

> Some changes were made on chain which should be reflected in the registry. https://app.circleci.com/pipelines/github/ethereum-optimism/superchain-registry/337/workflows/0c87fd15-22a3-4718-a198-87de3904de84/jobs/1334
> 
> I think we should consider running the CI checks nightly, since I only discovered this CI failure on a pull request with unrelated changes.

These checks should be pretty fast and cheap to execute, so we can execute them on a scheduled job fairly frequently to catch issues faster. We should also make sure failures of this scheduled job send an alert

geoknee:
- Fixes #89

I'm also running the `golang-tests` job nightly in addition to the `check-security-configs` job. Either of these can break due to changes external to the repo, so we will get notified within 24hrs if that happens.

I've manually verified the slack integration is working.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -114,7 +114,7 @@ jobs:
       - notify-failures-on-main
 
 workflows:
-  nightly:
+  hourly:
     jobs:
       - golang-test:
           context:
@@ -124,7 +124,7 @@ workflows:
             - slack
     triggers:
       - schedule: 
-          cron: "0 0 * * *"
+          cron: "0 * * * *"
           filters:
             branches:
               only:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,26 @@ version: 2.1
 
 orbs:
   go: circleci/go@1.8.0
+  slack: circleci/slack@4.10.1
 
+commands:
+  notify-failures-on-main:
+    description: "Notify Slack"
+    parameters:
+      channel:
+        type: string
+        default: C03N11M0BBN
+    steps:
+      - slack/notify:
+          channel: << parameters.channel >>
+          event: fail
+          template: basic_fail_1
+          branch_pattern: main
+      - slack/notify: # DELETEME for temporary testing only
+          channel: << parameters.channel >>
+          event: pass 
+          template: basic_fail_1
+          branch_pattern: gk/nightly-ci-run
 jobs:
   golang-lint:
     executor:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -106,13 +111,33 @@ jobs:
...
+      - check-security-configs:
+          context:
+            - slack
+    triggers:
+      - schedule: 
+          cron: "0 0 * * *"
+          filters:
+            branches:
+              only:
+                - main
```
mds1, line 131: In the monorepo we do this by defining [triggers](https://app.circleci.com/settings/project/github/ethereum-optimism/optimism/triggers) and using them [like this](https://github.com/ethereum-optimism/optimism/blob/39bc6f433c3344079dc4933688ad2e72064aec99/.circleci/config.yml#L2025-L2031).

I'm not sure if there's any reason to prefer that over the cron approach though.

.circleci/config.yml
```diff
@@ -106,13 +111,33 @@ jobs:
...
+    jobs:
+      - golang-test:
+          context:
+            - slack
+      - check-security-configs:
+          context:
+            - slack
+    triggers:
+      - schedule: 
+          cron: "0 0 * * *"
```
mds1, line 127: Since this job is lightweight, should we run it more frequently to catch issues/changes faster? It shouldn't fail often, but a 24h+ delay between an incorrect config and resolving it feels long. We have a scheduled job in the monorepo that runs hourly, so since there's precedent hourly would be my suggestion, though I don't feel too strongly here

.circleci/config.yml
geoknee: Yes I had spotted that different approach. I believe circle tried to [deprecate scheduled workflows in favour of scheduled pipelines, but decided not ](https://circleci.com/docs/migrate-scheduled-workflows-to-scheduled-pipelines/)to in the end (yet). 

I don't have a very strong opinion either way, other than that this way is already done. Another thing I like is that all the config is in one place next to the source code and I don't need to go hunting in some other dashboard to understand what is going to happen.

.circleci/config.yml
geoknee: Good point, I'm fine with running it hourly.

.circleci/config.yml
mds1: Cool I'm onboard with leaving this config as-is, will resolve this comment

.circleci/config.yml
geoknee: [f0e6afa](https://github.com/ethereum-optimism/superchain-registry/pull/92/commits/f0e6afa9a480067d68aba61bf2d58ec44299b0d9)

mds1 (approved):

PR #94, opened 2024-02-14
refactor: Use a single consistent location for `SystemConfigProxy` address

geoknee:
**Description**

* Moves the above contract address out of the chain config yaml file and into the extra / addresses file. 
* Removes exported struct field from ChainConfig object (GO module)
* Updates the contributing guide accordingly. 
* Unskips a test (it was skipped due to incomplete data, caused by inconsistent location for that data)

This will likely also be important for bytecode checking #84


sebastianst (commented):
There's a [field in the Go struct](https://github.com/ethereum-optimism/superchain-registry/blob/20fe2a67a623f456a0c6a3d7cd95646def906ed7/superchain/superchain.go#L52) for it. ~I did a quick check in the monorepo and it seems we never use the `SystemConfigAddr` field, wondering why we added it in the first place.~

~If we're sure we don't need it, then the field should also better be removed from the struct.~

Edit: We ofc use it in many different places. My quick check was just a very bad check 😅

ajsutton:
Removing myself from the reviewer list because a) I don't have enough context for this change b) I don't think my review counts anyway and c) Josh and Seb are already doing a better job of reviewing than I would. :)

geoknee:
Good point @trianglesphere @sebastianst . I found this reference to the struct member:

https://github.com/ethereum-optimism/optimism/blob/21c5fcbaa10b9696aab311b92aab3dacd7ccd3f2/op-node/rollup/superchain.go#L117

I see a couple of ways forward:

Option A:
1. Remove the field as Seb suggests.
2. Patch the code in the monorepo
3. Coordinate the two changes
 
Option B:
1. Keep the field in place, and have it populated from the new location when the `superchain` module inits. Init code can panic if no address is found.
2. No code changes in monorepo (backwards compatible change)
3. Still get improved consistency at the repo/file system level, but not as much improvement at the Go module interface level.

Option C. 
1. Make the `chain.yaml` file the single location for the `SystemConfigProxy` address (resolve inconsistency in the other direction)

geoknee:
> How would Option B look like? I'm not sure you could read the JSON file from a very different location during init of the Go module. The approach right now is to embed yaml data into the Go module, and I think this is quite nice and less error prone, so we should stick with that.

The Go module also embeds the data from those JSON files
https://github.com/ethereum-optimism/superchain-registry/blob/d33b16ab81c7d86aa7af1ee81aa202d57c9c1e26/superchain/superchain.go#L541-L544

So I think both Options A and B can work in principle, because it's possible to just rewire references to the address in question (this change doesn't remove the data from the Go module, it just places moves it around). 

So the "patch" in Option A would look like this:

https://github.com/ethereum-optimism/optimism/commit/00c3b173fd58a62cfd091b90347f59dccac20508

And in Option B there are no changes in the monorepo, the patching / rewiwing is done in `superchain`.

Having said all of that, I don't really like Option A or B, they seem kind of messy. So I can work on Option C if it looks like a consensus is forming around that. In general I would prefer to avoid duplicating data even with a check in CI -- the two distinct locations make a bit more sense now I understand why this contract address is "more special" than the others in the JSON files. 

> Who consumes the addresses in superchain/extra/addresses/addresses.json? NPM packages? Could we read the values at build time from the yaml files instead?

Yes I believe NPM packages. Your suggestion sounds OK to me, although it does inch be back towards Option A a little bit.

sebastianst:
I see we're already embedding the addresses in Go. In that case nevermind what I said about Option A 😅 Looks like it's a valid option then.

Since we already opened the flood gates of depending on the extra addresses in our node setup in the monorepo, we might as well remove the `system_config_addr` from the yaml and just use the `Addresses`. From what I can see, this would be more consistent after all.

BTW We already load the addresses in [line 59](https://github.com/ethereum-optimism/optimism/commit/00c3b173fd58a62cfd091b90347f59dccac20508#diff-1cc047e18c4eae6127d5aeb57c3d849595b2e003e5a4f72a194a3c895eb698d8L59) of your patch, so you could reuse that for the `SystemConfigProxy` address.

Checking for usages of the chain config `SystemConfigAddr` field in the monorepo, there are a few more, e.g. several in `op-chain-ops/upgrades/l1.go`. You can check how easy it is to use the addresses from the address list instead.

I'm starting to wonder more generally about how much sense the semantics makes between the `configs` and `extra` folders. The addresses in `extra/addresses` are necessary to the configuration of the op-node after all, so they are not really "extra". Now there are some addresses left in `configs`, like `ProtocolVersions`, batcher inbox address and system config, while others are in `extra/addresses`. Is there something I'm overlooking why we have that semantic split? Couldn't we move all the addresses into `addresses`, and maybe lift the whole `addresses` folder out of `extra`?

geoknee:
> I see we're already embedding the addresses in Go. In that case nevermind what I said about Option A 😅 Looks like it's a valid option then.
> 
> Since we already opened the flood gates of depending on the extra addresses in our node setup in the monorepo, we might as well remove the `system_config_addr` from the yaml and just use the `Addresses`. From what I can see, this would be more consistent after all.
> 
> BTW We already load the addresses in [line 59](https://github.com/ethereum-optimism/optimism/commit/00c3b173fd58a62cfd091b90347f59dccac20508#diff-1cc047e18c4eae6127d5aeb57c3d849595b2e003e5a4f72a194a3c895eb698d8L59) of your patch, so you could reuse that for the `SystemConfigProxy` address.
> 
> Checking for usages of the chain config `SystemConfigAddr` field in the monorepo, there are a few more, e.g. several in `op-chain-ops/upgrades/l1.go`. You can check how easy it is to use the addresses from the address list instead.

OK, thanks for the tips! I'll get a draft PR up in the monorepo that does the patching in a neat + complete way, and then we can execute Option A. 

> I'm starting to wonder more generally about how much sense the semantics makes between the `configs` and `extra` folders. The addresses in `extra/addresses` are necessary to the configuration of the op-node after all, so they are not really "extra". Now there are some addresses left in `configs`, like `ProtocolVersions`, batcher inbox address and system config, while others are in `extra/addresses`. Is there something I'm overlooking why we have that semantic split? Couldn't we move all the addresses into `addresses`, and maybe lift the whole `addresses` folder out of `extra`? (not as part of this PR ;) )

I wonder this too. I would be in favour of what you are suggesting unless someone knows a good reason not to.

validation/superchain-version_test.go
```diff
@@ -42,21 +42,20 @@ func TestSuperchainWideContractVersions(t *testing.T) {
 
 func TestContractVersions(t *testing.T) {
 	isExcluded := map[uint64]bool{
-		291:          true,
```
geoknee, line 45: There are more tests which can be unskipped now, for example the `validation/resource-config_test.go` test.

sebastianst:
@geoknee did https://github.com/ethereum-optimism/optimism/pull/9585#pullrequestreview-1891776340 and looks like a pretty neat change, I think we can go ahead here with Option A.

sebastianst (approved):
LGTM, waiting for rebase.

PR #97, opened 2024-02-21
WIP: add hardfork timestamp overrides

geoknee:
**Description**

Towards https://github.com/ethereum-optimism/client-pod/issues/605

### DIFF
--- a/superchain/superchain.go
+++ b/superchain/superchain.go
@@ -77,19 +77,19 @@ type ChainConfig struct {
 func (c *ChainConfig) replaceMissingOverridesWithDefaults(s Superchain) {
 
 	if c.CanyonTime == nil {
-		c.CanyonTime = s.Config.hardForkDefaults.canyonTime
+		c.CanyonTime = s.Config.canyonTime
 	}
 	if c.DeltaTime == nil {
-		c.DeltaTime = s.Config.hardForkDefaults.deltaTime
+		c.DeltaTime = s.Config.deltaTime
 	}
 	if c.EcotoneTime == nil {
-		c.EcotoneTime = s.Config.hardForkDefaults.ecotoneTime
+		c.EcotoneTime = s.Config.ecotoneTime
 	}
 	if c.FjordTime == nil {
-		c.FjordTime = s.Config.hardForkDefaults.fjordTime
+		c.FjordTime = s.Config.fjordTime
 	}
 	if c.RegolithTime == nil {
-		c.RegolithTime = s.Config.hardForkDefaults.regolithTime
+		c.RegolithTime = s.Config.regolithTime
 	}
 
 }
@@ -474,23 +474,6 @@ type hardForkConfigurationPrivate struct {
 	regolithTime *uint64 `yaml:"regolith_time,omitempty"`
 }
 
-func (h *hardForkConfigurationPrivate) UnmarshalYAML(unmarshal func(interface{}) error) error {
-	temp := HardForkConfiguration{}
-
-	err := unmarshal(&temp)
-	if err != nil {
-		return err
-	}
-
-	h.canyonTime = temp.CanyonTime
-	h.deltaTime = temp.DeltaTime
-	h.ecotoneTime = temp.EcotoneTime
-	h.fjordTime = temp.FjordTime
-	h.regolithTime = temp.RegolithTime
-
-	return nil
-}
-
 type SuperchainConfig struct {
 	Name string           `yaml:"name"`
 	L1   SuperchainL1Info `yaml:"l1"`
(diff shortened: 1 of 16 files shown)


geoknee:
The dependency graph is causing a bit of a pain here.
`validation`=>`op-geth`=>`superchain`, but both `validation` and `superchain` are in the same (this) repo. I think this will force me to remove the dependency `validation`=>`op-geth`, which was probably in the scope of the linked ticket anyway.

superchain/superchain.go
```diff
@@ -436,10 +461,10 @@ type SuperchainConfig struct {
...
 
 	// Hardfork Configuration
-	CanyonTime  *uint64 `yaml:"canyon_time,omitempty"`
-	DeltaTime   *uint64 `yaml:"delta_time,omitempty"`
-	EcotoneTime *uint64 `yaml:"ecotone_time,omitempty"`
-	FjordTime   *uint64 `yaml:"fjord_time,omitempty"`
+	canyonTime  *uint64 `yaml:"canyon_time,omitempty"`
+	deltaTime   *uint64 `yaml:"delta_time,omitempty"`
+	ecotoneTime *uint64 `yaml:"ecotone_time,omitempty"`
+	fjordTime   *uint64 `yaml:"fjord_time,omitempty"`
```
geoknee, line 467: Making these private so that they can still be specified and defaulted to, but so that consumers are not tempted to try and read them (they should read from the specific chain which will have the values overridden if appropriate).

sebastianst: I believe yaml, json etc marshaling code needs fields to be public (because of reflection).

geoknee: Good point. I think we can get around that by writing a custom `Unmarshal` method. Worth it IMO to avoid confusion.

superchain/superchain.go
```diff
@@ -60,6 +60,31 @@ type ChainConfig struct {
...
+	CanyonTime  *uint64 `yaml:"canyon_time,omitempty"`
+	DeltaTime   *uint64 `yaml:"delta_time,omitempty"`
+	EcotoneTime *uint64 `yaml:"ecotone_time,omitempty"`
+	FjordTime   *uint64 `yaml:"fjord_time,omitempty"`
+}
+
+// replaceMissingOverridesWithDefaults overwrites each unspecified hardfork activation time override
+// with the superchain default.
+func (c *ChainConfig) replaceMissingOverridesWithDefaults(s Superchain) {
+
```
sebastianst, line 74: nit: linter probably won't accept fist and last empty lines in function. The linter run currently errors.

superchain/superchain.go
```diff
@@ -60,6 +60,31 @@ type ChainConfig struct {
...
+	}
+	if c.DeltaTime == nil {
+		c.DeltaTime = s.Config.deltaTime
+	}
+	if c.EcotoneTime == nil {
+		c.EcotoneTime = s.Config.ecotoneTime
+	}
+	if c.FjordTime == nil {
+		c.FjordTime = s.Config.fjordTime
+	}
```
sebastianst, line 86: It is easy to miss adding new fields here. Could we add a test that fails if we add a field to the superchain config, but not to the chain config, and if present in both checks that these overrides are applied?

superchain/superchain.go
sebastianst: We should also add a comment above that these are potentially overridden by chain-specific values.

sebastianst (commented):
Couple comments. Would be great to already transfer all the custom timestamps out of op-node and op-geth into the chain configs within the scope of this PR.

superchain/superchain.go
sebastianst: Good idea. We could even create a struct like `ForkTimes` with all the timestamps, embed this into the chain config and reuse it for marshaling here.

superchain/superchain.go
geoknee: Be nice if my editor just fixed this for me automatically. Does yours?

sebastianst: yes, always applies gofmt on save :)

superchain/superchain.go
geoknee: Don't we want to allow chains to omit fields in the chain config, though? If we do not, then there's no point in having the fields in the superchain config and we can simplify things a lot.

geoknee: Hesitate to suggest it, but we could roll something with reflection to ensure we didn't forget to add all necessary fields here.

superchain/superchain.go
geoknee: I don't think the linter was actually upset by this. If we want it to be, it may require some configuration change.

geoknee:
In favour of #101

PR #98, opened 2024-02-21
Fix failing sepolia gas price oracle checks

geoknee:
* Ecotone was activated on Sepolia, triggering a different pathway in the tests.
* That pathway did not have a "ground truth" defined.
* The test panicked with an unhelpful message

This PR:
* updates the test to check explicitly for the existence of a ground truth, causing a fatal error with useful message if there is none
* adds in a ground truth for sepolia / ecotone
* some additional driveby changes tidying up / renaming things.


ajsutton (approved):
LGTM.

PR #101, opened 2024-02-23
Add per-chain hardfork time overrides

geoknee:
**Description**
Adds per-chain hardfork time overrides.

* Existing superchain-wide hardfork times remain in the `yaml` files, but are nested under a private struct member to prevent Go consumers accessing them (breaking change for `op-geth` and `monorepo`)
* The hardfork times are instead available in each chains config
* when the module inits, the hardfork times are hydrated and default back to the superchain wide values if no override exists
* the breaking change breaks the `validation` module in this repo because it depends on `op-geth`. Therefore I pointed `validation` at a fixed version of `op-geth` which itself points back to the new version of `superchain` 😵 
* Also updated monorepo to cope with breaking change (also need to update to fixed version of op-geth)
* Bit of a challenge to coordinate merging the changes in the three repos. 

**Tests**

* couple of "spot check" tests for the overrides
* test for new unmarshalling code

**Additional context**
Coupled PRS in the other repos removing overrides:
https://github.com/ethereum-optimism/op-geth/pull/252
https://github.com/ethereum-optimism/optimism/pull/9642

**Metadata**

Towards https://github.com/ethereum-optimism/client-pod/issues/605

### DIFF
--- a/superchain/configs/goerli-dev-0/superchain.yaml
+++ b/superchain/configs/goerli-dev-0/superchain.yaml
@@ -6,7 +6,7 @@ l1:
 
 protocol_versions_addr: "0x79ADD5713B383DAa0a138d3C4780C7A1804a8090"
 
-regolith_time: 0 # genesis, overriden by individual chains
-canyon_time: 1698436800 # October 27, 2023 8:00:00 PM UTC
-delta_time: 1701993600 # December 8, 2023 12:00:00 AM UTC, i.e. midnight Dec 7 to Dec 8
-ecotone_time: 1706126400 # Wed Jan 24 20:00:00 UTC 2024
+regolith_time: 0         # Thu  1 Jan 1970 00:00:00 UTC genesis, overriden by individual chains
+canyon_time: 1698436800  # Fri 27 Oct 2023 20:00:00 UTC
+delta_time: 1701993600   # Fri  8 Dec 2023 00:00:00 UTC
+ecotone_time: 1706126400 # Wed 24 Jan 2024 20:00:00 UTC
--- a/superchain/configs/goerli-dev-0/superchain.yaml
+++ b/superchain/configs/goerli-dev-0/superchain.yaml
@@ -6,7 +6,6 @@ l1:
 
 protocol_versions_addr: "0x79ADD5713B383DAa0a138d3C4780C7A1804a8090"
 
-regolith_time: 0         # Thu  1 Jan 1970 00:00:00 UTC genesis, overriden by individual chains
 canyon_time: 1698436800  # Fri 27 Oct 2023 20:00:00 UTC
 delta_time: 1701993600   # Fri  8 Dec 2023 00:00:00 UTC
 ecotone_time: 1706126400 # Wed 24 Jan 2024 20:00:00 UTC
--- a/superchain/configs/goerli-dev-0/superchain.yaml
+++ b/superchain/configs/goerli-dev-0/superchain.yaml
@@ -5,6 +5,8 @@ l1:
   explorer: https://goerli.etherscan.io
 
 protocol_versions_addr: "0x79ADD5713B383DAa0a138d3C4780C7A1804a8090"
+
+regolith_time: 0 # genesis, overriden by individual chains
 canyon_time: 1698436800 # October 27, 2023 8:00:00 PM UTC
 delta_time: 1701993600 # December 8, 2023 12:00:00 AM UTC, i.e. midnight Dec 7 to Dec 8
 ecotone_time: 1706126400 # Wed Jan 24 20:00:00 UTC 2024
--- a/superchain/configs/goerli/base.yaml
+++ b/superchain/configs/goerli/base.yaml
@@ -15,3 +15,5 @@ genesis:
     hash: "0xa3ab140f15ea7f7443a4702da64c10314eb04d488e72974e02e2d728096b4f76"
     number: 0
   l2_time: 1675193616
+
+regolith_time: 1683219600 # May 4, 2023 @ 5:00:00 pm UTC
--- a/superchain/configs/goerli/base.yaml
+++ b/superchain/configs/goerli/base.yaml
@@ -15,5 +15,3 @@ genesis:
     hash: "0xa3ab140f15ea7f7443a4702da64c10314eb04d488e72974e02e2d728096b4f76"
     number: 0
   l2_time: 1675193616 # Tue 31 Jan 2023 19:33:36 UTC
-
-regolith_time: 1683219600 # Thu  4 May 2023 17:00:00 UTC
--- a/superchain/configs/goerli/base.yaml
+++ b/superchain/configs/goerli/base.yaml
@@ -14,6 +14,6 @@ genesis:
   l2:
     hash: "0xa3ab140f15ea7f7443a4702da64c10314eb04d488e72974e02e2d728096b4f76"
     number: 0
-  l2_time: 1675193616
+  l2_time: 1675193616 # Tue 31 Jan 2023 19:33:36 UTC
 
-regolith_time: 1683219600 # May 4, 2023 @ 5:00:00 pm UTC
+regolith_time: 1683219600 # Thu  4 May 2023 17:00:00 UTC
--- a/superchain/configs/goerli/superchain.yaml
+++ b/superchain/configs/goerli/superchain.yaml
@@ -6,7 +6,7 @@ l1:
 
 protocol_versions_addr: "0x0C24F5098774aA366827D667494e9F889f7cFc08"
 
-regolith_time: 0 # genesis, overriden by individual chains
-canyon_time: 1699981200 # Tue Nov 14 17:00:00 UTC 2023
-delta_time: 1703116800  # Thu Dec 21 00:00:00 UTC 2023
-ecotone_time: 1707238800 # Tue Feb 6 17:00:00 UTC 2024
+regolith_time: 0         # Thu  1 Jan 1970 00:00:00 UTC genesis, overriden by individual chains
+canyon_time: 1699981200  # Tue 14 Nov 2023 17:00:00 UTC
+delta_time: 1703116800   # Thu 21 Dec 2023 00:00:00 UTC
+ecotone_time: 1707238800 # Tue  6 Feb 2024 17:00:00 UTC
--- a/superchain/configs/goerli/superchain.yaml
+++ b/superchain/configs/goerli/superchain.yaml
@@ -6,7 +6,6 @@ l1:
 
 protocol_versions_addr: "0x0C24F5098774aA366827D667494e9F889f7cFc08"
 
-regolith_time: 0         # Thu  1 Jan 1970 00:00:00 UTC genesis, overriden by individual chains
 canyon_time: 1699981200  # Tue 14 Nov 2023 17:00:00 UTC
 delta_time: 1703116800   # Thu 21 Dec 2023 00:00:00 UTC
 ecotone_time: 1707238800 # Tue  6 Feb 2024 17:00:00 UTC
--- a/superchain/configs/goerli/superchain.yaml
+++ b/superchain/configs/goerli/superchain.yaml
@@ -5,6 +5,8 @@ l1:
   explorer: https://goerli.etherscan.io
 
 protocol_versions_addr: "0x0C24F5098774aA366827D667494e9F889f7cFc08"
+
+regolith_time: 0 # genesis, overriden by individual chains
 canyon_time: 1699981200 # Tue Nov 14 17:00:00 UTC 2023
 delta_time: 1703116800  # Thu Dec 21 00:00:00 UTC 2023
-ecotone_time: 1707238800 # Tue Feb  6 17:00:00 UTC 2024
+ecotone_time: 1707238800 # Tue Feb 6 17:00:00 UTC 2024
--- a/superchain/configs/sepolia-dev-0/superchain.yaml
+++ b/superchain/configs/sepolia-dev-0/superchain.yaml
@@ -7,7 +7,8 @@ l1:
 protocol_versions_addr: "0x252CbE9517F731C618961D890D534183822dcC8d"
 superchain_config_addr: "0x02d91Cf852423640d93920BE0CAdceC0E7A00FA7"
 
-regolith_time: 0 # genesis
-canyon_time: 1698436800  # Fri Oct 27 20:00:00 UTC 2023
-delta_time: 1706555000   # Mon Jan 29 19:03:20 UTC 2024
-ecotone_time: 1706634000 # Tue Jan 30 17:00:00 UTC 2024
+regolith_time: 0         # Thu  1 Jan 1970 00:00:00 UTC (genesis)
+canyon_time: 1698436800  # Fri 27 Oct 2023 20:00:00 UTC
+delta_time: 1706555000   # Mon 29 Jan 2024 19:03:20 UTC
+ecotone_time: 1706634000 # Tue 30 Jan 2024 17:00:00 UTC
+
--- a/superchain/configs/sepolia-dev-0/superchain.yaml
+++ b/superchain/configs/sepolia-dev-0/superchain.yaml
@@ -7,7 +7,6 @@ l1:
 protocol_versions_addr: "0x252CbE9517F731C618961D890D534183822dcC8d"
 superchain_config_addr: "0x02d91Cf852423640d93920BE0CAdceC0E7A00FA7"
 
-regolith_time: 0         # Thu  1 Jan 1970 00:00:00 UTC (genesis)
 canyon_time: 1698436800  # Fri 27 Oct 2023 20:00:00 UTC
 delta_time: 1706555000   # Mon 29 Jan 2024 19:03:20 UTC
 ecotone_time: 1706634000 # Tue 30 Jan 2024 17:00:00 UTC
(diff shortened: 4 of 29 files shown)


superchain/superchain.go
```diff
@@ -440,13 +481,39 @@ type SuperchainConfig struct {
...
+	s.Name = temp.Name
+	s.L1 = temp.L1
+	s.ProtocolVersionsAddr = temp.ProtocolVersionsAddr
+	s.SuperchainConfigAddr = temp.SuperchainConfigAddr
+	s.hardForkDefaults = temp.HardForks
+
+	return nil
 }
 
+// TODO need to wrie a YAML unmarshaler for SuperchainConfig which handles the private fields properly
```
geoknee, line 515:
```suggestion
```

superchain/superchain.go
```diff
@@ -59,6 +66,40 @@ type ChainConfig struct {
...
+	// if c.RegolithTime == nil {
+	// 	c.RegolithTime = s.Config.hardForkDefaults.RegolithTime
+	// }
+
+	// if c.CanyonTime == nil {
+	// 	c.CanyonTime = s.Config.hardForkDefaults.CanyonTime
+	// }
+	//
+	// ...etc for each field in HardForkConfiguration
+
```
geoknee, line 102: A simpler alternative to this would be to _not_ have any superchain-wide defaults, and instead _require_ each chain to specify every hard fork timestamp.

validation/go.mod
```diff
@@ -4,11 +4,11 @@ go 1.21
 
 replace github.com/ethereum-optimism/superchain-registry/superchain => ../superchain
 
-replace github.com/ethereum/go-ethereum => github.com/ethereum-optimism/op-geth v1.101304.1
+replace github.com/ethereum/go-ethereum => github.com/ethereum-optimism/op-geth v1.101308.2-rc.1.0.20240222161420-b2b2a2ce17d5
```
geoknee, line 7: Here's a strategy we could employ to merge all of the in-flight work sanely:

1. Temporarily disable CI checks for linting and testing in `validation`
2. (optional) undo the changes to `validation` module on this PR
3. Merge this PR 
4. Update downstream PRs with the merge commit hash*
5. Return to this repo and reenable CI checks, update this go.mod file accordingly

* this makes me wonder if it is appropriate to have a more formal "release" process in this repo.

sebastianst: can't we just temporarily pin the op-geth commit of the PR that pulls in the superchain (also per temp commit)?

sebastianst: so the op-geth pr would reference a commit that has a failed validation check, but then we update this pr to point to the op-geth pr that points to this old temp commit, so we can already test validation while all PRs are still open.

geoknee: Yes I think this is doable and essentially the state of things as of now. I believe the validation checks running on this PR give us confidence that the overrides are being applied correctly.

 I suppose I was worried that we will be merging this PR to `main` despite it referencing a temporary commit of `op-geth`. But I guess this isn't so bad and that only needs to be the case for a short while. Probably no worse than disabling validation for a short while.

superchain/superchain.go
geoknee: If we want to continue this approach (using reflection) I am sure I can tidy this up even more / improve readability. The variable names aren't great and some of the reflection invocations could be abstracted a bit.

trianglesphere: Is there a way to always hydrate with the defaults & then set the overrides? Would that be cleaner than having to do this reflection.

geoknee: Conclusion: yes!

[set defaults first, then override by unmarshalling into nonempty struct](https://github.com/ethereum-optimism/superchain-registry/pull/101/commits/60086b977718e3cb6df27f83f55362c6dd287e32)

sebastianst (changes requested):
The solution looks good to me and we can go ahead moving over other fork times over here from https://github.com/ethereum-optimism/op-geth/pull/252

sebastianst:
And I just realized that we should probably just ignore Regolith because it will become irrelevant after Goerli is shut down, see https://github.com/ethereum-optimism/op-geth/pull/252#discussion_r1501676520
I suggest we move all other fork times over here.

superchain/configs/goerli/superchain.yaml
```diff
@@ -5,6 +5,8 @@ l1:
   explorer: https://goerli.etherscan.io
 
 protocol_versions_addr: "0x0C24F5098774aA366827D667494e9F889f7cFc08"
+
+regolith_time: 0 # genesis, overriden by individual chains
 canyon_time: 1699981200 # Tue Nov 14 17:00:00 UTC 2023
 delta_time: 1703116800  # Thu Dec 21 00:00:00 UTC 2023
-ecotone_time: 1707238800 # Tue Feb  6 17:00:00 UTC 2024
+ecotone_time: 1707238800 # Tue Feb 6 17:00:00 UTC 2024
```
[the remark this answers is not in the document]
geoknee: [745fb1e](https://github.com/ethereum-optimism/superchain-registry/pull/101/commits/745fb1e792c81bbbc29a2ec28831e990a55c91ec)

superchain/configs/goerli-dev-0/superchain.yaml
```diff
@@ -5,6 +5,8 @@ l1:
   explorer: https://goerli.etherscan.io
 
 protocol_versions_addr: "0x79ADD5713B383DAa0a138d3C4780C7A1804a8090"
+
+regolith_time: 0 # genesis, overriden by individual chains
 canyon_time: 1698436800 # October 27, 2023 8:00:00 PM UTC
 delta_time: 1701993600 # December 8, 2023 12:00:00 AM UTC, i.e. midnight Dec 7 to Dec 8
```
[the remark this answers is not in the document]
geoknee: [745fb1e](https://github.com/ethereum-optimism/superchain-registry/pull/101/commits/745fb1e792c81bbbc29a2ec28831e990a55c91ec)

superchain/configs/goerli/base.yaml
```diff
@@ -16,3 +16,4 @@ genesis:
     number: 0
   l2_time: 1675193616
 
+regolith_time: 1683219600 # May 4, 2023 @ 5:00:00 pm UTC
```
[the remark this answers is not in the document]
geoknee: [745fb1e](https://github.com/ethereum-optimism/superchain-registry/pull/101/commits/745fb1e792c81bbbc29a2ec28831e990a55c91ec)

superchain/superchain.go
```diff
@@ -42,6 +42,13 @@ type ChainGenesis struct {
 	ExtraData *HexBytes `yaml:"extra_data,omitempty"`
 }
 
+type HardForkConfiguration struct {
+	RegolithTime *uint64 `yaml:"regolith_time,omitempty"`
+	CanyonTime   *uint64 `yaml:"canyon_time,omitempty"`
+	DeltaTime    *uint64 `yaml:"delta_time,omitempty"`
+	EcotoneTime  *uint64 `yaml:"ecotone_time,omitempty"`
+	FjordTime    *uint64 `yaml:"fjord_time,omitempty"`
+}
```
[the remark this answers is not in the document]
geoknee: [7707698](https://github.com/ethereum-optimism/superchain-registry/pull/101/commits/7707698cd462953b70159e2aa9c8350e33800915)

superchain/superchain.go
```diff
@@ -440,11 +454,35 @@ type SuperchainConfig struct {
...
+func (s *SuperchainConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	temp := struct {
+		Name string           `yaml:"name"`
+		L1   SuperchainL1Info `yaml:"l1"`
+
+		ProtocolVersionsAddr *Address `yaml:"protocol_versions_addr,omitempty"`
+		SuperchainConfigAddr *Address `yaml:"superchain_config_addr,omitempty"`
+
+		HardForks HardForkConfiguration `yaml:",inline"`
+	}{}
```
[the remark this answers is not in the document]
geoknee: We can make it work by avoiding `yaml.Unmarshal` at the top level of the call stack [dd6c589](https://github.com/ethereum-optimism/superchain-registry/pull/101/commits/dd6c589725b8e08cd09fd2fc13d26f6f4bda9c8b)

superchain/superchain.go
```diff
@@ -440,11 +454,35 @@ type SuperchainConfig struct {
 	ProtocolVersionsAddr *Address `yaml:"protocol_versions_addr,omitempty"`
 	SuperchainConfigAddr *Address `yaml:"superchain_config_addr,omitempty"`
 
-	// Hardfork Configuration
-	CanyonTime  *uint64 `yaml:"canyon_time,omitempty"`
-	DeltaTime   *uint64 `yaml:"delta_time,omitempty"`
-	EcotoneTime *uint64 `yaml:"ecotone_time,omitempty"`
-	FjordTime   *uint64 `yaml:"fjord_time,omitempty"`
+	// Hardfork Configuration. These values may be overridden by individual chains.
+	hardForkDefaults HardForkConfiguration `yaml:",inline"`
```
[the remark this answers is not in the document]
geoknee: Good point, removed.

superchain/superchain.go
```diff
@@ -440,11 +454,35 @@ type SuperchainConfig struct {
...
+}
+
+// custom unmarshal function to allow yaml to be unmarshalled into unexported fields
+func (s *SuperchainConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
```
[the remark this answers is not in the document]
geoknee: Good spot, I must have found myself in an old version of that package's documentation somehow. Fixed now.

superchain/superchain.go
```diff
@@ -534,6 +572,9 @@ func init() {
 				panic(fmt.Errorf("failed to read superchain config %s/%s: %w", s.Name(), c.Name(), err))
 			}
 			var chainConfig ChainConfig
+
+			chainConfig.hydrateDefaultHardforkTimestamps(superchainEntry.Config)
```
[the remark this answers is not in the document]
geoknee: https://github.com/ethereum-optimism/superchain-registry/issues/103

superchain/superchain.go
geoknee: We need a test to show this works as expected. I suspect that the defaults are being overwritten with `nil` if the `ChainConfig` does not specify an override.

geoknee:
We should this override back into this repo https://github.com/ethereum-optimism/optimism/pull/9642/commits/77ac12d90ccb7e70dfb9b9a44eb69e6d98642cda

superchain/superchain.go
geoknee: Tests have now been added and the implementation has been reverted back to this reflection approach.

superchain/superchain.go
```diff
@@ -59,6 +66,34 @@ type ChainConfig struct {
...
+// with the superchain default.
+func (c *ChainConfig) setNilHardforkTimestampsToDefault(s *SuperchainConfig) {
+	cVal := reflect.ValueOf(c).Elem()
+	sVal := reflect.ValueOf(&s.hardForkDefaults).Elem()
+
+	hfcVal := reflect.ValueOf(HardForkConfiguration{})
+	for i := 0; i < hfcVal.NumField(); i++ {
+		hardForkName := hfcVal.Type().Field(i).Name
+		overrideValue := cVal.FieldByName(hardForkName)
+		if overrideValue.Interface().(*uint64) == nil {
```
sebastianst, line 84:
```suggestion
		if overrideValue.IsNil() {
```

superchain/superchain.go
```diff
@@ -59,6 +66,34 @@ type ChainConfig struct {
...
 	// This matches the resource filename, it is not encoded in the config file itself.
 	Chain string `yaml:"-"`
+
+	// Hardfork Configuration Overrides
+	HardForkConfiguration `yaml:",inline"`
+}
+
+// setNilHardforkTimestampsToDefault overwrites each unspecified hardfork activation time override
+// with the superchain default.
+func (c *ChainConfig) setNilHardforkTimestampsToDefault(s *SuperchainConfig) {
```
sebastianst, line 76: We might be able to simplify this if we iterate over the same structs like
```
    cVal := reflect.ValueOf(&c.HardForkConfiguration).Elem()
    sVal := reflect.ValueOf(&s.hardForkDefaults).Elem()

    for i := 0; i < cVal.NumField(); i++ {
        field := cVal.Field(i)
        if field.IsNil() {
            defaultValue := sVal.Field(i)
            field.Set(defaultValue)
        }
    }
}
```

superchain/superchain_test.go
```diff
@@ -395,3 +395,90 @@ func testNetworkUpgradeTimestampOffset(l2GenesisTime uint64, blockTime uint64, u
...
+		}}
+
+	t.Run("override: unmarshal with an override", func(t *testing.T) {
+		rawYAML := `canyon_time: 8`
+
+		c := ChainConfig{}
+		err := yaml.Unmarshal([]byte(rawYAML), &c)
+		c.setNilHardforkTimestampsToDefault(&s)
+
+		require.NoError(t, err)
```
sebastianst, line 452:
```suggestion
		err := yaml.Unmarshal([]byte(rawYAML), &c)
		require.NoError(t, err)

		c.setNilHardforkTimestampsToDefault(&s)
```
should assert right after. also in other tests.

superchain/superchain_test.go
```diff
@@ -395,3 +395,90 @@ func testNetworkUpgradeTimestampOffset(l2GenesisTime uint64, blockTime uint64, u
...
+	defaultCanyonTime := uint64(3)
+
+	// Set a ChainConfig to unmarshal into
+	// which already has a default value set
+	s := SuperchainConfig{
+		hardForkDefaults: HardForkConfiguration{
+			CanyonTime: &defaultCanyonTime,
+		}}
+
+	t.Run("override: unmarshal with an override", func(t *testing.T) {
```
sebastianst, line 445: Would be great to run these 3 tests in a test table instead. We could then also add 3 more cases for when the superchain `CanyonTime` field is nil. And one more complete test would also be good, where all currently known fork fields are set either in the superchain config, or in the chain config.

sebastianst (commented):
LGTM and I think the use of reflection is appropriate in this case. It's very simple code, yaml itself is also using reflection, and the code runs at init time, so no panic possible at (real) runtime of any project importing this.

A few suggestions to improve tests and to simplify the reflection code.

sebastianst:
> We should add this override back into this repo [ethereum-optimism/optimism@77ac12d](https://github.com/ethereum-optimism/optimism/commit/77ac12d90ccb7e70dfb9b9a44eb69e6d98642cda)

Agree, we can add this now as a `0` override to the op chain of `sepolia-devnet-0`

superchain/superchain.go
sebastianst: Embedded structs are still normal fields, so this should work.

superchain/superchain.go
geoknee: [68a9ffb](https://github.com/ethereum-optimism/superchain-registry/pull/101/commits/68a9ffb3519f18a2cb377062f2dfd9968392701a)

superchain/superchain.go
geoknee: [8177e13](https://github.com/ethereum-optimism/superchain-registry/pull/101/commits/8177e13de677aa47fcc190faf7bca38998e7598b)

geoknee:
> > We should add this override back into this repo [ethereum-optimism/optimism@77ac12d](https://github.com/ethereum-optimism/optimism/commit/77ac12d90ccb7e70dfb9b9a44eb69e6d98642cda)
> 
> Agree, we can add this now as a `0` override to the `op` chain of `sepolia-devnet-0`

Done in [8f7e057](https://github.com/ethereum-optimism/superchain-registry/pull/101/commits/8f7e057166411841be301b1c5e7d2b930aa132d1)

superchain/superchain_test.go
geoknee: [97db949](https://github.com/ethereum-optimism/superchain-registry/pull/101/commits/97db9495695936c9ba084aff1720cd8a1d5279d5)

geoknee: [2dcb036](https://github.com/ethereum-optimism/superchain-registry/pull/101/commits/2dcb0367ff3a685e44eed8de424de0234d6d6a6a)

superchain/configs/sepolia-dev-0/superchain.yaml
```diff
@@ -7,6 +7,7 @@ l1:
 protocol_versions_addr: "0x252CbE9517F731C618961D890D534183822dcC8d"
 superchain_config_addr: "0x02d91Cf852423640d93920BE0CAdceC0E7A00FA7"
 
-canyon_time: 1698436800  # Fri Oct 27 20:00:00 UTC 2023
-delta_time: 1706555000   # Mon Jan 29 19:03:20 UTC 2024
-ecotone_time: 1706634000 # Tue Jan 30 17:00:00 UTC 2024
+canyon_time: 1698436800  # Fri 27 Oct 2023 20:00:00 UTC
+delta_time: 1706555000   # Mon 29 Jan 2024 19:03:20 UTC
```
sebastianst, line 11: Let's actually set Canyon and Delta to 0 in this superchain config, and overwrite both values to these times in `base-devnet-0`. The reason is that any future chains joining this superchain will have Canyon and Delta activated at genesis already.

geoknee: [c2e25cc](https://github.com/ethereum-optimism/superchain-registry/pull/101/commits/c2e25ccf4d85f3a39d028452976c61696d82e093)

sebastianst (approved):

PR #104, opened 2024-02-26
update semver expectations after upgrade (mainnet)

geoknee:


ajsutton (approved):
Can we investigate ways that we can integrate these changes into the upgrade process rather than having to reactively update them?

PR #105, opened 2024-02-28
linting: add `gofumpt`

geoknee:
`gofumpt` is a more opinionated linter which (among other things) will complain if there are blank lines at the start or end of a function.

We have a few cases of that, and this rule often comes up during PR review. With this change, the rule is enforced in CI and applied on save for VSCode users. 

Questions:
* We may with to harmonize this with the approach in other optimism repos.


mds1 (approved):

.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
```
trianglesphere, line 1: TBH I don't love pulling in editor specific config to our repos.

.golangci.yml
```diff
@@ -4,7 +4,7 @@ run:
 linters:
   # also see https://golangci-lint.run/usage/linters/#enabled-by-default
   enable:
-    - goimports
+    - gofumpt
```
trianglesphere, line 7: Is `gofumpt` a superset of `goimports`? I know it's a superset of `gofmt`, but `goimports` is also a superset of `gofmt`

trianglesphere (commented):
I like integrating this via golangci lint. IDK if I'd want to use it to replace gofmt/goimports in people's normal formatting

.golangci.yml
geoknee: Good question. I think the answer is actually "no (but apparently don't worry about it?)" https://github.com/mvdan/gofumpt?tab=readme-ov-file#frequently-asked-questions

Reading that link, looks like we can enable both if we want to.

.vscode/settings.json
geoknee: I get that feeling a tiny bit, but wonder if the benefits outweigh the drawbacks. My overall aim is we talk about formatting changes as little as possible. This config makes it less likely for a developer to be surprised by linting failures on CI. 

I don't feel super strongly though. We can rip this out and let each dev sort their editor out by themselves.

mds1: > TBH I don't love pulling in editor specific config to our repos.

I thought this too but noticed we also have a `.vscode` in the monorepo from a few years back: https://github.com/ethereum-optimism/optimism/pull/1109

If we do remove this, we can add `.vscode` to the gitignore then explain how to set this up in CONTRIBUTING. Can be VSCode only instructions for now that are expanded over time

.golangci.yml
trianglesphere: Yea, let's enable both

.vscode/settings.json
geoknee: Sounds like a good compromise to me.

CONTRIBUTING.md
```diff
@@ -168,4 +168,18 @@ go run ./op-chain-ops/cmd/registry-data \
...
   --bytecodes-dir=$SUPERCHAIN_REPO/superchain/extra/bytecodes \
   --output=$SUPERCHAIN_REPO/superchain/extra/genesis/$SUPERCHAIN_TARGET/$CHAIN_NAME.json.gz
-```
\ No newline at end of file
+```
+
+## Setting up your editor for formatting and linting
+If you use VSCode, you can place the following in a `settings.json` file in the gitignored `.vscode` directory:
+
+```
```
mds1, line 176: nit

mds1, line 176: nit
```suggestion
```json
```

.vscode/settings.json
mds1: LGTM, let's also delete this existing `.vscode/settings.json` that's added in this PR

geoknee: [b97eb4e](https://github.com/ethereum-optimism/superchain-registry/pull/105/commits/b97eb4e5e050f8028fa6414e59b24288ae089758)

geoknee: [a3e513b](https://github.com/ethereum-optimism/superchain-registry/pull/105/commits/a3e513b23c1823faed9b3fcee450fb9833167155)

trianglesphere (approved):

PR #108, opened 2024-03-05
Do not attach privileged context to all CI runs

geoknee:
This is currently preventing checks running on PRs from external contributors.

This PR separates out two workflows:
* `hourly`: that runs on `main` on a schedule, and has the context
* `prs-check`: one that runs on a push to anything other than `main`, and does not have the context. 
  * the command that needs the context only actually needs it when on `main`, so we never hit a problem then


.circleci/config.yml
```diff
@@ -123,21 +123,21 @@ workflows:
...
-      - golang-test:
-          context:
-            - slack
+      - golang-test
       - check-codegen
       - check-forge-fmt
-      - check-security-configs:
-          context:
-            - slack
+      - check-security-configs
```
mds1, line 141: The hourly job only runs the `golang-test` and `check-security-configs` jobs. Just want to confirm those are the only ones we need on the schedule for now? It sounds correct based on the job names, but figured I'd double check.

.circleci/config.yml
```diff
@@ -123,21 +123,21 @@ workflows:
...
+      - schedule:
           cron: "0 * * * *"
           filters:
             branches:
               only:
                 - main
-  main:
+  pr-checks:
+    unless:
+        equal: [ main, << pipeline.git.branch >> ]
```
mds1, line 134: I think we should run this on main too, that way if the branch wasn't up to date with main before merge we check right away to make sure changes are still valid, then it will run on the hourly job afterwards?

mds1 (commented):
LGTM overall, two quick comments!

.circleci/config.yml
geoknee: Yeah, these are the only ones whose pass/fail nature can be changed without altering the source code.

.circleci/config.yml
geoknee: Good shout. Another option for combatting this is to require PRs are up to date before merging. But this is good, too [run prs-check on main, too](https://github.com/ethereum-optimism/superchain-registry/pull/108/commits/dd78be34848a898e4c83acf9b8f8b5fc6066db09)

mds1 (approved):
Thank you!

.circleci/config.yml
mds1: I think we might have already have the up to date requirement? Let's see
image

> We won't get the slack notification if this happens, BTW. Not until the next scheduled "hourly" job runs.

Good point. I think that is ok for now, especially if the above is correct. This way if the next hourly run does fail, and another PR happened to be merged in between, we'll have the run history on main to find the exact commit that caused the issue

PR #114, opened 2024-03-07
registry: MCP mainnet deploy

tynes:
**Description**

Deploys the MCP contracts to mainnet. Verifies on etherscan.
Implementation addresses are included so the canonical deployments
are tracked. In the future, we can move away from adding values to
network specific files if a standard create2 flow is adopted for
all chains. A future release can work towards this.

```bash
git checkout op-contracts/v1.3.0-rc.1
forge clean

export IMPL_SALT="op/acc"

forge script scripts/Deploy.s.sol \
  --sig 'deployImplementations()' \
  --rpc-url $ETH_RPC_URL \
  --private-key $PRIVATE_KEY \
  --broadcast \
  --verify
```


superchain/implementations/networks/mainnet.yaml
```diff
@@ -1,8 +1,15 @@
 l1_cross_domain_messenger:
+  2.3.0: 0xD3494713A5cfaD3F5359379DfA074E2Ac8C6Fd65
 l1_erc721_bridge:
+  2.1.0: 0xAE2AF01232a6c4a4d3012C5eC5b1b35059caF10d
 l1_standard_bridge:
+  2.1.0: 0x64B5a5Ed26DCb17370Ff4d33a8D503f0fbD06CfF
 l2_output_oracle:
   1.6.0: 0xB48B1827BC7218b1aB7B000b4f0416DF8F14B16A
+  1.8.0: 0xF243BEd163251380e78068d317ae10f26042B292
```
mds1, line 9: Should we delete the 1.6.0 version? It has immutables so it's not an MCP version, so I don't think we want that address here since it's OP Mainnet specific

mds1: I see in https://github.com/ethereum-optimism/superchain-registry/blob/main/CONTRIBUTING.md#implementations that it's supposed to be a canonical set of implementation deployments. I don't know what exactly that means pre-MCP, but it does make sense now, so I think we should delete 1.6.0 unless there is a purpose/use case for it I'm unaware of

tynes:
Pushed [a51070d](https://github.com/ethereum-optimism/superchain-registry/pull/114/commits/a51070dc140552e046e592b0c55fc15fd171827f) which removes the non mcp contract address from the registry, good catch

superchain/implementations/networks/mainnet.yaml
mds1: Resolved by @tynes in a51070dc140552e046e592b0c55fc15fd171827f

mds1 (approved):

tynes:
I cannot approve my own PR, wondering if i can get a review from @ethereum-optimism/go-reviewers

PR #115, opened 2024-03-08
Remove entire goerli and goerli-dev-0 superchains

geoknee:
Closes #85 

See also #116 , which is the minimal change to get tests passing again.

PR #116, opened 2024-03-08
skip GPO params check for goerli/op

geoknee:
The endpoint was shut down, causing tests to fail on `main`. This is the minimal change to get them passing again.


trianglesphere (approved):
Approving to make CI green, but goerli is EOL so we should fully remove goerli from the superchain registry

geoknee:
Yep that's in the pipeline #115

PR #121, opened 2024-03-12
Pull "adding a chain" section higher in CONTRIBUTING.md

geoknee:


CONTRIBUTING.md
```diff
@@ -4,8 +4,50 @@ See [Superchain Upgrades] OP-Stack specifications.
...
+
+### 4. Raise your Pull Request
+  Automated checks will run, and your PR will be reviewed in due course.
+
+
+
 ## Adding a superchain target
 
+> **Note**
+> This is an infrequent operation and unecessary if you are just looking to add a chain to an existing superchain.
```
mds1, line 49:
```suggestion
> This is an infrequent operation and unnecessary if you are just looking to add a new L2 chain.
```

geoknee: One of the scenarios we want to cover is the following: 

* say we already have a `holesky` superchain in place
* we launch a `holesky-dev-0/op-labs-dev-0` devnet chain and want to add that to the registry
* someone needs to follow the steps in this section despite "just looking to add a new L2 chain to an existing L1 chain"
* in short, superchain != L1

mds1: Sorry, maybe this is my lack of familiarity with devnets, but why isn't adding `holesky-dev-0` effectively adding a new devnet L1?

mds1: ahh yes that does make sense, thank you!

mds1 (approved):
Needs conflicts resolved but otherwise LGTM

PR #124, opened 2024-03-12
Split script between frontier chain and standard chain.

zchn:
1. We do not need all the existing checks to pass for Frontier Chains.
2. We know that if we register a chain's L1 contract addresses in the registry, we can get a significant information about the chain by inspecting the contracts.

As a result, this PR splits the minimum set of onboarding checks that a Frontier Chain needs to pass out of the standard checks, and updates the CONTRIBUTING.md to have separate processes for adding a standard chain v.s. a frontier chain.

All existing tests will continue to pass when someone follows the newly separately process to add a frontier chain, and we do not expect any impacts to any standard chains because the only file added will be a json file under https://github.com/ethereum-optimism/superchain-registry/tree/main/superchain/extra/addresses.

Tested with an example PR that only adds the json file in addresses: https://github.com/ethereum-optimism/superchain-registry/pull/123

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,20 +64,20 @@ EOF
 
 ### Adding a standard chain
 
-### Set env vars
+#### Set env vars
 
 To contribute a standard OP-Stack chain configuration, the following data is required: contracts deployment, rollup config, L2 genesis. We provide a tool to scrape this information from your local monorepo folder.
 
 First, make a copy of `.env.example` named `.env`, and alter the variables to appropriate values.
 
-### Run script
+#### Run script
 Then, run
 
 ```shell
 sh scripts/add-standard-chain.sh
 ```
 
-### Understand output
+#### Understand output
 The tool will write the following data:
 - The main configuration source, with genesis data, and address of onchain system configuration.
 - Addresses of L1 contracts. (Note that all L2 addresses are statically known addresses defined in the OP-Stack specification, and thus not configured per chain.)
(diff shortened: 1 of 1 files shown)


scripts/add-chain.sh
mds1, line 1: Should we delete this file?

scripts/add-standard-chain.sh
mds1, line 1: In the future we can consider deduping code by going back to a single script that takes a `[standard|frontier]` arg. But in the interest of time this is ok with me for now

scripts/add-chain.sh
mds1: AFAIK the only think that will break is the `https://github.com/ethereum-optimism/superchain-registry/blob/main/CONTRIBUTING.md#adding-a-chain` URL we've shared, since we've renamed the "Adding a chain" section

To preserve that we can restore that header and have subheaders

```markdown
## Adding a Chain

### Adding a Standard Chain

### Adding a Frontier Chain
```

Probably worth adding a table of contents with doctoc now too

mds1: yes after this is merged I will get a quick PR in to add a TOC and fix some shellcheck warnings

mds1 (approved):

CONTRIBUTING.md
```diff
@@ -62,17 +62,19 @@ EOF
 
 ## Adding a chain
 
+### Adding a standard chain
+
 ### Set env vars
```
geoknee, line 67: Nit: can we have each step nested under this new heading?
```suggestion
### Adding a standard chain

#### Set env vars
```

and so on

geoknee (commented):
I'm slightly concerned that there is nothing explicitly committed to the repository  to separate standard and frontier chains. 

To understand the difference, I need to read CONTRIBUTING.md, and then cross reference contents of a couple of different directories. 

I can foresee a future validation check just iterating through the `extra/addresses` folder, or some refactor happening without understanding the difference. 

I would prefer housing the data in a different folder, for example?

geoknee:
@zchn that works for me, thanks for the context.

PR #126, opened 2024-03-13
Mainnet MultiSig Sanity Checking

blmalone:
**Description**

The idea here is to have a similar task to `CheckSecurityConfigs` that'll give a consistent sanity check on our mainnet multisig configurations across Op stacks. 

Inherently, this feature requires us to actively keep track of signatories on the multisig contracts that we care about. If it's only for the Op Foundation and the Security Council then this would require a relatively low level of upkeep from us. 

**To Do:** 
- Decide when and how often this task should run (I don't see any reason why it isn't running _at least once per day_). 
- Integrate with Circle Ci or another Github Action like event based system

**Tests**

No tests, sanity check code. Tests could be added with more time. 

**Additional context**

Continually testing our assumptions about who the signer are on extremely important GnosisSafes in our system.

**Metadata**

N/A


scripts/CheckNetworkMultiSigConfigs.s.sol
```diff
@@ -0,0 +1,179 @@
...
+                string.concat(
+                    "Unsupported chain ID: ",
+                    vm.toString(block.chainid),
+                    ". Please call runOnDir(string,bool) directly."
+                )
+            );
+        }
+        string memory jsonDir = string.concat("superchain/extra/addresses/", network);
+        runOnDir(jsonDir);
+    }
```
mds1, line 48: Consider using forge-std's [StdChains](https://github.com/foundry-rs/forge-std/blob/master/src/StdChains.sol) to map from chain ID to network name using `getChain(chainId)`

blmalone:
@zchn No problem. No rush on my end here. Was a good exercise for me!

blmalone:
Going to close this PR - need more clarity around what checks we want to implement for multisigs. cc: @mds1

PR #128, opened 2024-03-14
update GPO params test expectation following mainnet Ecotone activation

geoknee:


superchain/params.go
```diff
@@ -108,6 +108,11 @@ var GasPriceOracleParams = map[string]UpgradeFilter{
 			Overhead: makeBigIntAndBounds(188, [2]int64{188, 188}),
 			Scalar:   makeBigIntAndBounds(684_000, [2]int64{684_000, 684_000}),
 		},
+		Ecotone: &EcotoneGasPriceOracleParamsWithBounds{
+			Decimals:          makeBigIntAndBounds(6, [2]int64{6, 6}),
+			BlobBaseFeeScalar: Uint32AndBounds{862_000, [2]uint32{500_000, 862_000}},
+			BaseFeeScalar:     Uint32AndBounds{7600, [2]uint32{970, 7600}},
```
geoknee, line 114: Setting these bounds wide enough to include all chains targeting mainnet, with the exception of PGN whose blobBaseFeeScalar has likely not yet been updated:

```
                Test:           TestGasPriceOracleParams/PGN_(424)
                Messages:       blobBaseFeeScalar 0 out of bounds [500000 862000]
```

mds1: FWIW once we refactor into the frontier/standard split, PGN will be a frontier chain so I don't think we'll need to worry about their ecotone params. So perhaps for now we can special case frontier chains in the check to they are skipped

geoknee: Thanks @mds1, that unblocks us! [7c14929](https://github.com/ethereum-optimism/superchain-registry/pull/128/commits/7c149298f1213db30e08c518794d7a9ba955d171)

trianglesphere (approved):

mds1 (approved):

PR #137, opened 2024-03-19
Enshrine standard vs frontier chains in repository directory structure

geoknee:
Towards #82 

* Go module exports and validation checks remain unchanged (covering standard directory only for now)
* This PR likely to cause some painful merge conflicts with other open PRs

Questions:
- [ ] have we considered using a flag instead of folder structure to delineate the type of chain? Noting that elsewhere we have heard the desire to get to a flatter repo structure whereas this is going in the other direction
- [ ] should we move any chains from standard to frontier now?
- [ ] should the frontier folder have the same structure as standard? 
- [ ] should codegen.js target both folders (currently just targets the standard one)
- [ ] should other validation checks be expanded to cover both folders? e.g. chain ID uniqueness check, arguably the lowest bar that any chain should be able to clear

Although the checks for each type of chain may differ, and we may not wish to export frontier chain information in e.g. the Go module, I think there are some advantages to keeping the structure the same.


mds1:
Some quick replies before I head out for the week. These are my initial reactions, so feel free to override anything here if you discuss and come to any decisions this week:

>  have we considered using a flag instead of folder structure to delineate the type of chain? Noting that elsewhere we have heard the desire to get to a flatter repo structure whereas this is going in the other direction

That's a good point, I would be ok with this. Perhaps the config file (e.g. [`op.yaml`](https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/configs/mainnet/op.yaml)) should have a `chain_kind: [frontier|standard]` entry.

We can always generate output files (or similar) that make it easier to see a summary of which are frontier and which are standard

>  should we move any chains from standard to frontier now?

Given "This PR likely to cause some painful merge conflicts with other open PRs", I'm not sure. We know the set of chains to move to frontier, but it will create more work for the open registry requests. I guess it depends how soon we think we can get those merged

>  should the frontier folder have the same structure as standard?

In general I'd say yes, acknowledging that some files may be missing from frontier (e.g. because the checks are less, and less data gets PR'd in). Though this becomes irrelevant if we go with flat folders

>  should codegen.js target both folders (currently just targets the standard one)
>  should other validation checks be expanded to cover both folders? e.g. chain ID uniqueness check, arguably the lowest bar that any chain should be able to clear

For both of these, I would say we maintain the status quo and separately track and decide which checks need to be in each. Because I agree at a minimum we should ensure some basic sanity checks, like all chains having unique chain IDs

geoknee:
Thanks for the feedback @mds1 ! I'll knock up another PR with the flag option, since I think that is going to be easier to manage.

geoknee:
Raised https://github.com/ethereum-optimism/superchain-registry/pull/140 . It's a much cleaner diff. The main downside is not being that easy for a human to drill into which chains are which type. But very easy for a machine.

PR #135, opened 2024-03-19
chore: Add dispute game factory address to sepolia/op.json

roninjin10:
**Description**

A clear and concise description of the features you're adding in this pull request.

**Tests**

Please describe any tests you've added. If you've added no tests, or left important behavior untested, please explain why not.

**Additional context**

Add any other context about the problem you're solving.

**Metadata**

- Fixes #[Link to Issue]

PR #144, opened 2024-03-26
feat: Add fault proof checks to `CheckSecurityConfigs` script

clabby:
## Overview

Modifies the `CheckSecurityConfigs` script to add support for checking fault proof contracts' owners and proxy administrators.


README.md
```diff
@@ -61,6 +61,8 @@ the following privilege grants and role designations:
       L2 senders.
    4. Trusted oracles. For example, `OptimismPortalProxy` specifies
       the L2 oracle they trust with the L2 state root information.
+      1. After the FPAC upgrade, the `OptimismPortalProxy` specifies the `DisputeGameFactory` they trust rather
+      than the legacy `L2OutputOracle` contract.
```
maurelian, line 65: Should also add update the priviledged roles under item 3 below.

scripts/CheckSecurityConfigs.s.sol
```diff
@@ -217,22 +241,41 @@ contract CheckSecurityConfigs is Script {
         return abi.decode(addrBytes, (address));
     }
 
-    function getAddresses(string memory addressesJsonPath) internal view returns (ProtocolAddresses memory) {
+    function getAddresses(string memory addressesJsonPath) internal view returns (ProtocolAddresses memory, bool) {
```
maurelian, line 244: I think we should rename this function now. 
Perhaps: 
`getAddressesAndFpacStatus()` I know it's clunky, but I think clarity is more important. Open to other names too.

maurelian (commented):
A couple nits, otherwise LGTM

superchain/extra/addresses/addresses.json
```diff
@@ -146,6 +146,9 @@
     "SystemConfigProxy": "0x034edD2A225f7f429A63E0f1D2084B9E0A93b538",
     "ProxyAdminOwner": "0xDEe57160aAfCF04c34C887B5962D0a69676d3C8B",
     "SystemConfigOwner": "0xfd1D2e729aE8eEe2E146c033bf4400fE75284301",
+    "DisputeGameFactoryProxy": "0x05F9613aDB30026FFd634f38e5C4dFd30a197Fa1",
+    "AnchorStateRegistryProxy": "0x1874A5CCA34eBCe29aDf1ebdd7fD2725fe184838",
+    "DelayedWETHProxy": "0x9bD56B189234FE25CF3547c4343e50C3259e383a",
```
mds1, line 151: We also need to include the implementation addresses for these, and the updated portal implementation, in https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/implementations/networks/sepolia.yaml. Then we can close #135 and replace it with this one

superchain/implementations/networks/sepolia.yaml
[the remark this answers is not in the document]
mds1: Actually on second thought, none of the implementations are tagged nor governance approved yet. So if we put them in now for completeness, we have to delete them later.

Longer term what we probably should do is tag any monorepo commit we deploy contracts from with a `-beta` version or similar, and the semvers in this file should be organized by their corresponding monorepo tag name. However, the full set of contracts for an FPAC system is really: the "release" we deployed a few FPAC sepolia contracts from + the `op-contracts/v1.3.0` release for the remainder of contracts. 

We don't really have a good way to specify that pairing, nor a good way to enforce that in the registry's checks. I'm not sure of the right solution here, but it hints we still need to iterate on our contracts release process and what a "standard config set of contracts" is concretely defined as, since it's not just a monorepo tag. cc @geoknee @maurelian for thoughts

superchain/implementations/networks/sepolia.yaml
```diff
@@ -9,6 +9,8 @@ l2_output_oracle:
 optimism_mintable_erc20_factory:
   1.9.0: 0xE01efbeb1089D1d1dB9c6c8b135C934C0734c846
 optimism_portal:
-  2.5.0: 0x2D778797049FE9259d947D1ED8e5442226dFB589
+  3.3.0: 0x215b3A2d1766616d1C4701f3D117348711BaAe93
```
clabby, line 12:
```suggestion
  2.5.0: 0x2D778797049FE9259d947D1ED8e5442226dFB589
  3.3.0: 0x215b3A2d1766616d1C4701f3D117348711BaAe93
```

sbvegan (commented):
Okay overall this looks good, just some notes and open questions:
- `README` update looks good. The diagram is a little wild but appears accurate.
- I ran the security script against with Sepolia and the checks all passed.
- ~How do I validate the semver is correct?~ I found the versions through etherscan and they match up.
- What do you think about Matt's comment? I noticed those implementation contracts are missing too, is that expected?

sbvegan (approved):
@clabby followed up with me on slack, lgtm to fix the CI and missing implementation contracts will come later

mds1 (approved):

PR #140, opened 2024-03-20
Enshrine standard vs frontier chains in yaml enum

geoknee:
Closes https://github.com/ethereum-optimism/security-pod/issues/82

I considered and discarded the alternative approach of having an even more deeply nested folder structure here #137.
This is a pretty lightweight way of demarcating the two types of chains, which:
* should be sufficient to switch the validation checks and so on easily between the two types
* minimizes conflicts with in flight PRs
* minimizes overall diff and chances of introducing a bug
* minimizes change to API of the Go module and repo more generally.

PR #145, opened 2024-03-27
Remove unnecessary information from frontier chains.

zchn:
Based on the process at
https://github.com/ethereum-optimism/superchain-registry/blob/main/CONTRIBUTING.md#adding-a-frontier-chain,
we would not have created these files for these chains because they
are frontier chains, so removing them to reduce the maintainance
burden. We can always add them back if we add more requirements for
frontier chain information in the registry, or if they become a
standard chain.


geoknee:
Closing because we are not adopting this approach #140

PR #153, opened 2024-04-02
validation: add global chainId check

geoknee:
Closes https://github.com/ethereum-optimism/security-pod/issues/100

Test currently failing due to name mismatches. 
It is likely and acceptable for a chain to already exist at chainId.network, so we need a way of checking whether it is the "same chain" or not. Using the name seems reasonable, since that it is [required to be unique also ](https://github.com/ethereum-lists/chains?tab=readme-ov-file#constraints). 

=== RUN   TestGloballyUniqueChainId
    /Users/georgeknee/code/ethereum-optimism/superchain-registry/validation/chainId_test.go:22: Chain OP-Mainnet with ID 10 exists in global chain list with name OP Mainnet
    /Users/georgeknee/code/ethereum-optimism/superchain-registry/validation/chainId_test.go:22: Chain Zora Sepolia with ID 999999999 exists in global chain list with name Zora Sepolia Testnet
    /Users/georgeknee/code/ethereum-optimism/superchain-registry/validation/chainId_test.go:22: Chain Lyra with ID 957 exists in global chain list with name Lyra Chain
    /Users/georgeknee/code/ethereum-optimism/superchain-registry/validation/chainId_test.go:22: Chain Orderly Network with ID 291 exists in global chain list with name Orderly Mainnet
    /Users/georgeknee/code/ethereum-optimism/superchain-registry/validation/chainId_test.go:22: Chain PGN with ID 424 exists in global chain list with name PGN (Public Goods Network)
    /Users/georgeknee/code/ethereum-optimism/superchain-registry/validation/chainId_test.go:22: Chain Base Sepolia with ID 84532 exists in global chain list with name Base Sepolia Testnet
    /Users/georgeknee/code/ethereum-optimism/superchain-registry/validation/chainId_test.go:22: Chain OP-Sepolia with ID 11155420 exists in global chain list with name OP Sepolia Testnet
    /Users/georgeknee/code/ethereum-optimism/superchain-registry/validation/chainId_test.go:22: Chain PGN Sepolia with ID 58008 exists in global chain list with name Sepolia PGN (Public Goods Network)


mds1:
My preferred solution would be:
- If the chain is not in the upstream repo: we do the chain ID + name + short name uniqueness checks ourselves
- If the chain is in the upstream repo: make sure everything added here matches

What do you think of that?

geoknee:
@mds1 I updated the logic to match your suggestion. 

I also went through the existing chains to add a short name and correct the existing names where discrepancies exist. 

Further, the contributing guide has been updated. 

Note that this will conflict with in flight PRs.

validation/uniqueness_test.go
```diff
@@ -0,0 +1,122 @@
...
+		if globalChainIds[uint(chain.ChainID)] != nil {
+			globalChainName := globalChainIds[uint(chain.ChainID)].Name
+			globalShortName := globalChainIds[uint(chain.ChainID)].ShortName
+			assert.Equal(t, globalChainName, chain.Name,
+				"Local chain name for %d does not match name from chainid.network", chain.ChainID)
+			assert.Equal(t, chain.ShortName, globalShortName,
+				"Local short chain name for %d does not match name from chainid.network", chain.ChainID)
+		}
+
+		assert.NoError(t, localChainIds.AddIfUnique(chain.ChainID))
```
geoknee, line 118: This is redundant since `OPChains` is already a "set" of chainIds (a mapping keyed by chainId).

We also have a check [here ](https://github.com/ethereum-optimism/superchain-registry/blob/fdd0d21715eea9fdfd48eac191e23d1249c59899/superchain/superchain_test.go#L42) which does not use the `OPChains` mapping.

superchain/configs/mainnet/op.yaml
```diff
@@ -1,4 +1,5 @@
-name: OP-Mainnet
+name: OP Mainnet
```
geoknee, line 1: We should verify if these changes will break something in the downstream dependencies (or even somewhere in this repo that I have missed).

geoknee: https://github.com/ethereum-optimism/optimism/pull/10043

geoknee: https://github.com/ethereum-optimism/op-geth/pull/287

mds1:
```
Checking OptimismPortalProxy 0x16Fc5058F25648194471939df75CF27A2fdC48BC
    -- Success: 0x189aBAAaa82DfC015A588A7dbaD6F13b1D3485Bc == 0x16Fc5058F25648194471939df75CF27A2fdC48BC.admin().
    -- Success: 0xDEe57160aAfCF04c34C887B5962D0a69676d3C8B == 0x16Fc5058F25648194471939df75CF27A2fdC48BC.GUARDIAN().
    !! Error calling 0x16Fc5058F25648194471939df75CF27A2fdC48BC.L2_ORACLE()
    !! Error: 0x90E9c4f8a994a250F6aEfd61CAFb4F2e895D458F != 0x16Fc5058F25648194471939df75CF27A2fdC48BC.L2_ORACLE(), 
             which is 0x0000000000000000000000000000000000000000
```
```
├─ [648] 0x16Fc5058F25648194471939df75CF27A2fdC48BC::L2_ORACLE() [staticcall]
    │   ├─ [192] 0x215b3A2d1766616d1C4701f3D117348711BaAe93::L2_ORACLE() [delegatecall]
    │   │   └─ ← EvmError: Revert
```

So it's reverting because `L2_ORACLE` method does not exist on the target contract. And the `0x215b3A2d1766616d1C4701f3D117348711BaAe93` contract is the FPAC OptimismPortal2 where the L2OO has been removed. So I think we need to use `chainUpgradedToFPAC` for `basesep` now. (I did not realize Base Sepolia upgraded to FPAC but seems they have)

geoknee:
Thanks, need to get better at parsing those traces :-) Is there anything we can do to make that easier

>  (I did not realize Base Sepolia upgraded to FPAC but seems they have)

I did run this check on `main` and it passes, so there must be something linked to my renaming changes on this branch  which are causing the test to fail. 

I think we actually just need to update `chainUpgradedToFPAC` for `opsep` (not `basesep`).

mds1:
> I think we actually just need to update chainUpgradedToFPAC for opsep (not basesep).

Ah yes, foundry runs tests in parallel so the console.logs get a little confusing. But given the address you are correct

scripts/CheckSecurityConfigs.s.sol
```diff
@@ -90,9 +90,9 @@ contract CheckSecurityConfigs is Script {
         // Only testnet chains may be added as an exception here.
         return keccak256(
             abi.encodePacked(
-                sliceString(addressesJsonPath, bytes(addressesJsonPath).length - 15, bytes(addressesJsonPath).length)
+                sliceString(addressesJsonPath, bytes(addressesJsonPath).length - 18, bytes(addressesJsonPath).length)
             )
-        ) == keccak256(abi.encodePacked("sepolia/op.json"));
```
geoknee, line 95: Really keen to replace this code (see https://github.com/ethereum-optimism/superchain-registry/pull/144).

mds1: Just pushed 3e61dba to clean this up, feel free to revert the commit if you prefer

geoknee:
In favour of #320

PR #154, opened 2024-04-02
validation: allow checks to be focussed to a single chain

geoknee:
Closes: https://github.com/ethereum-optimism/security-pod/issues/101


CONTRIBUTING.md
```diff
@@ -51,7 +51,16 @@ The format is a gzipped JSON `genesis.json` file, with either:
...
   Nodes can load the genesis block header, and state-sync to complete the node initialization.
 
-### 4. Raise your Pull Request
+### 4. Run tests locally
+Run the following command to run our validation checks, only on the chain you added:
+
+```
+cd validation
+go test ./... -v -focus=true -chain-id=<your-chain-id-here>
+```
```
geoknee, line 60: This is how you can now focus the validation checks to a single chain.

CONTRIBUTING.md
```diff
@@ -51,7 +51,16 @@ The format is a gzipped JSON `genesis.json` file, with either:
 - a `stateHash` attribute: to omit a large state (e.g. for networks with a re-genesis or migration history).
   Nodes can load the genesis block header, and state-sync to complete the node initialization.
 
-### 4. Raise your Pull Request
+### 4. Run tests locally
+Run the following command to run our validation checks, only on the chain you added:
```
mds1, line 55: nit
```suggestion
Run the following command to run the registry's validation checks, for only the chain you added:
```

validation/utils_test.go
```diff
@@ -51,6 +51,12 @@ func TestAreCloseInts(t *testing.T) {
 
 func SkipCheckIfFrontierChain(t *testing.T, chain superchain.ChainConfig) {
 	if chain.Type == superchain.Frontier {
-		t.Skip()
+		t.Skip("Skipping check (Frontier chain)")
+	}
+}
+
+func SkipCheckIfBlurred(t *testing.T, chain superchain.ChainConfig) {
+	if focus && focussedChainId != chain.ChainID {
```
mds1, line 59: Can we remove the `focus` bool to simplify the UX to:
```
go test ./... -v -chain-id=<your-chain-id-here>
```

And this check then becomes `if focussedChainId != chain.ChainID {`

geoknee: [simplify API for focussing tests](https://github.com/ethereum-optimism/superchain-registry/pull/154/commits/35669302c1b781efed6ffe63133cbe94ce643912)

validation/init_test.go
```diff
@@ -18,8 +19,14 @@ func init() {
...
 	flag.Parse()
+	var fcid uint64
+	var err error
 	if rawFocussedChainId != "" {
-		fcid, err := strconv.ParseUint(rawFocussedChainId, 10, 64)
+		if strings.HasPrefix(rawFocussedChainId, "0x") {
+			fcid, err = strconv.ParseUint(strings.TrimPrefix(rawFocussedChainId, "0x"), 16, 64)
+		} else {
+			fcid, err = strconv.ParseUint(rawFocussedChainId, 10, 64)
+		}
```
geoknee, line 29: This could be extracted into a helper with a unit test.

CONTRIBUTING.md
```diff
@@ -51,7 +51,16 @@ The format is a gzipped JSON `genesis.json` file, with either:
 - a `stateHash` attribute: to omit a large state (e.g. for networks with a re-genesis or migration history).
   Nodes can load the genesis block header, and state-sync to complete the node initialization.
 
-### 4. Raise your Pull Request
+### 4. Run tests locally
+Run the following command to run the registry's validation checks, for only the chain you added:
+
+```
+cd validation
+go test ./... -v -focus-chain-id=<your-chain-id-here>
```
trianglesphere, line 59: Could we consider ensuring that all per-chain tests use the test name & use the native `-run=regex` or does that not meet all of our requirements for this flag?

geoknee: Good suggestion. I think that could work, and would probably require less boiler plate code and a similar amount of diligence writing new tests. And it may allow focussing by name as well as by chain id, which would be nice.

geoknee:
In favour of #158

PR #157, opened 2024-04-02
validation: allow `L2OutputOracle` param checks with older ABI

### LINKED ISSUE #156: validation: Allow submissionInterval check for older contract versions
The deployed version is a valid ABI (it corresponds to the ABI at tag `op-contracts/v1.0.0`), so our checks should be updated to support it

_Originally posted by @mds1 in https://github.com/ethereum-optimism/superchain-registry/pull/107#discussion_r1546864145_

geoknee:
Closes #156

### DIFF
--- a/superchain/params.go
+++ b/superchain/params.go
@@ -39,9 +39,9 @@ type L2OOParams struct {
 
 // OPMainnetL2OOParams describes the L2OutputOracle parameters from OP Mainnet
 var OPMainnetL2OOParams = L2OOParams{
-	SubmissionInterval:        big.NewInt(120),
+	SubmissionInterval:        big.NewInt(1800),
 	L2BlockTime:               big.NewInt(2),
-	FinalizationPeriodSeconds: big.NewInt(12),
+	FinalizationPeriodSeconds: big.NewInt(604800),
 }
 
 // OPSepoliaL2OOParams describes the L2OutputOracle parameters from OP Sepolia
--- a/validation/internal/legacy/l2outputoracle-1.3.0.go
+++ b/validation/internal/legacy/l2outputoracle-1.3.0.go
@@ -1,0 +1,1218 @@
+// Code generated - DO NOT EDIT.
+// This file is a generated binding and any manual changes will be lost.
+
+package legacy
+
+import (
+	"errors"
+	"math/big"
+	"strings"
+
+	ethereum "github.com/ethereum/go-ethereum"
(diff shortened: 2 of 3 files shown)


validation/internal/legacy/l2outputoracle-1.3.0.go
```diff
@@ -0,0 +1,1218 @@
+// Code generated - DO NOT EDIT.
+// This file is a generated binding and any manual changes will be lost.
+
+package legacy
```
geoknee, line 4: Maybe a bit of a heavyweight way of solving this. I copypasted the bindings from an older commit in the monorepo.

mds1: This seems ok for now. Maybe we add a TODO comment that this can be deleted once chains in the registry are on MCP (op-contracts/v1.3.0)?

superchain/params.go
```diff
@@ -39,9 +39,9 @@ type L2OOParams struct {
 
 // OPMainnetL2OOParams describes the L2OutputOracle parameters from OP Mainnet
 var OPMainnetL2OOParams = L2OOParams{
-	SubmissionInterval:        big.NewInt(120),
+	SubmissionInterval:        big.NewInt(1800),
 	L2BlockTime:               big.NewInt(2),
-	FinalizationPeriodSeconds: big.NewInt(12),
+	FinalizationPeriodSeconds: big.NewInt(604800),
```
mds1, line 44: Huh, did we have the wrong values here before?

validation/internal/legacy/l2outputoracle-1.3.0.go
geoknee: [add TODO comment](https://github.com/ethereum-optimism/superchain-registry/pull/157/commits/9d37013ddb9bdacb03746ddf802939fb290a1557)

geoknee: We could also strip out a lot of this bindings file, and just keep what we need?

geoknee: [56eefb8](https://github.com/ethereum-optimism/superchain-registry/pull/157/commits/56eefb8ecdd0f98a024c01a767cf3066cd384c57)

mds1: Thank you!

mds1 (approved):

sebastianst (approved):

PR #162, opened 2024-04-04
enhance add-chain.sh to scrape missing addresses

### LINKED ISSUE #149: Support admin and service roles in add-standard-chain.sh
The automation in [add-standrd-chain.sh](https://github.com/ethereum-optimism/superchain-registry/blob/77a930120ec63dd50c43483c82b1a0a29939ed27/scripts/add-standard-chain.sh#L62) does not automatically add the following roles to the json file:

```
  "ProxyAdminOwner": "",
  "SystemConfigOwner": "",
  "Guardian": "",
  "Challenger": ""
```

As a result, people get the following errors: https://app.circleci.com/pipelines/github/ethereum-optimism/superchain-registry/1687/workflows/6b7258c4-deeb-499e-8f67-b7a6ef65ea58/jobs/5174?invite=true#step-102-4143_0

add-standard-chain.sh should be extended to automatically add these roles

geoknee:
Uses cast call, so introduces an extra dependency and network requests into this script

Closes #149

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,6 +8,9 @@ See [Superchain Upgrades] OP-Stack specifications.
 
 ## Adding a chain
 
+### 0. Install dependencies
+You will need [`jq`](https://jqlang.github.io/jq/download/) and [`foundry`](https://book.getfoundry.sh/getting-started/installation) installed.
+
 ### 1. Set env vars
 
 To contribute a standard OP-Stack chain configuration, the following data is required: contracts deployment, rollup config, L2 genesis. We provide a tool to scrape this information from your local [monorepo](https://github.com/ethereum-optimism/optimism) folder.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,7 @@ See [Superchain Upgrades] OP-Stack specifications.
 ## Adding a chain
 
 ### 0. Install dependencies
-You will need [`jq`](https://jqlang.github.io/jq/download/) and [`foundry`](https://book.getfoundry.sh/getting-started/installation) installed.
+You will need [`jq`](https://jqlang.github.io/jq/download/) and [`foundry`](https://book.getfoundry.sh/getting-started/installation) installed, as well as Go.
 
 ### 1. Set env vars
 
--- a/scripts/add-chain.sh
+++ b/scripts/add-chain.sh
@@ -83,19 +83,56 @@ ecotone_time: $(jq -j .ecotone_time $ROLLUP_CONFIG)
 EOF
 
 
-# add extra addresses data
+# infer appropriate L1_RPC_URL
+case $SUPERCHAIN_TARGET in
+    "mainnet")
+        L1_RPC_URL="https://ethereum-mainnet-rpc.allthatnode.com"
+        ;;
+    "sepolia")
+        L1_RPC_URL="https://ethereum-sepolia-rpc.allthatnode.com"
+        ;;
+    *)
+        echo "Unsupported Superchain Target $SUPERCHAIN_TARGET"
+        exit 1
+        ;;
+esac
+
+# scrape addresses from static deployment artifacts
+AddressManager=$(jq -j .address $DEPLOYMENTS_DIR/AddressManager.json)
+L1CrossDomainMessengerProxy=$(jq -j .address $DEPLOYMENTS_DIR/L1CrossDomainMessengerProxy.json)
+L1ERC721BridgeProxy=$(jq -j .address $DEPLOYMENTS_DIR/L1ERC721BridgeProxy.json)
+L1StandardBridgeProxy=$(jq -j .address $DEPLOYMENTS_DIR/L1StandardBridgeProxy.json)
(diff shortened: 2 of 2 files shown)


scripts/add-chain.sh
```diff
@@ -78,19 +78,56 @@ genesis:
 EOF
 
 
-# add extra addresses data
+# infer appropriate L1_RPC_URL
+case $SUPERCHAIN_TARGET in
+    "mainnet")
+        L1_RPC_URL="https://ethereum-mainnet-rpc.allthatnode.com"
```
geoknee, line 84: This URL already exists in the `superchain/configs/mainnet/superchain.yaml` file. For now I will just replicate it, didn't seem worth parsing yaml in a shell script to dedupe.

Perhaps +1 point for rewriting this script in a different language if we want to do that.

scripts/add-chain.sh
```diff
@@ -78,19 +78,56 @@ genesis:
...
+L1StandardBridgeProxy=$(jq -j .address $DEPLOYMENTS_DIR/L1StandardBridgeProxy.json)
+L2OutputOracleProxy=$(jq -j .address $DEPLOYMENTS_DIR/L2OutputOracleProxy.json)
+OptimismMintableERC20FactoryProxy=$(jq -j .address $DEPLOYMENTS_DIR/OptimismMintableERC20FactoryProxy.json)
+SystemConfigProxy=$(jq -j .address $DEPLOYMENTS_DIR/SystemConfigProxy.json)
+OptimismPortalProxy=$(jq -j .address $DEPLOYMENTS_DIR/OptimismPortalProxy.json)
+SystemConfigProxy=$(jq -j .address $DEPLOYMENTS_DIR/SystemConfigProxy.json)
+
+# scrape remaining address live from the chain
+SuperchainConfig=$(cast call $OptimismPortalProxy "superchainConfig()(address)" -r $L1_RPC_URL) || "" # first command could fail if bedrock
+Guardian=$(cast call $SuperchainConfig "guardian()(address)" -r $L1_RPC_URL) || $(cast call $OptimismPortalProxy "GUARDIAN()(address)" -r $L1_RPC_URL) # first command could fail if bedrock
```
geoknee, line 108: Another +1 point for transcribing into a different language: the error handling here _should_ work, but it is not pretty.

scripts/add-chain.sh
```diff
@@ -78,19 +78,56 @@ genesis:
...
+Challenger=$(cast call $L2OutputOracleProxy "challenger()(address)" -r $L1_RPC_URL) # first command could fail if FPAC: TODO figure out how to get this role if FPAC
```
geoknee, line 109: Highlighting TODO

scripts/add-chain.sh
```diff
@@ -78,19 +78,56 @@ genesis:
...
+ProxyAdmin=$(jq -j .address $DEPLOYMENTS_DIR/ProxyAdmin.json)
+ProxyAdminOwner=$(cast call $ProxyAdmin "owner()(address)" -r $L1_RPC_URL)
+SystemConfigOwner=$(cast call $SystemConfigProxy "owner()(address)" -r $L1_RPC_URL)
```
geoknee, line 112: Another +1 for transcribing into e.g. Golang: we have utilities for retrying/ backoff if these calls get rate limited. 

Y

CONTRIBUTING.md
```diff
@@ -8,6 +8,9 @@ See [Superchain Upgrades] OP-Stack specifications.
 
 ## Adding a chain
 
+### 0. Install dependencies
+You will need [`jq`](https://jqlang.github.io/jq/download/) and [`foundry`](https://book.getfoundry.sh/getting-started/installation) installed.
```
geoknee, line 12: Do we have a sense of how "bad" it is to require foundry, here? We are already requiring Go, so if the script were transcribed to Go we wouldn't need to add the dependency.

scripts/add-chain.sh
mds1: Another idea is to use [yq](https://github.com/mikefarah/yq) as a drop in replacement for jq, and it also handles YAML

scripts/add-chain.sh
mds1: cc @clabby guidance here

scripts/add-chain.sh
mds1: I think `cast` might have some retry logic built in, but ultimately no objections to migrating this script to another language eventually

CONTRIBUTING.md
mds1: Currently you need foundry to deploy contracts for a chain anyway, so I think this is ok for now

scripts/add-chain.sh
clabby: Ah interesting - there's no external getter for the `CHALLENGER` role in the `PermissionedDisputeGame`. We can add one if we strictly require it, though will require another upgrade on our end.

Possible to retrieve from the deploy configuration, or is this strictly meant to reference on-chain data?

mds1: Got it, thanks! Since its currently only OP Sepolia with FPAC, we can probably leave this TODO unresolved for now since other chains have the `challenger` method. Longer terms we should definitely add getters for these

mds1 (approved):

PR #119, opened 2024-03-11
Add Superchain Config for Manta Pacific

TheAppleTucker:
**Description**

Add the basic superchain config for Manta Pacific.

PR #163, opened 2024-04-04
Update CODEOWNERS per new team responsibilities

zchn:
Security team will continue to own the 3 existing PRs to support them across the finish line:
1. https://github.com/ethereum-optimism/superchain-registry/pull/125
2. https://github.com/ethereum-optimism/superchain-registry/pull/119
3. https://github.com/ethereum-optimism/superchain-registry/pull/107

Making this change so future onboarding PRs will go directly to @OPMattie 's team

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -10,10 +10,6 @@
 /validation @ethereum-optimism/go-reviewers
 /superchain @ethereum-optimism/go-reviewers
 
-## JS&TS
-*.js @roninjin10
-package.json @roninjin10
-
 ## Solidity
 *.sol @ethereum-optimism/security-reviewers
 


.github/CODEOWNERS
```diff
@@ -1,15 +1,25 @@
...
 *.go @ethereum-optimism/go-reviewers
 /go.work @ethereum-optimism/go-reviewers
 /.golangci.yml @ethereum-optimism/go-reviewers
 /validation @ethereum-optimism/go-reviewers
 /superchain @ethereum-optimism/go-reviewers
-/superchain/extra/addresses @ethereum-optimism/security-reviewers
 
+## JS&TS
 *.js @roninjin10
 package.json @roninjin10
```
mds1, line 15: We should probably put someone else here since @roninjin10 will be leaving soon 😢

We have a `package.json`, 1 JS file, no TS files, and no deps. The main thing in the `package.json` is an alias for running the codgen script. Potential solution:
- Run the codegen script as part of `add-chain.sh` (cc @geoknee)
- Then we can remove the `package.json` (or do we currently use this for npm releases @roninjin10?)
- Make security the codeowner for TS/JS for now

.github/CODEOWNERS
```diff
@@ -1,15 +1,25 @@
...
 
+# Code and script ownership
+
+## Go
 *.go @ethereum-optimism/go-reviewers
 /go.work @ethereum-optimism/go-reviewers
 /.golangci.yml @ethereum-optimism/go-reviewers
 /validation @ethereum-optimism/go-reviewers
 /superchain @ethereum-optimism/go-reviewers
-/superchain/extra/addresses @ethereum-optimism/security-reviewers
```
mds1, line 9: Was `/superchain/extra/addresses` intentionally removed? Perhaps we split the fallback owners for the `/superchain` dir to `@ethereum-optimism/go-reviewers` for `*.go` + `go.mod` + `go.sum` files, and `@sbvegan @ZakAyesh @OPMattie` for the rest

.github/CODEOWNERS
mds1: Sorry, I don't think I see the security JS/TS codeowner line?

.github/CODEOWNERS
mds1: > That's more or less what I am doing in this PR except that for implementations that's more of a superchain-ops and chain upgrade thing than a chain onboarding thing? i.e. chains being added do not need to modify /superchain/implementations

So the contributing file says:

> Per superchain a set of canonical implementation deployments, per semver version, is tracked.
As default, an empty collection of deployments can be set:

I think this should be tweaked to clarify that it's the "governance approved standard implementations starting from `op-contracts/v1.3.0` and later" (i.e. MCP and later). In which case I agree with the `/superchain/implementations @ethereum-optimism/security-reviewers` assignment

.github/CODEOWNERS
mds1: Ah I was looking for JS/TS explicitly. This works, resolving!

roninjin10: We need package.json so you can install superchain repo like `npm install https://github.com/ethereum-optimism/superchain-registry/tree/main/superchain/configs/goerli`. It also is used for npm releases.

I think security team can get by owning it just fine as I wouldn't expect the package.json to change much if ever.

If we wanted a js person to still own it @nitaliano is best bet

roninjin10: correction the only change to it is bumping the version number if releasing a new version to npm.

It's perfectly reasonable IMO to expect people to install from github though like the command I put above

mds1: Got it! Do you know if any projects that install this repo in that way? And presumably the idea is they can then just import this repo's JSON files into JS/TS projects? (The reason for asking is tangential to this PR—we were looking for that type of usage tohelp us with our effort to find all current consumers https://www.notion.so/oplabs/Superchain-Registry-Consumers-19c4e715c75341db813f3fed37e03e5c?pvs=4)

Re publishing to npm, I see we have a `publish-bot` job in CI but it seems unused

roninjin10: Yea it's actually pretty common on blockchain industry specifically. Before we switched to using foundry submodules optimism used github installations to install solidity contract packages that were otherwise not.

Realistically most JavaScript consumers will be getting constants like contract addresses from a higher level library like `viem` or `tevm/opstack` as well.

mds1: Ok got it, thank you! I've noted this in the above notion doc and will resolve this comment now

.github/CODEOWNERS
mds1: Per our slack conversation, we can keep the codeowners here as is, and make the above copy tweaks to clarify intent either in this PR or a future one

mds1: Tracking in https://github.com/ethereum-optimism/superchain-registry/issues/164

mds1 (approved):

PR #165, opened 2024-04-05
fix: scrape hardfork times in add-chain.sh

geoknee:
This is an import step since we enabled per chain overrides of hardfork times https://github.com/ethereum-optimism/superchain-registry/pull/101.

### DIFF
--- a/scripts/add-chain.sh
+++ b/scripts/add-chain.sh
@@ -75,6 +75,11 @@ genesis:
     hash: "$(jq -j .genesis.l2.hash $ROLLUP_CONFIG)"
     number: $(jq -j .genesis.l2.number $ROLLUP_CONFIG)
   l2_time: $(jq -j .genesis.l2_time $ROLLUP_CONFIG)
+
+canyon_time: $(jq -j .canyon_time $ROLLUP_CONFIG)
+delta_time: $(jq -j .delta_time $ROLLUP_CONFIG)
+ecotone_time: $(jq -j .ecotone_time $ROLLUP_CONFIG)
+
 EOF
 
 


scripts/add-chain.sh
```diff
@@ -75,6 +75,11 @@ genesis:
     hash: "$(jq -j .genesis.l2.hash $ROLLUP_CONFIG)"
     number: $(jq -j .genesis.l2.number $ROLLUP_CONFIG)
   l2_time: $(jq -j .genesis.l2_time $ROLLUP_CONFIG)
+
+canyon_time: $(jq -j .canyon_time $ROLLUP_CONFIG)
+delta_time: $(jq -j .delta_time $ROLLUP_CONFIG)
+ecotone_time: $(jq -j .ecotone_time $ROLLUP_CONFIG)
```
geoknee, line 81: Desiderata for improvements here:
* we pipe the values through `date -r <timestamp> -u ` and append this as a comment on each line, so it matches: 
https://github.com/ethereum-optimism/superchain-registry/blob/119f493760a5d2de4c9cb878dcfa03077e24d2bc/superchain/configs/sepolia/superchain.yaml#L10-L12
* we iterate through all fields ending in `_time`, to catch future hardforks without needing to update the script
* instead of filling with `null` when the `jq` query fails, we should use `nil` or even better just omit the entire line from the output yaml file. 

IMO we tackle these in a followup issue / PRs. This PR is reasonably urgent, because it will cause tests to incorrectly fail on in flight standard chain PRs.

mds1: Works for me, let's just remember to track this follow up

CONTRIBUTING.md
```diff
@@ -37,6 +37,9 @@ sh scripts/add-chain.sh frontier
 ### 3. Understand output
 The tool will write the following data:
 - The main configuration source, with genesis data, and address of onchain system configuration.
+> **Note**
+> Hardfork override times will be included. For standard chains, you must have all hardforks activated (see the neighbouring superchain.yaml file). It is not possible to override a superchain-wide hardfork time with `nil`.
```
mds1, line 41: So we no longer use the superchain.yaml file as a default and now always "duplicate" the timestamps in the local file?

geoknee: We do use the `superchain.yaml` file as a default, but override it when there are non-nil values in the `chain-name.yaml` file. It is not necessary to duplicate the values in general.

We should only include the timestamp locally when it differs, yes.

mds1: > We should only include the timestamp locally when it differs, yes.

This PR always includes it though IIUC?

geoknee: Yes, correct. I'll file an issue to followup to change that.

scripts/add-chain.sh
geoknee: #166

CONTRIBUTING.md
geoknee: #166

geoknee:
Hardfork overrides can be set in CLI flags without being set in `rollup.json`, meaning we may not pick them up even with the changes from this PR.

mds1 (approved):

PR #172, opened 2024-04-11
fix(hack): skip sepolia devnet

mds1:
I'm not sure what this might break (so let's see if CI still passes), however this is required because without this the `op-upgrade` script in the monorepo does not work properly, for the reason explained in the code comments. We use `op-upgrade` to generate the `input.json` bundles used for upgrades in the superchain-ops repo


superchain/init.go
```diff
@@ -19,6 +19,13 @@ func init() {
 	}
 	// iterate over superchain-target entries
 	for _, s := range superchainTargets {
+		// TODO Hackily skip sepolia devnet for now. This is because we populate `Implementations` and
+		// other maps based on chain ID, but the Sepolia and Sepolia devnet target have the same chain
+		// ID, meaning the Sepolia devnet data overwrites the Sepolia data. The actual fix is to key off
```
trianglesphere, line 24: Do you mean same L1 chain ID?

mds1: Yep exactly

trianglesphere:
CI did fail, but IDK if it's the related or not. I'm fine with this change as long as we document that it's the L1 chain ID & track that we need to re-enable this check at a later point.

mds1:
Thanks @trianglesphere, just pushed https://github.com/ethereum-optimism/superchain-registry/commit/56c51a3a46113f8f974a31b5ae9972b262f1b5df to clarify that it's the L1 chain ID

superchain/init.go
geoknee: Could we move this check down to L107? Most of the init function keys by L1 chain name, so could stay in place and we just skip the implementation part?

mds1: @geoknee Done in 5112ebf

trianglesphere (approved):

PR #175, opened 2024-04-12
Update mainnet semver following MCP upgrade

geoknee:
This PR has two changes
- Updates mainnet semver following MCP upgrade.
- Updates ProxyAdminOwner for sepolia/zora (from #176). This is an expected change as Conduit recently changed the owner from an EOA to a Safe.

We merged #176 into this branch because both of these PRs are required to get CI passing, so we make sure CI passes on this branch before merging into main


validation/superchain-version_test.go
```diff
@@ -111,7 +112,7 @@ func checkSemverForContract(t *testing.T, contractName string, contractAddress *
 	actualSemver, err := getVersionWithRetries(context.Background(), common.Address(*contractAddress), client)
 	require.NoError(t, err, "Could not get version for %s", contractName)
 
-	require.Condition(t, func() bool { return isSemverAcceptable(desiredSemver, actualSemver) },
+	assert.Condition(t, func() bool { return isSemverAcceptable(desiredSemver, actualSemver) },
```
mds1, line 115: Just curious to learn more about the require to assert change here?

mds1 (approved):

validation/superchain-version_test.go
geoknee: You got it, that's exactly my thinking here.

sebastianst (approved):

Issue #168, opened 2024-04-11
Declare validation expectations in a TOML file


geoknee:
This should be more easily parsed by humans than baking into `_test.go` files. 

If possible it should split out expectations per superchain level and possibly also testnet/prod too.

geoknee:
Note that `semver.yaml` is already somewhat in this direction. However:

* it lives in `superchain` module 
* it is consumed by downstream packages such as `op-chain-ops`
* yaml not toml

geoknee:
Suggestion from @mds1 about using scientific notation for numbers to make them easier for humans to parse:

https://github.com/ethereum-optimism/superchain-registry/pull/185#discussion_r1568032360

Issue #151, opened 2024-04-01
Update op-node and op-geth to understand supechainlevel (standard vs frontier)


geoknee:
> I would prefer to make this change in op-node, where it can differentiate based on the new superchain level. The rationale being that I think we ultimately want to deprecate the go exports here that are consumed by op-node/op-geth/etc and instead have op-node/etc consume data from generated TOML files—this should be more flexible and easier to maintain, since TOML can be consumed by other languages without the registry needing bindings/releases for every supported language

_Originally posted by @mds1 in https://github.com/ethereum-optimism/superchain-registry/pull/140#discussion_r1544584285_

geoknee:
Noting that the onboarding flow for chains https://github.com/ethereum-optimism/client-pod/issues/859 requires chains to be run-able in op-node / op-geth _before_ they are classed as standard chains.

PR #183, opened 2024-04-15
add clarifying comments to the validation checks

tessr:
Trying to leave things better than I found them ✨  and adding comments for things I was confused by when reading the validation checks.

### DIFF
--- a/superchain/superchain.go
+++ b/superchain/superchain.go
@@ -499,6 +499,10 @@ func unMarshalSuperchainConfig(data []byte, s *SuperchainConfig) error {
 	return yaml.Unmarshal(data, temp)
 }
 
+// Superchain represents a set of OP Chains which must all upgrade together.
+// We currently have three Superchains: mainnet, which includes OP Mainnet;
+// sepolia, which includes OP Sepolia; and sepolia-dev-0, which is our development
+// Sepolia environment.
 type Superchain struct {
 	Config SuperchainConfig
 
--- a/validation/doc.go
+++ b/validation/doc.go
@@ -1,20 +1,17 @@
 // Package validation contains logic to confirm that an OP Chain is either a valid Standard Chain
 // or a valid Frontier Chain. These checks are provided as Go tests.
 
-// It performs the following checks:
-
-// For Standard Chains:
-
+// It performs the following checks, for all chains:
 // * The Chain ID is unique.
 // * There is a valid, public RPC endpoint which returns a valid response.
+//   (This is checked implicitly; other checks will fail if we don't have a valid RPC endpoint.)
+//
+// For Standard Chains, we perform the following additional checks:
+//
 // * The resource metering config matches the Standard config.
 //   This includes the resource limit, elasticity multiplier, base fee maximum change denominator, minimum base fee, maximum base fee, and maximum gas.
 // * There is a valid gas price oracle, and the parameters (blob base fee scalar and base feed scalar) are within the permitted range.
-// * The L2 output oracle parameters (submission interval, block time and finalization period) all match the Standard config.
+// * The L2 output oracle parameters (submission interval, block time and finalization period) all match the values for OP Mainnet.
 // * The calculated genesis hash matches the hash declared in the chain’s config file.
-
-// For Frontier Chains:
-// * The Chain ID is unique.
-// * There is a valid, public RPC endpoint which returns a valid response.
 
 package validation
--- a/validation/doc.go
+++ b/validation/doc.go
@@ -1,0 +1,20 @@
+// Package validation contains logic to confirm that an OP Chain is either a valid Standard Chain
+// or a valid Frontier Chain. These checks are provided as Go tests.
+
+// It performs the following checks:
+
+// For Standard Chains:
+
+// * The Chain ID is unique.
+// * There is a valid, public RPC endpoint which returns a valid response.
+// * The resource metering config matches the Standard config.
+//   This includes the resource limit, elasticity multiplier, base fee maximum change denominator, minimum base fee, maximum base fee, and maximum gas.
+// * There is a valid gas price oracle, and the parameters (blob base fee scalar and base feed scalar) are within the permitted range.
+// * The L2 output oracle parameters (submission interval, block time and finalization period) all match the Standard config.
+// * The calculated genesis hash matches the hash declared in the chain’s config file.
+
+// For Frontier Chains:
+// * The Chain ID is unique.
+// * There is a valid, public RPC endpoint which returns a valid response.
+
+package validation


validation/doc.go
```diff
@@ -0,0 +1,20 @@
+// Package validation contains logic to confirm that an OP Chain is either a valid Standard Chain
+// or a valid Frontier Chain. These checks are provided as Go tests.
```
geoknee, line 2: 👍

validation/doc.go
```diff
@@ -0,0 +1,20 @@
...
+// * There is a valid, public RPC endpoint which returns a valid response.
+// * The resource metering config matches the Standard config.
+//   This includes the resource limit, elasticity multiplier, base fee maximum change denominator, minimum base fee, maximum base fee, and maximum gas.
+// * There is a valid gas price oracle, and the parameters (blob base fee scalar and base feed scalar) are within the permitted range.
+// * The L2 output oracle parameters (submission interval, block time and finalization period) all match the Standard config.
+// * The calculated genesis hash matches the hash declared in the chain’s config file.
+
+// For Frontier Chains:
+// * The Chain ID is unique.
+// * There is a valid, public RPC endpoint which returns a valid response.
```
geoknee, line 18: I'm slightly hesitant about committing a list of checks here, since it can easily get stale as we add or modify checks and forget to update these comments. But perhaps the advantages outweigh that?

I think in the longer term, doing https://github.com/ethereum-optimism/superchain-registry/issues/168 will help folks see at-a-glance what validation is being done.

If we do want to put this list in, then:

* I would suggest pulling out the uniqueness check into a "for all chains" section
* The "valid RPC endpoint" check is implicit at the moment. One of the other checks will fail if the RPC endpoint doesn't work.

validation/doc.go
```diff
@@ -0,0 +1,20 @@
...
+// It performs the following checks:
+
+// For Standard Chains:
+
+// * The Chain ID is unique.
+// * There is a valid, public RPC endpoint which returns a valid response.
+// * The resource metering config matches the Standard config.
+//   This includes the resource limit, elasticity multiplier, base fee maximum change denominator, minimum base fee, maximum base fee, and maximum gas.
+// * There is a valid gas price oracle, and the parameters (blob base fee scalar and base feed scalar) are within the permitted range.
+// * The L2 output oracle parameters (submission interval, block time and finalization period) all match the Standard config.
```
geoknee, line 13: I think we aren't checking against any Standard config for this one. We actually check against the values used on the OP chain in each superchain https://github.com/ethereum-optimism/superchain-registry/blob/f1be06177cd430401e12992ff7db1ec3608581b7/validation/l2oo_test.go#L61-L66

validation/doc.go
tessr: That's a fair concern! I added this comment because I spent about half an hour trying to understand what each check was doing. These checks are effectively part of our product surface area, so I can definitely imagine partners doing the same. 

I agree that #168 will help a lot once it's done. Two other alternatives in the interim:

1. We could add comments to the top of each file instead, so that it's more likely we'd remember to update them.
2. We could just make the tests _extremely_ clean and easy to read. I know this is fairly subjective but I think we could clean them up a little bit so that the things that are being checked are more obvious.

validation/doc.go
tessr: Ah, good catch, I'll update this (in case we decide to merge it)

validation/doc.go
tessr: Also:

> The "valid RPC endpoint" check is implicit at the moment. One of the other checks will fail if the RPC endpoint doesn't work.

Is this true also for Frontier Chains, which skip most of the other checks?

superchain/superchain.go
```diff
@@ -499,6 +499,10 @@ func unMarshalSuperchainConfig(data []byte, s *SuperchainConfig) error {
 	return yaml.Unmarshal(data, temp)
 }
 
+// Superchain represents a set of OP Chains which must all upgrade together.
```
mds1, line 502: Hmm this currently is not true. Eventually it will be true for premier chains, and maybe for standard chains, but not for frontier chains.

tessr: Interesting. What's the right way to think about this struct? 

If we were trying to make it true... Would each Frontier Chain get its own `Superchain`?

mds1: I'm not sure which is why I did not propose a better phrasing lol 😅 Deferring to @geoknee here since he's more familiar with the code and where that struct is used.

validation/doc.go
geoknee: > Is this true also for Frontier Chains, which skip most of the other checks?

Good question. I recently added this check which applies to all chains regardless of frontier/standard status https://github.com/ethereum-optimism/superchain-registry/pull/184.

superchain/superchain.go
geoknee: I think "superchain" and "supechain target" are used interchangeably. See this PR https://github.com/ethereum-optimism/superchain-registry/pull/186/files.

I think this is an example of where the data structures and terminology was put in place _before_ we got the "new" definitions of standard/frontier/premier, so back when we had a somewhat stricter idea of what a superchain was.

It's still a useful struct for defining sets of chains
In terms of how the struct is actually used there are a few places:
1.  The superchain-wide contract versions check. When we want to loop over each superchain and check the version of the superchain-wide contracts, we can loop over the `Superchains` object and receive a `Superchain` struct inside the loop and make our checks. 

https://github.com/ethereum-optimism/superchain-registry/blob/db623716fc8b7632228b3e8ab0829b05a06a7640/validation/superchain-version_test.go#L50-L51

2. It is exported from the Go module, and used in `op-node` and `op-chain-ops` (for the upgrades). You can search for `superchain.Superchain` to see where. 

For e.g. Frontier chains where we don't expect them to upgrade together, we could consider separating them from the idea of being "in a superchain". This could help reduce confusion, and it might also have a nice side effect of avoiding bloating the downstream modules with a bunch of chains which are mostly irrelevant to them.  See also https://github.com/ethereum-optimism/superchain-registry/issues/151.

validation/doc.go
geoknee: > I added this comment because I spent about half an hour trying to understand what each check was doing. These checks are effectively part of our product surface area, so I can definitely imagine partners doing the same.

Totally hear you on this. I like your option 2.

> We could just make the tests extremely clean and easy to read

I think there is a lot of scope for cleaning up the tests. I can work on this in tandem with https://github.com/ethereum-optimism/superchain-registry/issues/168. Also happy to make a new ticket to track this work, so it can be prioritised.

geoknee: @tessr we recently tidied up the declaration of the tests into a single file https://github.com/ethereum-optimism/superchain-registry/blob/main/validation/validation_test.go . Hopefully it is pretty clear which tests run on which class of chains now.

Issue #96, opened 2024-02-20
Consider extracting exported data into the filesystem itself


geoknee:
This will make it available to future consumers of the repo that want to access such data _not_ through the Go module.

See also #36 and #93

Example: 

This code was added for validation, but included in the `superchain` module in case it may be useful to consumers of that module:
https://github.com/ethereum-optimism/superchain-registry/blob/a87e7b7a2f774051885eb9032f9d0150a4cf4746/superchain/params.go#L40

It should _either_ not be exported at all, and housed in the `validation` module, _or_ extracted into a suitable e.g. `toml` or `json` file.

BlocksOnAChain:
@geoknee ok to close this one?

PR #185, opened 2024-04-16
Add  `standard-config.toml` and use in validation tests

geoknee:
Towards #168 

See also #96

### DIFF
--- a/validation/standard-config.toml
+++ b/validation/standard-config.toml
@@ -1,8 +1,8 @@
 # Standard Chain Config Parameters
 [resource_config]
-# maximum_base_fee = "ffffffffffffffffffffffffffffffff"
 max_resource_limit = 20000000
 elasticity_multiplier = 10
 base_fee_max_change_denominator = 8
 minimum_base_fee = 1000000000
 system_tx_max_gas = 1000000
+maximum_base_fee = "340_282_366_920_938_463_463_374_607_431_768_211_455" # ffffffffffffffffffffffffffffffff"
--- a/validation/standard-config.toml
+++ b/validation/standard-config.toml
@@ -1,6 +1,5 @@
 # Standard Chain Config Parameters
 [resource_config]
-
 max_resource_limit = 20000000
 elasticity_multiplier = 10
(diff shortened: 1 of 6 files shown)


validation/standard-config.toml
```diff
@@ -0,0 +1,9 @@
+# Standard Chain Config Parameters
+[resource_config]
+
```
mds1, line 3: nit to remove the blank line
```suggestion
```

validation/standard-config.toml
```diff
@@ -0,0 +1,9 @@
...
+max_resource_limit = 20000000
+elasticity_multiplier = 10
+base_fee_max_change_denominator = 8
+minimum_base_fee = 1000000000
+system_tx_max_gas = 1000000
+maximum_base_fee = "340_282_366_920_938_463_463_374_607_431_768_211_455"
```
mds1, line 9: not a blocker, but can we support more human readable values? (guessing we have to use a string for the last one)
```suggestion
max_resource_limit = 2e7
elasticity_multiplier = 10
base_fee_max_change_denominator = 8
minimum_base_fee = 1e9
system_tx_max_gas = 1e6
maximum_base_fee = "340_282_366_920_938_463_463_374_607_431_768_211_455" # 2^128 - 1
```

mds1 (commented):
two nits, but love this overall 👌

validation/standard-config.toml
geoknee: This is possible, but some amount of work since it won't work out of the box. We can circle back to this when we tackle things like ranges  `[a,b]` and wildcards `*`.

validation/standard-config.toml
geoknee: [323a96b](https://github.com/ethereum-optimism/superchain-registry/pull/185/commits/323a96b095dc46328cdafdecdbf64091cf8d970b)

validation/standard-config.toml
geoknee: I've added a link to this idea in  #168

refcell (approved):

PR #186, opened 2024-04-16
add definition for superchain target

geoknee:


CONTRIBUTING.md
```diff
@@ -79,12 +79,20 @@ When opening a PR:
 Once the PR is opened, the same automated checks from Step 4 will then run on your PR, and your PR will be reviewed in due course. Once these checks pass the PR will be merged.
 
 ## Adding a superchain target
+A superchain target defines a set of layer 2 chains which share a `SuperchainConfig` contract deployment on layer 1. It is usually named after the layer 1 chain, possibly with an extra identifier to distinguish devnets.
```
mds1, line 82:
```suggestion
A superchain target defines a set of layer 2 chains which share a `SuperchainConfig` and `ProtocolVersions` contract deployment on layer 1. It is usually named after the layer 1 chain, possibly with an extra identifier to distinguish devnets.
```

mds1 (approved):
One nit but lgtm otherwise, cc @blmalone since we talked about superchain target definition recently

PR #188, opened 2024-04-17
Move GPO and L200 params into `standard-config.toml` file

geoknee:
Follows up on #185.

Towards #168 

See also #96


validation/standard-config-mainnet.toml
```diff
@@ -0,0 +1,23 @@
+# Standard Chain Config Parameters
```
mds1, line 1: once https://github.com/ethereum-optimism/specs/pull/134 is merged, we should revisit this file to ensure everything matches

validation/resource-config_test.go
```diff
@@ -36,7 +36,7 @@ func TestResourceConfig(t *testing.T) {
 		actualResourceConfig, err := getResourceConfigWithRetries(context.Background(), common.Address(contractAddress), client)
 		require.NoErrorf(t, err, "RPC endpoint %s: %s", rpcEndpoint)
 
-		require.Equal(t, bindings.ResourceMeteringResourceConfig(StandardConfig.ResourceConfig), actualResourceConfig, "resource config unacceptable")
+		require.Equal(t, bindings.ResourceMeteringResourceConfig(StandardConfigMainnet.ResourceConfig), actualResourceConfig, "resource config unacceptable")
```
mds1, line 39: Why do we hardcode mainnet here? Instead should we have a single `StandardConfig` loaded in init and passed into tests, where the data loaded is based on the L1 chain ID? (or similar deduping instead of repeating the switch-case statements I see in gpo-params-test and l2oo_test)

geoknee: I think for these particular parameters there has not yet been a "need" to separate out by superchain target, so all standard chains are checked against the mainnet expectations.

I think you are right though, we should be loading in the appropriate set of parameters (which exist in both TOML files). I can clean this up.

validation/standard-config.go
```diff
@@ -0,0 +1,49 @@
+package validation
+
+import "math/big"
+
+// StandardConfig is keyed by superchain target, e.g. "mainnet" or "sepolia" or "sepolia-dev-0"
+var StandardConfig map[string]*StandardConfigTy
```
geoknee, line 6: Let's make a new issue to use a string enum type for this key and other places the superchain name is used.

validation/resource-config_test.go
geoknee: [refactor StandardConfig to be a mapping keyed by superchain name](https://github.com/ethereum-optimism/superchain-registry/pull/188/commits/66e8b48c65821d197a3d49570a0502c2e001d20a)

validation/init.go
```diff
@@ -20,11 +20,18 @@ func init() {
...
 	}
+
 	StandardConfig["sepolia"] = new(StandardConfigTy)
 	err = decodeTOMLFileIntoConfig("standard-config-sepolia.toml", StandardConfig["sepolia"])
 	if err != nil {
 		panic(err)
 	}
+
+	StandardConfig["sepolia-dev-0"] = new(StandardConfigTy)
+	err = decodeTOMLFileIntoConfig("standard-config-sepolia-dev-0.toml", StandardConfig["sepolia-dev-0"])
```
geoknee, line 31: I'm including this at the moment for completeness, I do not believe any checks are actually using it presently.

validation/standard-config-mainnet.toml
geoknee: https://github.com/ethereum-optimism/superchain-registry/issues/194

PR #196, opened 2024-04-18
fix(sepolia/metal+mode): update forks overrides

emilianobonassi:
**Description**

Update override for past forks for mode and metal testnets.

**Tests**

No test, used internal deployment/configuration artifacts.

**Additional context**

Part of the on-going MCP upgrade.


emilianobonassi:
ci fails on the mode/metal guardians - taking a look

mds1:
Thanks, I just pushed a commit that adds the the metal sepiloa genesis check back. It still fails, but would like to get that resolved here if we can. cc @geoknee for tips on debugging the genesis failures

mds1 (approved):
Chatted with @geoknee and the genesis check failures will be handled separately, so reverted my last commit

sbvegan (approved):
I can't independently verify these timestamps, so we're going with your internal deployment/configuration artifacts

Issue #177, opened 2024-04-12
bug: fix error handling in `add-chain.sh`


mds1:
https://github.com/ethereum-optimism/superchain-registry/pull/162 added some new calls, and handles errors like this:

```sh
SuperchainConfig=$(cast call $OptimismPortalProxy "superchainConfig()(address)" -r $L1_RPC_URL) || "" # first command could fail if bedrock
```

However, the `||` syntax to catch the error does not seem to work. This may be a shell-specific syntax

cc @geoknee

Issue #166, opened 2024-04-05
Improve hardfork timestamp override handling in add-chain.sh


geoknee:
Desiderata for improvements here:
* we pipe the values through `date -r <timestamp> -u ` and append this as a comment on each line, so it matches: 
https://github.com/ethereum-optimism/superchain-registry/blob/119f493760a5d2de4c9cb878dcfa03077e24d2bc/superchain/configs/sepolia/superchain.yaml#L10-L12
* we iterate through all fields ending in `_time`, to catch future hardforks without needing to update the script
* instead of filling with `null` when the `jq` query fails, we should use `nil` or even better just omit the entire line from the output yaml file. 
* EDIT: if the timestamp matches the superchain default in the neighbouring `superchain.yaml` file, we should NOT add the override at all.

IMO we tackle these in a followup issue / PRs. This PR is reasonably urgent, because it will cause tests to incorrectly fail on in flight standard chain PRs.

mds1:
We should also perform a check that all existing files are consistent with the conventions above and modify them if so

PR #204, opened 2024-04-23
Convert add-chain.sh into a go package

bitwiseguy:
**Description**

This PR converts most of the existing `add-chain.sh` shell script code into go. As the `add-chain.sh` shell script grows more complex, it requires us to cover more edge cases and complex conditional statements/branching logic. Golang has the following advantages compared to shell scripting:
* better suited to handle complex logic
* easier to write unit/e2e tests
* easier to modularize
* provides type safety

**Tests**

An end-to-end test has been added at `add-chain/e2e_test.go`.

**Additional context**

This aims to be backwards compatible with the existing documentation in [CONTRIBUTING.md](https://github.com/ethereum-optimism/superchain-registry/blob/main/CONTRIBUTING.md#2-run-script) that instructs the user to execute `sh ./scripts/add-chain.sh standard`. Therefore, no documentation changes need to be made as a result of this work.

The `add-chain.sh` script becomes a wrapper that invokes two go programs:

*  [1] `superchain-registry/add-chain`
*  [2] [`optimism/op-chain-ops/cmd/registry-data`](https://github.com/ethereum-optimism/optimism/tree/develop/op-chain-ops/cmd/registry-data)

Follow-up work could include moving [2] from the `optimism` monorepo into a new directory within the `superchain-registry` repo: `superchain-registry/registry-data`. This would require removing some light dependencies that program currently has on `op-service` utility functions, but it would allow all of the scripts/programs related to adding a chain to the superchain-registry to live in the `superchain-registry` repo.

**Metadata**

- Fixes https://github.com/ethereum-optimism/superchain-registry/issues/177
- Fixes https://github.com/ethereum-optimism/superchain-registry/issues/166


addchain/contract_addresses.go
```diff
@@ -0,0 +1,136 @@
...
+	SuperchainConfig  = "SuperchainConfig"
+	Guardian          = "Guardian"
+	Challenger        = "Challenger"
+	ProxyAdminOwner   = "ProxyAdminOwner"
+	SystemConfigOwner = "SystemConfigOwner"
+)
+
+func readAddressesFromChain(contractAddresses map[string]string, l1RpcUrl string) error {
+	// SuperchainConfig
+	address, err := executeCommand("cast", []string{"call", contractAddresses[OptimismPortalProxy], "superchainConfig()(address)", "-r", l1RpcUrl})
```
bitwiseguy, line 36: Not sure what is lighter-weight: 
1. using the `cast call` commands within go
2. creating a go-ethereum client and attaching contract abis
3. some other lightweight ethereum client I'm not aware of?

addchain/main.go
```diff
@@ -0,0 +1,117 @@
...
+	// Get the current script's directory
+	scriptDir, err := os.Getwd()
+	if err != nil {
+		fmt.Println("Error getting current directory:", err)
+		os.Exit(1)
+	}
+	superchainRepoPath := filepath.Dir(scriptDir) // Parent directory
+
+	// Load environment variables
+	viper.SetConfigName(".env")             // name of config file (without extension)
```
bitwiseguy, line 38: Not sure if the `viper` package is the best option to pull in env vars. Will look at other options

addchain/genesis.go
```diff
@@ -0,0 +1,32 @@
...
+	// Set current working directory to the monorepo directory
+	if err := os.Chdir(monorepoDir); err != nil {
+		return fmt.Errorf("failed to change directory to monorepo: %w", err)
+	}
+
+	cmd := exec.Command("go", "run", "./op-chain-ops/cmd/registry-data",
+		"--l2-genesis", genesisConfig,
+		"--bytecodes-dir", filepath.Join(superchainRepoPath, "superchain", "extra", "bytecodes"),
+		"--output", filepath.Join(genesisDir, chainName+".json.gz"))
+
```
geoknee, line 26: Should we consider putting this new package into the `registry-data` command? If we did so, we might be freer to use libraries such as go-ethereum in order to get the job done. 

We want to avoid pulling such dependencies into the `superchain` module, since that is upstream of `op-geth`.

addchain/utils.go
```diff
@@ -0,0 +1,38 @@
...
+
+func inferRpcUrl(superchainTarget string) (string, error) {
+	switch superchainTarget {
+	case "mainnet":
+		return "https://ethereum-mainnet-rpc.allthatnode.com", nil
+	case "sepolia":
+		return "https://ethereum-sepolia-rpc.allthatnode.com", nil
+	}
+	return "", fmt.Errorf("Unsupported Superchain Target %s\n", superchainTarget)
+}
```
geoknee, line 25: See https://github.com/ethereum-optimism/superchain-registry/pull/162#discussion_r1551458856

add-chain/go.mod
```diff
@@ -0,0 +1,33 @@
+module github.com/ethereum-optimism/superchain-registry/add-chain
+
+go 1.21.9
```
trianglesphere, line 3: Does each component have it's own go.mod file? It seems a bit easier to only have one.

add-chain/testdata/monorepo/deployments/.chainId
```diff
@@ -0,0 +1 @@
+11155111
```
geoknee, line 1: There is a lot of data committed in this folder, most of which I think we don't need. We only need the address from each deployment. 

I suggest we commit test data in the newer deployment artifact system, see https://github.com/ethereum-optimism/superchain-registry/pull/200 . It's just a single json file with a list of addresses. This will save us a a lot of lines.

If we really want to test agasint legacy deployment artifacts formats, we can strip the artifacts down so they only contain the information we are looking for.

geoknee: `genesis.json` is also quite a huge file -- it is in the `testdata` folder, so are we sure this won't mean downstream dependencies need to download it?

Another issue is that this file is also in a format which is constantly changing. See this comment https://github.com/ethereum-optimism/optimism/pull/10231#discussion_r1577627727 .

add-chain/testdata/superchain/configs/sepolia/expected.yaml
```diff
@@ -0,0 +1,19 @@
+name: awesomechain
+chain_id: 42069
+public_rpc: http://awe.some.rpc
+sequencer_rpc: http://awe.some.seq.rpc
+explorer: https://awesomescan.org
+superchain_level: 2
+batch_inbox_addr: "0xff00000000000000000000000000000000042069"
+genesis:
```
geoknee, line 8: I think this file should have some spaces between sections, see the exising config yaml files.

add-chain/testdata/superchain/configs/sepolia/expected.yaml
```diff
@@ -0,0 +1,19 @@
...
+  l1:
+    hash: "0x1fc104fe244042459314bc1d51a230320c64a3d47cd06b7998aa3f57b9b0fc79"
+    number: 5753349
+  l2:
+    hash: "0x0ceefff2113580e4d665c324661e4982a824025bf97285cf09a52715b6f89083"
+    number: 0
+  l2_time: 1713792864
+canyon_time: 0
+delta_time: null
+ecotone_time: null
```
geoknee, line 18: This is fine for now, but will change when we tackle https://github.com/ethereum-optimism/superchain-registry/issues/166

add-chain/utils.go
```diff
@@ -0,0 +1,38 @@
...
+
+func inferRpcUrl(superchainTarget string) (string, error) {
+	switch superchainTarget {
+	case "mainnet":
+		return "https://ethereum-mainnet-rpc.allthatnode.com", nil
+	case "sepolia":
+		return "https://ethereum-sepolia-rpc.allthatnode.com", nil
+	}
+	return "", fmt.Errorf("unsupported Superchain Target %s", superchainTarget)
+}
```
geoknee, line 25: We should get these RPC urls from this file:

https://github.com/ethereum-optimism/superchain-registry/blob/f56b630610ea41bb7e23d262de6f7cbb897d6ca4/superchain/configs/mainnet/superchain.yaml#L4-L5

add-chain/go.mod
geoknee: One advantage of having separate modules is that it can avoid dependency cycles. The `add-chain` module is free to depend on `go-ethereum`, whereas doing that inside `superchain` is probably a problem since `op-geth` depends on `superchain`. https://www.notion.so/oplabs/Superchain-registry-module-structure-4b81c28b6a1e4e169056378eacdd9f14

add-chain/e2e_test.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	var testYaml RollupConfig
+	if err := yaml.Unmarshal(testBytes, &testYaml); err != nil {
+		return false, err
+	}
+
+	return reflect.DeepEqual(expectYaml, testYaml), nil
```
geoknee, line 97: Will this check fail on whitespace changes? Maybe a nit, but I think it's nice for us to be strict about the overal format of the document. Could we just compare the raw bytes of the file?

add-chain/contract_addresses.go
```diff
@@ -0,0 +1,159 @@
...
+	address, err := executeCommand("cast", []string{"call", contractAddresses[OptimismPortalProxy], "superchainConfig()(address)", "-r", l1RpcUrl})
+	address = strings.Join(strings.Fields(address), "") // remove whitespace
+	if err != nil || address == "" || address == "0x" {
+		contractAddresses[SuperchainConfig] = ""
+	} else {
+		contractAddresses[SuperchainConfig] = address
+	}
+
+	// Guardian
+	address, err = executeCommand("cast", []string{"call", contractAddresses[SuperchainConfig], "guardian()(address)", "-r", l1RpcUrl})
```
geoknee, line 46: Is it worth factoring this into a `castCall(address string, calldata string)` helper?

add-chain/.env.test
```diff
@@ -0,0 +1,10 @@
+# Required for both Standard Chains and Frontier Chains
```
geoknee, line 1: Should this file live in the testdata dir?

add-chain/go.mod
geoknee, line 3: @bitwiseguy shall we harmonize the `go` version here across all modules to `1.21` for now?

geoknee (changes requested):
Looking good so far @bitwiseguy. Left a few comments. 

I ran this on my local test chain files and got this output:

```
Adding standard chain to superchain-registry...

Chain Name:                     awesomechain
Superchain target:              sepolia
Superchain-registry repo dir:   /Users/georgeknee/code/ethereum-optimism/superchain-registry
Reading from monrepo directory: ../optimism
With deployments directory:     ../optimism/packages/contracts-bedrock/deployments/devnetL1
Rollup config:                  ../optimism/op-node/rollup.json
Genesis config:                 ../optimism/op-node/genesis.json
Public RPC endpoint:            http://awe.some.rpc
Sequencer RPC endpoint:         http://awe.some.seq.rpc
Block Explorer:                 https://awesomescan.org

Attempting to read from ../optimism/op-node/rollup.json
Rollup config successfully constructed
Genesis system config written to: /Users/georgeknee/code/ethereum-optimism/superchain-registry/superchain/extra/genesis-system-configs/sepolia/awesomechain.json
Rollup config written to: /Users/georgeknee/code/ethereum-optimism/superchain-registry/superchain/configs/sepolia/awesomechain.yaml
Contract addresses read from deployments directory: ../optimism/packages/contracts-bedrock/deployments/devnetL1
failed to read addresses from chain: could not retrieve address for Guardian
*********************
FAILED: add-chain
exit status 1
```

addchain/contract_addresses.go
geoknee: I think this is fine as is, we can circle back to this question in a future refactor when we pull in the `registry-data` tool.

add-chain/utils.go
bitwiseguy: Fixed here: https://github.com/ethereum-optimism/superchain-registry/pull/204/commits/7205dbe433e9dcf31c3376e6104fb1f08e9c62ed

addchain/utils.go
bitwiseguy: Fixed here: https://github.com/ethereum-optimism/superchain-registry/pull/204/commits/7205dbe433e9dcf31c3376e6104fb1f08e9c62ed

addchain/main.go
bitwiseguy: This is working so will leave as is. Can replace in a future PR if we want

add-chain/go.mod
bitwiseguy: Fixed in https://github.com/ethereum-optimism/superchain-registry/pull/204/commits/1b2967aaae1057c94f408e931162cbcf28a2b043

add-chain/testdata/monorepo/deployments/.chainId
bitwiseguy: Removed individual contract json files from `testdata` and are now reading contract addresses from `.deploy` file instead: https://github.com/ethereum-optimism/superchain-registry/pull/204/commits/61ea05f6243255f92b1fac7539c84666d6b7dd29

add-chain/.env.test
bitwiseguy: Fixed in https://github.com/ethereum-optimism/superchain-registry/pull/204/commits/cb6864a042dd45cc0615de28402658cecd361bf3

add-chain/contract_addresses.go
bitwiseguy: Refactored here: https://github.com/ethereum-optimism/superchain-registry/pull/204/commits/2fd6532f5eecadaa8fca62395d7301b23a1bc2b4

bitwiseguy:
@geoknee regarding your [test failure ](https://github.com/ethereum-optimism/superchain-registry/pull/204#pullrequestreview-2024516646): I'm hoping [this commit](https://github.com/ethereum-optimism/superchain-registry/pull/204/commits/ccd2fdd9e51dc6e45030d4399ca1b9f9e73634ae) will fix it because I believe the smart contract function name is case sensitive. Let me know if it helps.

add-chain/testdata/monorepo/deployments/.chainId
bitwiseguy: `genesis.json` is actually not used by the `add-chain` module. Rather, it is needed for the `registry-data` module. Removed that file in this commit: https://github.com/ethereum-optimism/superchain-registry/pull/204/commits/c331023f015b721710d745e8c55f945271cf7c93

add-chain/testdata/superchain/configs/sepolia/expected.yaml
bitwiseguy: Fixed here: https://github.com/ethereum-optimism/superchain-registry/pull/204/commits/7013cd31745c57b2b069cf957329c3d8561445df

add-chain/e2e_test.go
bitwiseguy: Updated in https://github.com/ethereum-optimism/superchain-registry/pull/204/commits/bcdfb7cb102f0c8e752794ff7b51a3af081154a2

geoknee:
> @geoknee regarding your [test failure ](https://github.com/ethereum-optimism/superchain-registry/pull/204#pullrequestreview-2024516646): I'm hoping [this commit](https://github.com/ethereum-optimism/superchain-registry/pull/204/commits/ccd2fdd9e51dc6e45030d4399ca1b9f9e73634ae) will fix it because I believe the smart contract function name is case sensitive. Let me know if it helps.

This ended up being a problem with my local setup. I fixed that and the script worked!

geoknee:
Looks like tests failed at least once because of rate limiting. To combat that we could follow this pattern:
 
https://github.com/ethereum-optimism/superchain-registry/blob/1e360a88a9ae71c730ee633f9b4bba0d68382168/validation/l2oo_test.go#L101-L104

add-chain/contract_addresses.go
```diff
@@ -0,0 +1,160 @@
...
+
+	"github.com/ethereum-optimism/superchain-registry/superchain"
+)
+
+type AddressData struct {
+	Address string `json:"address"`
+}
+
+var (
+	// Addresses to retrieve from JSON
```
trianglesphere, line 17: Can we add the DA Challenge contract? It's also possible to merge this first before adding support for plasma mode.

add-chain/.gitignore
```diff
@@ -0,0 +1 @@
+testdata/**/awesomechain*
```
geoknee, line 1: Perhaps a comment about why this is ignored?

superchain/superchain.go
```diff
@@ -56,18 +79,18 @@ const (
 	Frontier SuperchainLevel = 1
 )
 
-type ChainConfig struct {
+type RollupConfig struct {
```
geoknee, line 82: Why was it necessary to rename this?

superchain/superchain.go
```diff
@@ -485,7 +574,7 @@ type SuperchainConfig struct {
 	SuperchainConfigAddr *Address `yaml:"superchain_config_addr,omitempty"`
 
 	// Hardfork Configuration. These values may be overridden by individual chains.
-	hardForkDefaults HardForkConfiguration
+	HardForkDefaults HardForkConfiguration
```
geoknee, line 577: Was it necessary to make this an exported field? The intention was to make this unexported, to force consumers to read the overriden values from each chain.

add-chain/main.go
```diff
@@ -0,0 +1,134 @@
...
+		return fmt.Errorf("superchain target directory not found. Please follow instructions to add a superchain target in CONTRIBUTING.md")
+	}
+
+	rollupConfig, err := constructRollupConfig(rollupConfigPath, chainName, publicRPC, sequencerRPC, explorer, superchainLevel)
+	if err != nil {
+		return fmt.Errorf("failed to construct rollup config: %w", err)
+	}
+
+	targetFilePath := filepath.Join(targetDir, chainName+".yaml")
+	err = writeChainConfig(rollupConfig, targetFilePath, superchainRepoPath, superchainTarget)
```
geoknee, line 107: Seems to be a mismatch here between writeChainConfig and rollupConfig, see other comments about naming these types.

add-chain/contract_addresses.go
```diff
@@ -0,0 +1,160 @@
...
+	if err != nil {
+		contractAddresses[SuperchainConfig] = ""
+	} else {
+		contractAddresses[SuperchainConfig] = address
+	}
+
+	// Guardian
+	address, err = castCall(contractAddresses[SuperchainConfig], "guardian()(address)", l1RpcUrl)
+	if err != nil {
+		address, err = castCall(contractAddresses[OptimismPortalProxy], "guardian()(address)", l1RpcUrl)
```
geoknee, line 48: This error handling could be improved, I think. The cast call above could fail for a number of reasons, and probably only some of them indicate the reason we want to try the second cast call. Some comments explaining which versions of the contracts require which calls would help readability and also help us maintain this section of the code going forward. 


I guess we don't have test coverage for both code paths. Do you think that is something we should add?

geoknee: I think the ideal state is that we detect the version of the contracts in use, and pull in versioned Go bindings that map to those contract ABIs, and then make the appropriate calls using the go-ethereum library plus retry logic. Not a blocker for this PR, but curious to know your thoughts.

add-chain/testdata/monorepo/op-node/rollup.json
```diff
@@ -0,0 +1,34 @@
+{
```
geoknee, line 1: Not sure if we discussed this already, but did we decide not to include `genesis.json` input files and associated checks in the tests? If so, perhaps best to record the reason why in a readme. The readme can also explain choices such as the separate module.

geoknee (approved):
This looks good to me but left quite a few comments.

superchain/superchain.go
bitwiseguy: Not necessary, just seemed like a more descriptive name to distinguish between a "SuperchainConfig" and a "RollupConfig". "ChainConfig" felt ambiguous to me but can revert if you want

superchain/superchain.go
bitwiseguy: Originally was reading these defaults from within the `add-chain` module. But refactored and am using `rollupConfig.SetDefaultHardforkTimestampsToNil` method to read the defaults so can change `HardForkDefaults` be to unexported

add-chain/main.go
bitwiseguy: I'd prefer to change this function to `writeRollupChainConfig` and keep the struct name `RollupConfig` over `ChainConfig`. Are you good with that?

add-chain/contract_addresses.go
bitwiseguy: I'm not sure which versions of contracts expect one call over the other. How can I best determine that information? 

This logic mirrors what is in the existing `add-chain.sh` script, with the added benefit that the conditional to trigger the second `castCall` will more reliably get executed (due to more robust error-handling), compared to the following command, which would only execute the first `cast call`: https://github.com/ethereum-optimism/superchain-registry/blob/main/scripts/add-chain.sh#L126

bitwiseguy: Agreed on the ideal state and addressing that in a follow-up PR

superchain/superchain.go
geoknee: I think "rollup config" is already used for `rollup.json`. I don't feel hugely strongly but I think the best thing to do is pinch this renaming change off into a different PR. If you revert it, it should reduce the diff here, which I think is helpful. Optional then whether you want to make another PR with the change and we can get some more opinions.

add-chain/main.go
geoknee: See other comment, let's make it consistent one way or the other for sure. I think I prefer sticking with `ChainConfig` for now.

add-chain/contract_addresses.go
geoknee: Yeah what you have is a definite improvement and should close the associated issue. Agreed about improving in followup PRs.

About contract versions, I don't know the answer to your question. Perhaps @mds1 can pitch in?

mds1: Agreed with this plan to key off the version in use. We need the mappings (from monorepo release tag to contract version) for bytecode verification also, and since we are doing that in a follow up PR, I will make a note to get this done instead of providing it right now—let me know if that works

add-chain/testdata/monorepo/deployments/.deploy
```diff
@@ -0,0 +1,11 @@
+{
+	"AddressManager": "0xBBCfC68f590ae60Feb1551F4C639E3Eab5ea7BA9",
+	"L1CrossDomainMessengerProxy": "0x6b95418119Aa1bc41D521377D98E6c367817A755",
+	"L1ERC721BridgeProxy": "0x460Cb843a61CeB6cFB3BC3373FC405744f5be4ee",
+	"L1StandardBridgeProxy": "0xC4Ed1Ae4671B38A44B3cea27E1F7462cFe6B1A87",
+	"L2OutputOracleProxy": "0x3F8A862E63E759a77DA22d384027D21BF096bA9E",
```
mds1, line 6: Can we have a second test for chains on fault proofs? These would have no L2OO, but would have other contracts like the Dispute Game Factory. OP Sepolia can be used as a reference here for now

add-chain/testdata/superchain/extra/addresses/sepolia/expected.json
```diff
@@ -0,0 +1,16 @@
+{
+  "AddressManager": "0xbbcfc68f590ae60feb1551f4c639e3eab5ea7ba9",
+  "Challenger": "0xe8D525b30dD4C406570d47396b5b237DB30F13F1",
+  "Guardian": "0xe8D525b30dD4C406570d47396b5b237DB30F13F1",
+  "L1CrossDomainMessengerProxy": "0x6b95418119aa1bc41d521377d98e6c367817a755",
+  "L1ERC721BridgeProxy": "0x460cb843a61ceb6cfb3bc3373fc405744f5be4ee",
+  "L1StandardBridgeProxy": "0xc4ed1ae4671b38a44b3cea27e1f7462cfe6b1a87",
+  "L2OutputOracleProxy": "0x3f8a862e63e759a77da22d384027d21bf096ba9e",
```
mds1, line 8: same comment on a fault proof variant

add-chain/chain_config.go
```diff
@@ -0,0 +1,121 @@
...
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/ethereum-optimism/superchain-registry/superchain"
+	"gopkg.in/yaml.v3"
+)
+
+// constructRollupConfig creates and populates a RollupConfig struct by reading from an input file and
+// explictly setting some additional fields to input argument values
```
mds1, line 16:
```suggestion
// explicitly setting some additional fields to input argument values
```

scripts/add-chain.sh
```diff
@@ -1,161 +1,15 @@
...
-
-# create genesis-system-config data
-# (this is deprecated, users should load this from L1, when available via SystemConfig).
-mkdir -p $SUPERCHAIN_REPO/superchain/extra/genesis-system-configs/$SUPERCHAIN_TARGET
-jq -r .genesis.system_config $ROLLUP_CONFIG > $SUPERCHAIN_REPO/superchain/extra/genesis-system-configs/$SUPERCHAIN_TARGET/$CHAIN_NAME.json
-
-
 # create extra genesis data
 mkdir -p $SUPERCHAIN_REPO/superchain/extra/genesis/$SUPERCHAIN_TARGET
 cd $MONOREPO_DIR
```
mds1, line 15: UX nit that wasn't changed in this PR so feel free to postpone—we should save off the current dir, then `cd $originalDir` back to that after the below `go run` command. If the `go run` command can fail, we should put that `cd $originalDir` in a trap so it always runs

superchain/superchain.go
bitwiseguy: Reverted

add-chain/main.go
bitwiseguy: Ok will stick with original naming as you suggest

add-chain/contract_addresses.go
bitwiseguy: That's perfect. Thanks @mds1

add-chain/testdata/monorepo/op-node/rollup.json
bitwiseguy: I'm not sure what variations of `genesis.json` and `rollup.json` are expected. Are you able to point me to any reference material that shows all the variations of those inputs that `add-chain` should be able to successfully process?

geoknee: I don't know about reference material, but here is a doc which covers some genesis changes impacting the `registry-data` tool which we could in theory track through the tests in this module. It's another thing we could circle back on.

bitwiseguy: I think your comment is missing a link? Agreed that we could circle back on this. This PR mainly aimed to convert the existing code from shell script to go to allow us to more easily handle some of these types of changes. Rather than making all of the upgrades in the same PR, I think it would be best to address those upgrades as follow-up work

scripts/add-chain.sh
geoknee: I think the directory is only changed in a subshell when this script is run with `sh`? If I'm right then we don't need to make any changes here.

add-chain/testdata/monorepo/op-node/rollup.json
geoknee: Agreed. Here is the link I forgot https://www.notion.so/oplabs/Superchain-Genesis-Hash-checks-60e173e9fb0c43e88a81d0666cb3d582

add-chain/testdata/monorepo/deployments/.deploy
mds1: > Can definitely add this test. Is [this](https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/extra/addresses/sepolia/op.json) the file you want to use as a reference? Note that it does include a `L2OutputOracleProxy` address.

Good call. I opened https://github.com/ethereum-optimism/superchain-registry/pull/218 to fix this, but it's actually failing, so we'll need a fix on the checks

> Would you expect the `add-chain` script to behave differently for chains on fault proofs other than the output file in `superchain/extra/addresses/<chain_name>.json` would have a different set of contracts?

I don't think so, but the checks required are different (e.g. we have no L2OO params to check, and we'll need to define standard config params for fault proofs)

scripts/add-chain.sh
mds1: Ah you might be right—I recall running a bash script recently were it was changed, but might be misremembering that situation, and I did not test this :)

add-chain/testdata/monorepo/op-node/rollup.json
bitwiseguy: Issue created [here](https://github.com/ethereum-optimism/superchain-registry/issues/220) to capture that work.

go.work
```diff
@@ -1,6 +1,7 @@
-go 1.21
+go 1.21.9
 
 use (
-    ./superchain
-    ./validation
+	./add-chain
```
geoknee, line 4: We should add the new module to the CI linting and testing steps.

add-chain/contract_addresses.go
bitwiseguy: Created issue [here](https://github.com/ethereum-optimism/superchain-registry/issues/221) to capture that work

add-chain/testdata/monorepo/deployments/.deploy
bitwiseguy: Issue created here to capture the work involved with this update: https://github.com/ethereum-optimism/superchain-registry/issues/222. Will address in a follow-up PR

add-chain/testdata/superchain/extra/addresses/sepolia/expected.json
bitwiseguy: Responded in other comment

mds1 (approved):

PR #205, opened 2024-04-24
chore: update OP Sepolia proxy admin owner

mds1:
This change was made as part of our effort to align the OP Sepolia config with the OP Mainnet config: https://github.com/ethereum-optimism/superchain-ops/tree/main/tasks/sep/006-transfer-admin


sbvegan (approved):
`ProxyAdmin` state change matches up. lgtm

PR #207, opened 2024-04-25
SubmissionInterval not more equality check

blmalone:
Output frequency can be 24 hours or less: https://specs.optimism.io/protocol/configurability.html#policy-parameters 

That's **86400** seconds.


blmalone:
@geoknee removing all the exceptions cause some checks to fail on `finalizationPeriodSeconds`.

PR #208, opened 2024-04-25
standard-config: loosen bounds on `l2oo.submissionInterval`

geoknee:
Output frequency can be 24 hours or less: https://specs.optimism.io/protocol/configurability.html#policy-parameters

### DIFF
--- a/validation/l2oo_test.go
+++ b/validation/l2oo_test.go
@@ -19,6 +19,12 @@ import (
 	"github.com/ethereum/go-ethereum/ethclient"
 )
 
+type L2OOParams struct {
+	SubmissionInterval        *big.Int `toml:"submission_interval"`         // Interval in blocks at which checkpoints must be submitted.
+	L2BlockTime               *big.Int `toml:"l2_block_time"`               // The time per L2 block, in seconds.
+	FinalizationPeriodSeconds *big.Int `toml:"finalization_period_seconds"` // The minimum time (in seconds) that must elapse before a withdrawal can be finalized.
+}
+
 func TestL2OOParams(t *testing.T) {
 	isExcluded := map[uint64]bool{
 		999999999: true, // sepolia/zora    Incorrect submissionInterval, wanted 120 got 180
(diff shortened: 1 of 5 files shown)


validation/l2oo_test.go
```diff
@@ -19,33 +19,23 @@ import (
...
-		8866:      true, // mainnet/superlumio Incorrect submissionInterval
-	}
-
-	checkEquality := func(a, b *big.Int) func() bool {
-		return (func() bool { return (a.Cmp(b) == 0) })
-	}
-
-	incorrectMsg := func(name string, want, got *big.Int) string {
-		return fmt.Sprintf("Incorrect %s, wanted %d got %d", name, want, got)
+		999999999: true, // sepolia/zora  Incorrect finalizationPeriodSeconds, 604800 is not within bounds [12 12]
```
geoknee, line 30: `finalizationPeriodSeconds` on Sepolia is an obvious place to look next. I suspect being strict about `12` is not what we want to do.

validation/l2oo_test.go
```diff
@@ -19,33 +19,23 @@ import (
 	"github.com/ethereum/go-ethereum/ethclient"
 )
 
+type L2OOParams struct {
+	SubmissionInterval        *big.Int
+	L2BlockTime               *big.Int
+	FinalizationPeriodSeconds *big.Int
```
blmalone, line 25: Do we want to change the name of this variable to be challenge period? Maybe not as I know the on-chain function is `finalizationPeriodSeconds()`

blmalone (approved):
LGTM 👍🏻

validation/l2oo_test.go
geoknee: (So keep both names around, one on the data scraping side, one on the spec side).

bitwiseguy (approved):
Lgtm

validation/l2oo_test.go
bitwiseguy: Where is the `604800` value coming from?

validation/l2oo_test.go
mds1: Just unresolved for visibility: It's a bit unfortunate there's two names, I'm indifferent to the name used as long as it's clear they are the same. Maybe just a comment in the toml file is sufficient:
```toml
challenge_period_seconds = [12, 12] # i.e. finalization period
```

geoknee: [9cb8ac3](https://github.com/ethereum-optimism/superchain-registry/pull/208/commits/9cb8ac3b779d189455dd9c46ad01439cf1edcc6a)

validation/l2oo_test.go
geoknee: 604800 is being used on sepolia/zora. It's one week's worth of seconds.

PR #210, opened 2024-04-26
validation: add `TestGenesisHashAgainstRPC`

geoknee:
We are already running a check on the genesis hash, comparing the declared genesis hash submitted in the chain config with one compute by reconstructing the genesis block from the "extra" submitted data. That test has some exclusions which are currently being investigated. 

This adds another sanity check, that the declared genesis hash matches what you get over the RPC.

### DIFF
--- a/validation/superchain-genesis_test.go
+++ b/validation/superchain-genesis_test.go
@@ -82,5 +82,4 @@ func TestGenesisHashAgainstRPC(t *testing.T) {
 			checkOPChainHashAgainstRPC(t, chain)
 		})
 	}
-
 }
--- a/validation/superchain-genesis_test.go
+++ b/validation/superchain-genesis_test.go
@@ -55,7 +55,6 @@ func TestGenesisHash(t *testing.T) {
 
 func TestGenesisHashAgainstRPC(t *testing.T) {
 	isExcluded := map[uint64]bool{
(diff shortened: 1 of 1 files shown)


validation/superchain-genesis_test.go
```diff
@@ -49,3 +52,35 @@ func TestGenesisHash(t *testing.T) {
 		})
 	}
 }
+
+func TestGenesisHashAgainstRPC(t *testing.T) {
+	isExcluded := map[uint64]bool{
+		10:       true, // mainnet/op (Genesis Block Hash declared as 0xdbf6a80fef073de06add9b0d14026d6e5a86c85f6d102c36d3d8e9cf89c2afd3, but RPC returned 0x7ca38a1916c42007829c55e69d3e9a73265554b586a499015373241b8a3fa48b)
```
geoknee, line 58: This appears to be a known exception, see https://github.com/ethereum-optimism/op-geth/blob/4fb6cf678c6cd0b5189332517110b48615401b16/core/superchain.go#L90

trianglesphere: The reason it's excluded is that the `genesis` in rollup.config is not always block 0. It's the op-node genesis, but when there was the bedrock transition, the genesis in the config is not block number 0, but a higher block.

We can either skip transitioned networks or fetch the given genesis block number and only skip networks without a public endpoint

trianglesphere (approved):

PR #211, opened 2024-04-26
update expected version of optimism_portal

geoknee:
Gets the Go CI checks  passing. 

Following [FPAC Boogaloo.](https://www.notion.so/oplabs/FPAC-Sepolia-Testnet-Upgrade-Mk-2-Boogaloo-fa4fe2ea61cf4abf9a46a0b141700438?p=2d8ef6ee892347509c552477987f676c&pm=s)


mds1:
In check-security-configs, the portal's `GUARDIAN()` and `SYSTEM_CONFIG()` methods are now removed and replaced with `guardian()` and `systemConfig()`, so we also need to account for that

PR #224, opened 2024-05-07
Add support for plasma chains

bitwiseguy:
**Description**

Updates the `superchain.ChainConfig` struct and the `add-chain` package to process `plasma` fields if they exist.

**Tests**

Modified `add-chain/e2e_test.go` so that it is more modular and now tests a vanilla "baseline" chain and a "plasma" chain. Added test files in `add-chain/testdata` that define input files and expected output files.

Previously we only tested a vanilla "baseline" chain. Since we now have two chain configurations we are testing, all previous `testdata/` files have been renamed to include the suffix `_baseline` to distinguish from `_plasma` files:
* `rollup.json` --> `rollup_baseline.json`
* `expected.json` --> `expected_baseline.json`
* etc...

**Metadata**

- Fixes https://github.com/ethereum-optimism/client-pod/issues/801


add-chain/testdata/superchain/extra/addresses/sepolia/expected_plasma.json
```diff
@@ -0,0 +1,17 @@
+{
+  "AddressManager": "0xbbcfc68f590ae60feb1551f4c639e3eab5ea7ba9",
+  "Challenger": "0xe8D525b30dD4C406570d47396b5b237DB30F13F1",
+  "DAChallengeAddress": "0x3333333333333333333300000000000000000000",
```
bitwiseguy, line 4: @trianglesphere the `DAChallengeAddress` is the only difference between this .json file produced by a vanilla chain versus a `plasma` chain. Is that correct?

add-chain/testdata/superchain/configs/sepolia/expected_plasma.yaml
```diff
@@ -0,0 +1,24 @@
...
+genesis:
+  l1:
+    hash: 0x1fc104fe244042459314bc1d51a230320c64a3d47cd06b7998aa3f57b9b0fc79
+    number: 5753349
+  l2:
+    hash: 0x0ceefff2113580e4d665c324661e4982a824025bf97285cf09a52715b6f89083
+    number: 0
+  l2_time: 1713792864 # Mon 22 Apr 2024 13:34:24 UTC
+
+plasma:
```
bitwiseguy, line 20: @trianglesphere this additional `plasma` field and the fields nested underneath it are the only differences between this .yaml file produced by a vanilla chain and a plasma chain. Is that correct?

add-chain/testdata/superchain/extra/addresses/sepolia/expected_plasma.json
trianglesphere: This would be the only address difference.

add-chain/testdata/superchain/configs/sepolia/expected_plasma.yaml
trianglesphere: Yes that is correct. Do we want to duplicate the `da_challenge_contract_address`?

bitwiseguy: Are you saying we should de-duplicate by removing the address from this .yaml file and just rely on the address being in the [.json file](https://github.com/ethereum-optimism/superchain-registry/blob/7f3dc3e4d185465712835d8505e20bc6f7bcc17c/add-chain/testdata/superchain/extra/addresses/sepolia/expected_plasma.json#L4) along with the other contract addresses?

trianglesphere: Yea, I think so. We do that with the other addresses

superchain/superchain.go
```diff
@@ -99,10 +99,52 @@ type ChainConfig struct {
 	// This matches the resource filename, it is not encoded in the config file itself.
 	Chain string `yaml:"-"`
 
+	// UsePlasma is activated when the chain is in plasma mode.
+	UsePlasma    *bool        `json:"use_plasma,omitempty" yaml:"use_plasma,omitempty"`
+	Plasma       PlasmaConfig `json:"plasma,omitempty" yaml:"plasma,omitempty"`
+	PlasmaConfig `yaml:",inline"`
```
geoknee, line 105: Will this result in plasma config being duplicated? From what I can tell, after `CheckPlasma()` is called, we will have `nil` values embedded / inlined, and non-nil values nested under `Plasma`. I wonder if there is a better approach?

superchain/superchain.go
```diff
@@ -99,10 +99,52 @@ type ChainConfig struct {
...
+	DAChallengeAddress *Address `json:"da_challenge_contract_address,omitempty" yaml:"da_challenge_contract_address,omitempty"`
+	// DA challenge window value set on the DAC contract. Used in plasma mode
+	// to compute when a commitment can no longer be challenged.
+	DAChallengeWindow *uint64 `json:"da_challenge_window,omitempty" yaml:"da_challenge_window,omitempty"`
+	// DA resolve window value set on the DAC contract. Used in plasma mode
+	// to compute when a challenge expires and trigger a reorg if needed.
+	DAResolveWindow *uint64 `json:"da_resolve_window,omitempty" yaml:"da_resolve_window,omitempty"`
+}
+
+func (c *ChainConfig) CheckPlasma() error {
```
geoknee, line 121: Is it a bit surprising that a function called `CheckFoo` actually ends up mutating the receiver? See other comment.

geoknee (changes requested):
Looks good to me, one thing to address around unmarshalling / exporting the new data.

superchain/superchain.go
bitwiseguy: [This](https://github.com/ethereum-optimism/superchain-registry/blob/5c3f1ca3313bfe61bc6c112d2fb6828171293795/add-chain/testdata/superchain/configs/sepolia/expected_plasma.yaml) is what the .yaml will look like.

The code that removes the inline fields is [here](https://github.com/ethereum-optimism/superchain-registry/blob/5c3f1ca3313bfe61bc6c112d2fb6828171293795/superchain/superchain.go#L138-L142). This works because these fields types are pointers and the yaml tag includes the `omitempty` directive

superchain/superchain.go
bitwiseguy: Would it suffice to just rename this function to better describe its functionality? Perhaps `PlasmaCheckAndTransform`? If not, would you prefer to have separate structs for `Unmarshaling` from the `rollup.json` versus writing to into the .yaml file?

superchain/superchain.go
geoknee: I'm more concerned about the Go types exported from the `superchain` module. Can you confirm what they will look like?

superchain/superchain.go
geoknee: I think renaming would be good. Don't have strong opinions on separate structs, but just keen to leave all data structures in a neat state which won't cause confusion downstream.

bitwiseguy: Good call. Ended up using a separate struct for `Unmarshal` to reduce downstream confusion risk: https://github.com/ethereum-optimism/superchain-registry/pull/224/commits/c038f4c9529edb19c328c51ace862ec818d4ae3c

superchain/superchain.go
bitwiseguy: Good call. Ended up using a separate struct for Unmarshal to reduce downstream confusion risk: https://github.com/ethereum-optimism/superchain-registry/commit/c038f4c9529edb19c328c51ace862ec818d4ae3c

bitwiseguy:
@trianglesphere should we update the [check-security-configs script](https://github.com/ethereum-optimism/superchain-registry/blob/main/scripts/CheckSecurityConfigs.s.sol#L16) to perform any checks on the `DAChallenge` contract for plasma chains?

trianglesphere:
> @trianglesphere should we update the [check-security-configs script](https://github.com/ethereum-optimism/superchain-registry/blob/main/scripts/CheckSecurityConfigs.s.sol#L16) to perform any checks on the DAChallenge contract for plasma chains?

We could potentially require minimum DA Challenge / Resolve windows or validate the windows in the SC against what's expect in the config. I don't think we need to block this PR on those check though.

trianglesphere (approved):

mds1 (approved):

PR #227, opened 2024-05-09
superchain: add dispute game factory

tynes:
**Description**

Adds the dispute game factory as part of the superchain
ops tooling. It was added to a yaml file but it cannot
be accessed in go because it was not added to the various
structs.

Note that the abstractions are really bad right now (i wrote a bunch of
them) because I had to make a ton of changes to add a single contract.
Ideally we only have to make a couple of changes. I think i made
all changes required to add the dispute game factory.

Useful for https://github.com/ethereum-optimism/optimism/pull/10466


Inphi (approved):

geoknee:
We are already tracking this contract now.

PR #229, opened 2024-05-12
Add gas limit validation check

geoknee:
Towards https://github.com/ethereum-optimism/client-pod/issues/841


validation/standard/standard-config-mainnet.toml
```diff
@@ -21,3 +21,6 @@ scalar = [684_000, 684_000]
 decimals = [6, 6]
 blob_base_fee_scalar = [0, 10_000_000]
 base_fee_scalar = [0, 10_000_000]
+
+[system_config]
+gas_limit = [1000, 60_000_000]
```
trianglesphere, line 26: Min L2 gas limit should be more like 8 million. It needs to be greater than the deposit gas limit (set via MAX_RESOURCE_LIMIT in the portal contract + the amount of gas allocated to the L1 info deposit)

mds1: For reference, on OP Mainnet the minimum gas limit is 21M. We probably should use this value as the minimum until we have a good reason to lower it further.

```
$ cast call 0x229047fed2591dbec1eF1118d64F7aF3dB9EB290 "minimumGasLimit()(uint)"
21000000 [2.1e7]
```

geoknee: [21c5062](https://github.com/ethereum-optimism/superchain-registry/pull/229/commits/21c50620edf85ad520cd46533260532b92ae1567)

validation/utils_test.go
```diff
@@ -23,13 +24,27 @@ var isBigIntWithinBounds = func(actual *big.Int, bounds [2]*big.Int) bool {
...
-var isWithinBounds = func(actual uint32, bounds [2]uint32) bool {
+func isWithinBounds[T uint32 | uint64](actual T, bounds [2]T) bool {
 	if bounds[1] < bounds[0] {
 		panic("bounds are in wrong order")
 	}
 	return (actual >= bounds[0] && actual <= bounds[1])
 }
 
+// assertInBounds fails the test (but not immediately) if the passed param is outside of the passed bounds.
+var assertInBounds = func(t *testing.T, name string, got *big.Int, want [2]*big.Int) {
```
bitwiseguy, line 35: Consider changing this name `assertInBounds` --> `assertBigIntInBounds` to keep consistent format with `assertIntInBounds`

bitwiseguy (approved):

validation/standard/config.go
```diff
@@ -42,8 +46,9 @@ type EcotoneGasPriceOracleBounds struct {
 	BaseFeeScalar     Uint32Bounds `toml:"base_fee_scalar"`
 }
 
-type StandardConfigTy struct {
+type ConfigTy struct {
```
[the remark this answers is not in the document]
geoknee: It stands for "type", I made this more explicit in [3993b4f](https://github.com/ethereum-optimism/superchain-registry/pull/229/commits/3993b4f962d26ef24dcdfa8637cfcdf012a5005b). Because I want to export a singleton variable caleld `Config`, I had to call this something else. Otherwise no special meaning.

PR #232, opened 2024-05-14
feat: exclude `sealchain1` & `sealchain2` from semver and the bounds …

Ethnical:
The seal `sealchain1` & `sealchain2` are two chains pre-fault proofs thus the version of `<3.8.0`. 
Also, the [Challenge Period](https://specs.optimism.io/protocol/withdrawals.html#withdrawal-flow) exclusion is cause because the upperbounds of `12 seconds` so 1 block is clearly *not enough* for the current use cases. 
This is worth noting that theses chains are temporary and will be removed in the future days as we mentionned with @geoknee.

### DIFF
--- a/validation/l2oo_test.go
+++ b/validation/l2oo_test.go
@@ -31,6 +31,8 @@ func TestL2OOParams(t *testing.T) {
 		1740:      true, // sepolia/metal Incorrect finalizationPeriodSeconds, 604800 is not within bounds [12 12]
 		919:       true, // sepolia/mode  Incorrect finalizationPeriodSeconds, 180 is not within bounds [12 12]
 		11155420:  true, // sepolia/op No L2OO because this chain uses Fault Proofs https://github.com/ethereum-optimism/superchain-registry/issues/219
+		6619:      true, // sepolia/sealchain1 1800 is not within bounds [12 12].
+		18714:     true, // sepolia/sealchain2 1800 is not within bounds [12 12].
 	}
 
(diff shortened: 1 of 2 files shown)


geoknee (approved):

ipatka:
> The seal `sealchain1` & `sealchain2` are two chains pre-fault proofs thus the version of `<3.8.0`. Also, the [Challenge Period](https://specs.optimism.io/protocol/withdrawals.html#withdrawal-flow) exclusion is cause because the upperbounds of `12 seconds` so 1 block is clearly _not enough_ for the current use cases. This is worth noting that theses chains are temporary and will be removed in the future days as we mentionned with @geoknee.



> The seal `sealchain1` & `sealchain2` are two chains pre-fault proofs thus the version of `<3.8.0`. Also, the [Challenge Period](https://specs.optimism.io/protocol/withdrawals.html#withdrawal-flow) exclusion is cause because the upperbounds of `12 seconds` so 1 block is clearly _not enough_ for the current use cases. This is worth noting that theses chains are temporary and will be removed in the future days as we mentionned with @geoknee.



> The seal `sealchain1` & `sealchain2` are two chains pre-fault proofs thus the version of `<3.8.0`. Also, the [Challenge Period](https://specs.optimism.io/protocol/withdrawals.html#withdrawal-flow) exclusion is cause because the upperbounds of `12 seconds` so 1 block is clearly _not enough_ for the current use cases. This is worth noting that theses chains are temporary and will be removed in the future days as we mentionned with @geoknee.

I pulled these changes into the SEAL PR https://github.com/ethereum-optimism/superchain-registry/pull/225 as well

Issue #178, opened 2024-04-12
fix: avoid skipping checks for conduit chains


mds1:
From https://github.com/ethereum-optimism/superchain-registry/pull/107#pullrequestreview-1998488407

> Approving and merging this PR, but we'll later need to investigate the Genesis test failures that are currently being skipped for Metal Sepolia, Metal Mainnet, and Superlumio Mainnet
> 
> We also are skipping the L2OO tests for those chains + Mode Sepolia, and should confirm that's ok.

geoknee:
https://www.notion.so/oplabs/Superchain-Genesis-Hash-checks-60e173e9fb0c43e88a81d0666cb3d582

geoknee:
The key thing is if ecotone is activated at or before genesis. this affects genesis block because of these lines https://github.com/ethereum-optimism/op-geth/blob/9617f2a7dd0ab80c468d6e474e3cc26be9572c97/core/genesis.go#L538-L552

geoknee:
Here's one way of fixing this https://github.com/ethereum-optimism/superchain-registry/compare/gk/skip-default-hf-times?expand=1. 

The idea is to not specify hardfork times in a superchain wide way any more. Every chain has to specify their own hard fork times.

Pros:
* remove a lot of complexity in superchain registry Go code
* genesis hash checks pass again
* no downstream changes necessary, op-geth and op-node inspect the per chain values anyway

Cons:
* we may need to go back and get the correct values for chains, I don't think we can tell if we have them correct at the moment
* more repetition across configs (not a big deal, I think)
* no longer enshrining "upgrade together" flow into the data structures. But, we can probably still do that using other tooling, and write whatever checks and balances we want. 


Here is the PR where the behaviour was added https://github.com/ethereum-optimism/superchain-registry/pull/101

PR #233, opened 2024-05-14
Introduce `superchain_time` field into each chain's config

geoknee:
Closes https://github.com/ethereum-optimism/superchain-registry/issues/178

This field is used to declare **when** a chain has committed to activating hardforks with the rest of the superchain. 

In practice this means that the individual chain config will have default values from `superchain.yaml` injected, for any such value which is later than `superchain_time`.  This allows chains to activate hardforks at arbitrary times (or not at all) up until `superchain_time`, but forces them to use superchain-wide activation times following that. 

For now only the OP chains have this field set. I think that when a chain becomes a standard chain, it should have `superchain_time` set, forcing it to have hardforks in lockstep from that point on. I enshrined that idea in the contributing guide on this PR, happy to discuss alternatives.


superchain/configs/mainnet/superlumio.yaml
```diff
@@ -16,3 +16,5 @@ genesis:
     hash: "0x01711dc53a82967a34f4b38f7dff3ae8fa760882bb1e2ace50bb01e9ff420a9b"
     number: 0
   l2_time: 1708984631
+
+ecotone_time: 1708984632
```
geoknee, line 20: This is almost certainly incorrect -- I have reached out to try and get the correct value. It "works" in the sense that the genesis block hash validation check passes, and the GasPriceOracle validation check knows to use the "ecotone" path.

geoknee: Fixed with updated times

superchain/superchain.go
```diff
@@ -132,16 +135,18 @@ func (c *ChainConfig) SetDefaultHardforkTimestampsToNil(s *SuperchainConfig) {
...
 	sVal := reflect.ValueOf(&s.hardForkDefaults).Elem()
 
 	for i := 0; i < reflect.Indirect(cVal).NumField(); i++ {
 		overrideValue := cVal.Field(i)
-		if overrideValue.IsNil() {
-			defaultValue := sVal.Field(i)
-			overrideValue.Set(defaultValue)
+		defaultValue := sVal.Field(i)
+		if overrideValue.IsNil() && !defaultValue.IsNil() {
+			if c.SuperchainTime != nil && reflect.Indirect(defaultValue).Uint() >= *c.SuperchainTime {
```
sebastianst, line 147: You can merge these two. And the inverse `c.SuperchainTime != nil` check you can actually move to the top of `setNilHardforkTimestampsToDefault` as an early exit if it's `nil`.

sebastianst (commented):
Do we have a CI step that shows which chain configs effectively changed by this PR? If not, this would be a great check to add, similar to our k8s diff-bot. Because IIUC we want all chain configs to stay effectively the same, besides `metal-mainnet`, right?

superchain/configs/mainnet/superlumio.yaml
```diff
@@ -16,3 +16,5 @@ genesis:
     hash: "0x01711dc53a82967a34f4b38f7dff3ae8fa760882bb1e2ace50bb01e9ff420a9b"
     number: 0
   l2_time: 1708984631
+
+ecotone_time: 1710374401
```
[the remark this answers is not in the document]
geoknee: Good call out. If you check the full genesis.json file [here](https://api.conduit.xyz/file/v1/optimism/genesis/pontem-mainnet-0), you'll see that actually canyone was not activated at genesis on this chain (no shanghai time), and our genesis hash validation check confirms that. I can double check with them, our tooling was not scraping the hardfork times properly when the chain was added, so we need to manually put the correct times in now.

superchain/configs/mainnet/base.yaml
```diff
@@ -17,3 +17,5 @@ genesis:
     hash: "0xf712aa9241cc24369b143cf6dce85f0902a9731e70d66818a3a5845b296c73dd"
     number: 0
   l2_time: 1686789347 # Thu 15 Jun 2023 00:35:47 UTC
+
+superchain_time: 0 # hardfork times inherited from superchain.yaml
```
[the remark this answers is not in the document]
geoknee: [b8ec2b2](https://github.com/ethereum-optimism/superchain-registry/pull/233/commits/b8ec2b2d0c20220eb3796e7f76d245215d8cec7e)

superchain/configs/sepolia-dev-0/oplabs-devnet-0.yaml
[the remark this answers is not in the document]
geoknee: [1f0ecec](https://github.com/ethereum-optimism/superchain-registry/pull/233/commits/1f0ecec019330bc55bf4204a50920cb9b4af48a7)

CONTRIBUTING.md
```diff
@@ -44,7 +44,7 @@ sh scripts/add-chain.sh frontier
 The tool will write the following data:
 - The main configuration source, with genesis data, and address of onchain system configuration. These are written to `superchain/configs/superchain_target/chain_short_name.yaml`.
 > **Note**
-> Hardfork override times will be included. For standard chains, you must have all hardforks activated (see the neighbouring superchain.yaml file). It is not possible to override a superchain-wide hardfork time with `nil`.
+> Hardfork override times, where they have been set, will be included. If and when a chain becomes a standard chain, future hardfork activation times will be inherited from superchain-wide values in the neighboring superchain.yaml file.
```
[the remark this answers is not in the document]
geoknee: [5311b66](https://github.com/ethereum-optimism/superchain-registry/pull/233/commits/5311b663bff4d78d0139ef659315bb5602d69ee1)

superchain/superchain.go
geoknee: [45f44ae](https://github.com/ethereum-optimism/superchain-registry/pull/233/commits/45f44ae5e44a2627d68d1b9ee11a798abaedbc83)

superchain/configs/sepolia/metal.yaml
```diff
@@ -17,4 +17,5 @@ genesis:
     number: 0
   l2_time: 1708129620
 
-delta_time: 1708385400   # Mon Feb 19 2024 23:30:00
\ No newline at end of file
+delta_time: 1708385400   # Mon Feb 19 2024 23:30:00
+ecotone_time: 1708534800
```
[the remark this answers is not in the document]
geoknee: [a75b6ba](https://github.com/ethereum-optimism/superchain-registry/pull/233/commits/a75b6bae4a55f9e3dead2cb79ef56b5757443900)

geoknee:
> Do we have a CI step that shows which chain configs effectively changed by this PR? If not, this would be a great check to add, similar to our k8s diff-bot. Because IIUC we want all chain configs to stay effectively the same, besides `metal-mainnet`, right?

Yes we want most chain configs to stay unchanged, except those that were wrong! We know that metal sepolia and superlumio were wrong, because of the genesis hash check. 

I can look into the automated check, so at least we know what is changing. Might be something to do manually for now.

geoknee:
I implemented the diff tool. For now it just fails the test when there is a diff. For this particular PR, we would want to merge it as long as the diff is the one we want.

superchain/configs/mainnet/superlumio.yaml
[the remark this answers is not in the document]
geoknee: I think it doesn't make a difference to the genesis hash check, which is our immediate blocker, since any time after genesis is equivalent with respect to that hash. But probably we don't want to omit those upgrades altogether, might be safest to set them to the same as the `ecotone_time` until we can get the exact time from the conduit folks.

sebastianst (approved):
This looks great, thanks! Love the new config diff ci step.

One minor improvement, if we swap the order in the diff, then the `+` lines will be the new changes. I believe it is the other way around right now.

I assume this is the superlumio diff we expect here (besides the +/-)?
```diff
--- ./generate-rollup-config/output-base/mainnet-SuperLumio.json	2024-05-15 15:58:44.948968529 +0000
+++ ./generate-rollup-config/output-HEAD/mainnet-SuperLumio.json	2024-05-15 15:58:40.412998242 +0000
@@ -23,8 +23,8 @@
   "l1_chain_id": 1,
   "l2_chain_id": 8866,
   "regolith_time": 0,
-  "canyon_time": 1704992401,
-  "delta_time": 1708560000,
+  "canyon_time": 1710374401,
+  "delta_time": 1710374401,
   "ecotone_time": 1710374401,
```

sbvegan (approved):

.github/workflows/rollup-config-diff.yml
```diff
@@ -0,0 +1,39 @@
+name: Rollup Config Diff
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
```
mds1, line 5: Looks like this is the default set so this is ok, but what is synchronize? I don't see an explanation of it in the [github docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request)

geoknee: I believe it is when a new commit is pushed to a PR.

.github/workflows/rollup-config-diff.yml
```diff
@@ -0,0 +1,39 @@
...
+
+jobs:
+  compute-rollup-config-diff:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
```
mds1, line 13: v4 is currently the latest
```suggestion
      uses: actions/checkout@v4
```

CONTRIBUTING.md
```diff
@@ -44,7 +44,7 @@ sh scripts/add-chain.sh frontier
 The tool will write the following data:
 - The main configuration source, with genesis data, and address of onchain system configuration. These are written to `superchain/configs/superchain_target/chain_short_name.yaml`.
 > **Note**
-> Hardfork override times will be included. For standard chains, you must have all hardforks activated (see the neighbouring superchain.yaml file). It is not possible to override a superchain-wide hardfork time with `nil`.
+> Hardfork override times, where they have been set, will be included. If and when a chain becomes a standard chain, a superchain_time is set in the chain config. From that time on, future hardfork activation times which are missing from the chain config will be inherited from superchain-wide values in the neighboring superchain.yaml file.
```
mds1, line 47: nits
```suggestion
> Hardfork override times, where they have been set, will be included. If and when a chain becomes a standard chain, a `superchain_time` is set in the chain config. From that time on, future hardfork activation times which are missing from the chain config will be inherited from superchain-wide values in the neighboring `superchain.yaml` file.
```

geoknee:
> This looks great, thanks! Love the new config diff ci step.
> 
> One minor improvement, if we swap the order in the diff, then the `+` lines will be the new changes. I believe it is the other way around right now.
> 
> I assume this is the superlumio diff we expect here (besides the +/-)?
> 
> ```diff
> --- ./generate-rollup-config/output-base/mainnet-SuperLumio.json	2024-05-15 15:58:44.948968529 +0000
> +++ ./generate-rollup-config/output-HEAD/mainnet-SuperLumio.json	2024-05-15 15:58:40.412998242 +0000
> @@ -23,8 +23,8 @@
>    "l1_chain_id": 1,
>    "l2_chain_id": 8866,
>    "regolith_time": 0,
> -  "canyon_time": 1704992401,
> -  "delta_time": 1708560000,
> +  "canyon_time": 1710374401,
> +  "delta_time": 1710374401,
>    "ecotone_time": 1710374401,
> ```

This is the diff we expect **including the current +/- parity**. The canyon and delta times were changed to later times. 

So I don't think we need to swap any orders?

geoknee:
> Two nits but otherwise LGTM, CI is also failing with `Exiting with status 1 because at least one difference was found`. Approving since I'm ok with merging once those are resolved

Thanks. This is actually an expected failure, since for this PR we want to change the effective rollup config for these chains (it is incorrect as is, causing genesis hash checks to fail). So when we're ready I'll merge it despite the failure (it won't run on the `main` branch). Longer term, we can improve how that action works to be more of an informative check like the argoCD diffbot. For now this check is mainly useful to get confidence on this PR, it may also transform into a check which runs when a new chain is added, to ensure the new data results in the correct config.

superchain/configs/sepolia/metal.yaml
```diff
@@ -17,4 +18,5 @@ genesis:
     number: 0
   l2_time: 1708129620
 
-delta_time: 1708385400   # Mon Feb 19 2024 23:30:00
\ No newline at end of file
+delta_time: 1708385400 # Tue 16 Jan 2024 15:30:00 UTC
+ecotone_time: 1708534800 # Thu 18 Jan 2024 03:00:00 UTC
```
sebastianst, line 22: Something is wrong here. I get
```
date -u -r 1708385400
Mon Feb 19 23:30:00 UTC 2024

date -u -r 1708534800
Wed Feb 21 17:00:00 UTC 2024
```

sebastianst:
> 
> This is the diff we expect **including the current +/- parity**. The canyon and delta times were changed to later times.
> 
> So I don't think we need to swap any orders?

@geoknee I see, I assumed that removal of canyon in sepolia-metal was a hint that the order is wrong. The latest sepolia metal diff is
```diff
Files ./generate-rollup-config/output-base/sepolia-Metal Sepolia.json and ./generate-rollup-config/output-HEAD/sepolia-Metal Sepolia.json are different:
--- "./generate-rollup-config/output-base/sepolia-Metal Sepolia.json"	2024-05-15 17:28:43.848672025 +0000
+++ "./generate-rollup-config/output-HEAD/sepolia-Metal Sepolia.json"	2024-05-15 17:28:39.268612165 +0000
@@ -23,7 +23,6 @@
   "l1_chain_id": 11155111,
   "l2_chain_id": 1740,
   "regolith_time": 0,
-  "canyon_time": 1699981200,
   "delta_time": 1708385400,
   "ecotone_time": 1708534800,
```
 I see that we set
```yaml
superchain_time: null

delta_time: 1708385400
ecotone_time: 1708534800
```
for sepolia metal. So when does Canyon activate now?

superchain/configs/sepolia/metal.yaml
geoknee: fixed here [985db7b](https://github.com/ethereum-optimism/superchain-registry/pull/233/commits/985db7bec8e2f944659500928e1c7f34b148229e)

geoknee:
> for sepolia metal. So when does Canyon activate now?
I set this to block number 1 (2 secs after genesis), pending further clarification on the actual value.

superchain/configs/mainnet/superlumio.yaml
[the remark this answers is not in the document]
sebastianst: Note that in general, we don't want to do or allow activation of forks at the same block, if they don't activate at genesis, but at least be spaced one block apart.

PR #235, opened 2024-05-15
feat: Update OP Sepolia addresses

clabby:
## Overview

Updates the addresses on OP Sepolia in accordance w/ the upgrade in https://github.com/ethereum-optimism/superchain-ops/pull/200


superchain/extra/addresses/sepolia/op.json
```diff
@@ -10,8 +10,12 @@
   "ProxyAdminOwner": "0x1Eb2fFc903729a0F03966B917003800b145F56E2",
   "SystemConfigOwner": "0xfd1D2e729aE8eEe2E146c033bf4400fE75284301",
   "DisputeGameFactoryProxy": "0x05F9613aDB30026FFd634f38e5C4dFd30a197Fa1",
-  "AnchorStateRegistryProxy": "0xbCC6AB30a6A8aFb030112182c45D3cb64EF11952",
-  "DelayedWETHProxy": "0xCa36B00f885D83068f0CfFEe381F0c974de1d58e",
+  "AnchorStateRegistryProxy": "0x218CD9489199F321E1177b56385d333c5B598629",
+  "DelayedWETHProxy": "0xF3D833949133e4E4D3551343494b34079598EA5a",
```
mds1, line 14: Interesting, why did the proxy addresses change?

superchain/implementations/networks/sepolia.yaml
```diff
@@ -13,7 +13,10 @@ optimism_portal:
   2.5.1: 0xFF6Ce11Cd5094569E4Fc67C5b508F22754dA0667
   3.3.0: 0x215b3A2d1766616d1C4701f3D117348711BaAe93
   3.8.0: 0x62749b56f045D30E1971FFA9880D496c701Eb386
+  3.10.0: 0x35028bAe87D71cbC192d545d38F960BA30B4B233
 dispute_game_factory:
   0.2.0: 0x1B3cfC1F692076538c9B53750EBc412042D48605
+  1.0.0: 0xA51bea7E4d34206c0bCB04a776292F2f19F0BeEc
```
mds1, line 19: Looks like implementation addresses for `DelayedWETHProxy` and `AnchorStateRegistryProxy` are missing from this file

superchain/extra/addresses/sepolia/op.json
ajsutton: These are a bit weird to have a proxy for - they're actually behind the `FaultDisputeGame` implementation already and aren't accessed directly. So when we update `FaultDisputeGame` it points to a new proxy because we generally don't want to change the behaviour for existing games (which continue to use the old `FaultDisputeGame` for their lifetime).  These proxies really only exist to give us maximal flexibility if we need to intervene manually.

mds1: I see, thanks

sebastianst (approved):

mds1 (approved):

sbvegan (approved):

PR #238, opened 2024-05-16
remove devrel from CODEOWNERS

geoknee:
they will be reinstated when we get to "early access launch".

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -9,13 +9,10 @@
 /.golangci.yml @ethereum-optimism/go-reviewers
 /validation @ethereum-optimism/go-reviewers
 /superchain @ethereum-optimism/go-reviewers
+/add-chain @ethereum-optimism/go-reviewers
 
 ## Solidity
 *.sol @ethereum-optimism/security-reviewers
 
-# Superchain configuration ownership
-/superchain/configs @sbvegan @ZakAyesh @OPMattie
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -9,13 +9,10 @@
...
 /superchain @ethereum-optimism/go-reviewers
+/add-chain @ethereum-optimism/go-reviewers
 
 ## Solidity
 *.sol @ethereum-optimism/security-reviewers
 
-# Superchain configuration ownership
-/superchain/configs @sbvegan @ZakAyesh @OPMattie
-/superchain/extra @sbvegan @ZakAyesh @OPMattie
-/superchain/implementations @ethereum-optimism/security-reviewers
```
clabby, line 19: This should probably stick around

clabby (approved):

.github/CODEOWNERS
geoknee: OK -- what's the reason? Security are currently the default owners (see top of file). We will probably change the default to node squad. Do you think this should remain as security?

clabby: Not super opinionated, but security does tend to be a bit closer to the contracts.

Let's get this in for now, and we can follow up on who the right owners for the subsets of the registry are.

sebastianst (approved):

PR #242, opened 2024-05-16
feat: Add FP contracts to Go bindings

clabby:
## Overview

Adds the Fault Proof contracts to the Go bindings for the `superchain-registry`.


superchain/superchain.go
```diff
@@ -351,6 +386,34 @@ func (c ContractImplementations) Resolve(versions ContractVersions) (Implementat
 	if implementations.SystemConfig, err = resolve(c.SystemConfig, versions.SystemConfig); err != nil {
 		return implementations, fmt.Errorf("SystemConfig: %w", err)
 	}
+	// If the L2OutputOracle is not specified, we can assume that the L2OutputOracle is not used, and fault proofs are
+	// activated.
+	if implementations.L2OutputOracle, err = resolve(c.L2OutputOracle, versions.L2OutputOracle); err != nil {
```
sebastianst, line 391: Wondering if we want to introduce a sentinel error for `resolve` and check it here specifically. Looks like `resolve` can have different reasons for error.

superchain/superchain.go
```diff
@@ -418,6 +489,20 @@ func (c ContractVersions) VersionFor(contractName string) (string, error) {
...
+	case "AnchorStateRegistry":
+		version = c.AnchorStateRegistry
+	case "FaultDisputeGame":
+		version = c.FaultDisputeGame
+	case "PermissionedDisputeGame":
+		version = c.PermissionedDisputeGame
+	case "MIPS":
+		version = c.MIPS
+	case "PreimageOracle":
+		version = c.PreimageOracle
```
sebastianst, line 505: nit, everywhere else you keep the same order, and MIPS comes last after PreimageOracle, which is nice to read and makes things easier to verify. Here the last two are swapped.

sebastianst (approved):

geoknee (approved):

sbvegan (approved):

tynes:
nothing wrong with this PR but more of a architecture nit, this diff is way too large for what its adding, makes it very error prone to make diffs, a very similar type lives in the monorepo [here](https://github.com/ethereum-optimism/optimism/blob/8377204bd56bbb0aaf28ce0d2ee670c865e1d535/op-chain-ops/genesis/config.go#L690), would be nice to dedupe

ajsutton:
> nothing wrong with this PR but more of a architecture nit, this diff is way too large for what its adding, makes it very error prone to make diffs, a very similar type lives in the monorepo [here](https://github.com/ethereum-optimism/optimism/blob/8377204bd56bbb0aaf28ce0d2ee670c865e1d535/op-chain-ops/genesis/config.go#L690), would be nice to dedupe

I think `op-chain-ops` ultimately should migrate to just using these bindings from the super chain registry - can't do it the other way or we'd create a dependency cycle.  I'm not sure we want to expand this work to start touching the genesis generation code though - feels like scope creep and not something we have much context on.

geoknee:
We could potentially reduce a lot of boilerplate in this part of the codebase by using some reflection. We're using that in a few other choice places and is (I think) the right solution when we want to iterate over all members of a struct, would reduce the diffs on PRs like this going forward. If there is support for this idea, I can mix it in to Node squad's work on the Superchain Registry in coming weeks.

ajsutton:
@geoknee Reflection definitely seems useful for the places where we're mapping string names to struct vars.  If possible we should avoid the need for those lookups and just use the struct directly, but I'm guessing we can't manage that.

superchain/implementations/networks/sepolia.yaml
```diff
@@ -4,22 +4,28 @@ l1_erc721_bridge:
   2.1.0: 0xAE2AF01232a6c4a4d3012C5eC5b1b35059caF10d
 l1_standard_bridge:
   2.1.0: 0x64B5a5Ed26DCb17370Ff4d33a8D503f0fbD06CfF
-l2_output_oracle:
-  1.8.0: 0xF243BEd163251380e78068d317ae10f26042B292
```
mbaxter, line 8: We still have a list of "global" L2OO implementations defined in `implementations.yaml` [here](https://github.com/ethereum-optimism/superchain-registry/blob/5b11cbf3f63d9b66259d1120126403fa36cd7ca6/superchain/implementations/implementations.yaml#L13-L16) that is [merged into the network config here](https://github.com/ethereum-optimism/superchain-registry/blob/f68cf836cb872913b5a175c999dd4b566fd1b886/superchain/superchain.go#L582).  Given we still have L2OO impls defined, just wanted to check that it makes sense to remove the implementation here?

superchain/superchain.go
```diff
@@ -351,30 +388,66 @@ func (c ContractImplementations) Resolve(versions ContractVersions) (Implementat
...
+		}
+	} else if err != nil {
+		return implementations, fmt.Errorf("L2OutputOracle: %w", err)
+	}
 	return implementations, nil
 }
 
 // resolve returns a VersionedContract that matches the passed in semver version
 // given a set of addresses.
 func resolve(set AddressSet, version string) (VersionedContract, error) {
```
mbaxter, line 426: Updated the way `resolve` functions here.  We now just perform exact matching to resolve the `version` against the list of implementations (`set`).

Previously, an empty `version` would be compared to each version in `set`, and the underlying `semver` package would [silently sort this invalid version as "less than"](https://cs.opensource.google/go/x/mod/+/master:semver/semver.go;l=126) the valid versions.  This caused an empty version to be fulfilled with the latest implementation.

superchain/superchain.go
mbaxter: Updated to check explicitly for an `ErrEmptyVersion` error

superchain/implementations/networks/sepolia.yaml
ajsutton: I lean towards removing this - nothing should use it anymore so we shouldn't need to keep a record of the custom location this version wound up on for sepolia.

ajsutton (approved):
LGTM. Exact version matches make a lot more sense.
As a sanity check, I hacked the code below into the monorepo after pointing it at super chain-registry with these changes:
```
	chain := chaincfg.ChainByName("op-sepolia")
	proxy := superchain.Addresses[chain.ChainID].DisputeGameFactoryProxy
	fmt.Printf("Proxy Address: %v\n", proxy)
```

and it printed the right address.  So the loading of fault proof contract addresses is working correctly.

PR #246, opened 2024-05-17
swapping the content of the contibuting doc and the readme

sbvegan:
**Description**

A clear and concise description of the features you're adding in this pull request.

**Tests**

Please describe any tests you've added. If you've added no tests, or left important behavior untested, please explain why not.

**Additional context**

Add any other context about the problem you're solving.

**Metadata**

- Fixes #[Link to Issue]

PR #248, opened 2024-05-20
improvements to the readme / contributing refactor

geoknee:
See #246 

* Moves the "add superchain target" material back to contributing (this is a pretty rare operation and most likely done by us)
* adds a TOC to the contributing page
* a few other small typo and broken link fixes

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,24 +1,98 @@
 # Superchain Registry Contributing Guide
+The Superchain Registry repository contains:
+* raw ["per-chain" config data](./README.md#3-understand-output) in `yaml` and `json/json.gz` files arranged in a semantically meaningful directory structure
+* [superchain-wide config data](#superchain-wide-config-data)
+* a Go workspace with
+  - a [`superchain`](#superchain-go-module) module
+  - a [`validation`](#validation-go-module) module
+  - an [`add-chain`](#add-chain-go-module) module
+  - The modules are tracked by a top level `go.work` file. The associated `go.work.sum` file is gitignored and not important to typical workflows, which should mirror those of the [CI configuration](.circleci/config.yml).
+* a Forge/Solidity script [`CheckSecurityConfigs`](#checksecurityconfigs)
+* Automatically generated summary `chainIds.json` file
 
-See [Superchain Upgrades] OP-Stack specifications.
 
-[Superchain Upgrades]: https://specs.optimism.io/protocol/superchain-upgrades.html
+## Superchain-wide config data
+A superchain target defines a set of layer 2 chains which share a `SuperchainConfig` and `ProtocolVersions` contract deployment on layer 1. It is usually named after the layer 1 chain, possibly with an extra identifier to distinguish devnets.
 
-## Superchain Go Module
 
-Superchain configs can be imported as Go-module:
+> **Note**
+> Example: `sepolia` and `sepolia-dev-0` are distinct superchain targets, although they are on the same layer 1 chain.
+
+### Adding a superchain target
+A new Superchain Target can be added by creating a new superchain config directory,
+with a `superchain.yaml` config file.
+
+> **Note**
+> This is an infrequent operation and unecessary if you are just looking to add a chain to an existing superchain.
+
+Here's an example:
+
+```bash
+cd superchain-registry
+
+export SUPERCHAIN_TARGET=goerli-dev-0
+mkdir superchain/configs/$SUPERCHAIN_TARGET
+
+cat > superchain/configs/$SUPERCHAIN_TARGET/superchain.yaml << EOF
+name: Goerli Dev 0
+l1:
+  chain_id: 5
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -1,24 +1,98 @@
...
+protocol_versions: 1.0.0
+superchain_config: 1.1.0
+```
+See the `semver.yaml` files in existing superchain targets for the latest set of contracts to specify.
+
+### `implementations`
+Per superchain a set of canonical implementation deployments, per semver version, is tracked.
+As default, an empty collection of deployments can be set:
+```bash
+cat > superchain/implementations/networks/$SUPERCHAIN_TARGET.yaml << EOF
```
mds1, line 70: I still don't (and never fully did) understand the difference between this and `semver.yaml`.  In this [notion doc](https://www.notion.so/oplabs/Smart-Contract-Release-Planning-a51c1ab2413949f3a0513e42b80c0061?pvs=4) we recently defined a mapping from `op-contracts/vX.Y.Z` release to the expected semvers for each contract. Ultimately, the verification we should be doing in the superchain registry is making sure a chain's contracts is on one of those `op-contracts` tags, as determined by the semvers (and bytecode validation, since a contract can lie about its semver).

So what exactly is supposed to be in each of these files that lists contract versions, in relation to the linked version definitions?

CONTRIBUTING.md
```diff
@@ -1,24 +1,98 @@
...
-## Superchain Go Module
+Per chain and supechain-wide configs and extra data are embedded into the `superchain` go module, which can be imported like so:
 
-Superchain configs can be imported as Go-module:
 ```
 go get github.com/ethereum-optimism/superchain-registry/superchain@latest
 ```
-See [`op-chain-ops`] for config tooling and
- for smart-contract bindings.
+The configs are consumed by downstream OPStack software, i.e. `op-geth` and `op-node`.
```
mds1, line 88: nit, in the docs is stylized like this
```suggestion
The configs are consumed by downstream OP Stack software, i.e. `op-geth` and `op-node`.
```

CONTRIBUTING.md
```diff
@@ -106,4 +180,27 @@ graph TD
...
+    "go.formatTool": "gofumpt",
+    "go.lintTool": "golangci-lint",
+    "go.lintOnSave": "workspace",
+    "gopls": {
+        "formatting.gofumpt": true,
+    },
+}
+```
+The `semver.yaml` files each represent the semantic versioning lockfile for the all of the smart contracts in that superchain.
+It is meant to be used when building transactions that upgrade the implementations set in the proxies.
```
mds1, line 199: Wrong section?

CONTRIBUTING.md
```diff
@@ -106,4 +180,27 @@ graph TD
...
+
+
+
+## Links
+See [Superchain Upgrades] OP-Stack specifications.
```
mds1, line 204: same nit, ref https://docs.optimism.io/stack/getting-started
```suggestion
See [Superchain Upgrades] OP Stack specifications.
```

CONTRIBUTING.md
geoknee: Good catch, [9e2725d](https://github.com/ethereum-optimism/superchain-registry/pull/248/commits/9e2725d49aee8ebdc1e2096b1815ed0845211168)

CONTRIBUTING.md
geoknee: Thanks for raising this, I have also never really understood the role of this file. I think resolving this is very important, but somewhat outside the scope of this PR so if possible would prefer not to block on it.

The document you linked is very clear and I would be happy to overhaul the validation checks and declaration files have in the registry to match it. 

Proposal:
* `semver.yaml` -> `standard-contract-versions.toml` (or some better name), with now encodes the mapping from monorepo tags to a list of (contract,version) pairs as described in your doc. Lives next to the standard config toml files in the `validation` module, we no longer export that information from `superchain`. 
* validation check ensures each chain has a set of contract versions which _exactly match_ 1:1 to at least one monorepo tag mapping entry in the above file. From what I can tell this can be just a dumb string equality check, since the versions are not true semantic versions and we won't do any true semantic version comparisons.
* bytecode validation is something which is not yet in scope for the superchain registry checks but we are looking at pulling that in to scope.


The "implementations" yaml files are currently not referenced by any validation check in this repo. They are referenced in one place in  `op-chain-ops` https://github.com/ethereum-optimism/optimism/blob/69bb36c9ba95705cc730d1e145501a510c448775/op-chain-ops/cmd/op-upgrade/main.go#L226-L246 . You might be better placed than me to suggest how we tidy that up.

mds1: Agreed on this not blocking the PR, so created https://github.com/ethereum-optimism/client-pod/issues/879 to track this. 

In general, I would like to better understand the original intent behind the current architecture to make sure this proposal is ok (I mentioned this in the linked issue, and we can continue the discussion there)

mds1 (approved):

Issue #36, opened 2023-12-04
feat: Support all OP Stack components


clabby:
**Is your feature request related to a problem? Please describe.**
Currently, the alternative implementations of OP Stack components, apart from `op-erigon` and `asterisc`, cannot take advantage of the superchain registry. This creates a sharded collection of chain configurations, where components like `op-reth`, `magi`, `hildr`, `op-besu`, `cannon-rs`, the Rust FPP, and whatever comes in the future that isn't written in Golang, cannot take advantage of this package.

**Describe the solution you'd like**
Modularize the superchain registry such that it can be used by all OP Stack components, preparing for a future where our critical infrastructure is implemented in many more languages than just `Go`.

Ideally, this means defining a core implementation, preferably in a language such as `C`, `C++`, or `Rust` that is more friendly for interop, and then creating FFI bindings that enable several front-end APIs to be built for the library.

**Describe alternatives you've considered**
1. Implement separate versions of the `superchain-registry` repository for each language that the OP Stack's core components are built upon. This would require a decent bit of duplicated code, increasing the possibility of divergent configuration between the implementations.
2. Create FFI bindings from the Go implementation. Go does not lend itself well towards being called via FFI, and `Cgo` is a bit horrendous to work with when the direction is `<language>` -> `Go` rather than `Go` -> `<language>`. It requires `cbindgen` to embed Go types and semantics into the C header, etc. Rather than doing this, we should choose a more FFI friendly language that is closer to `C` (or, even `C`/`C++` itself) for the base implementation.

**Additional context**
Modularizing the superchain registry will pay dividends in the future, offering a central, interoperable library for every OP Stack component to share the same configuration utilities.

ajsutton:
Before diving into building sharable code, we should review how we approach configuration.  It's usually better to have a well defined config file format so that each client can then parse it using its own code rather than needing to embed a shared library.  I'm not sure of the details of why we've got specific go code in this repo and it may be the right option  but we should evaluate the options for making config available to alternate clients and not just dive into building reusable code.

protolambda:
>  I'm not sure of the details of why we've got specific go code in this repo

To have have a form that we can read and validate the configuration consistency of. E.g. address-format (checksums), completeness (missing references), and bytecode hashing are all tested with Go. And then it's better to expose the struct types publicly, so we do not have to duplicate them in the optimism monorepo or op-geth software.

---

[...]

PR #259, opened 2024-06-04
feat: rust bindings

merklefruit:
**Description**

Adds minimal rust bindings to the superchain registry package.

1. Static files are embedded with the help of [`include_dir`](https://docs.rs/include_dir/latest/include_dir/index.html) in the crate's binary
2. They are loaded lazily as [`lazy_static`](https://docs.rs/lazy_static/latest/lazy_static/)s and exported for usage.

[Ctor](https://github.com/mmastrac/rust-ctor) could be used alternatively of lazy_static, however I feel like lazy evaluation is an important feature and consider it preferable (and less experimental) than global constructors.

**Tests**

- decode statically embedded files into the new superchain types
- more tests are still needed to make sure everything is decoded properly

**Metadata**

- Fixes part of #36 
- Fixes https://github.com/ethereum-optimism/kona/issues/193


bindings/rust/src/superchain.rs
```diff
@@ -0,0 +1,235 @@
...
+
+pub type Superchains = HashMap<String, Superchain>;
+pub type OPChains = HashMap<u64, ChainConfig>;
+pub type Addresses = HashMap<u64, AddressList>;
+pub type GenesisSystemConfigs = HashMap<u64, GenesisSystemConfig>;
+pub type Implementations = HashMap<String, ContractImplementations>;
+
+// TODO: should we use alloy for these?
+pub type Address = String;
+pub type Hash = String;
```
merklefruit, line 14: this is intentionally left as a question: any opinions?

bindings/rust/src/lib.rs
```diff
@@ -0,0 +1,57 @@
+#![doc = include_str!("../README.md")]
+#![warn(missing_debug_implementations, missing_docs, rustdoc::all)]
+#![deny(unused_must_use, rust_2018_idioms)]
+#![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
+
```
refcell, line 5: Can we make this crate `no_std` and switch the `std::collections::HashMap` data structure to use the `no_std` [hashbrown](https://docs.rs/hashbrown/latest/hashbrown/) crate like `kona` uses (`hashbrown = "0.14.3"`).

```suggestion
#![no_std]

```

bindings/rust/src/superchain.rs
```diff
@@ -0,0 +1,300 @@
...
+/// Map of chain IDs to their address lists.
+pub type Addresses = HashMap<u64, AddressList>;
+
+/// Map of chain IDs to their chain's genesis system configurations.
+pub type GenesisSystemConfigs = HashMap<u64, GenesisSystemConfig>;
+
+/// Map of superchain names to their implementation contract semvers.
+pub type Implementations = HashMap<String, ContractImplementations>;
+
+// TODO: should we use alloy for these?
```
refcell, line 21: Yes definitely 👍

bindings/rust/src/superchain.rs
```diff
@@ -0,0 +1,300 @@
...
+    /// L2 genesis block
+    pub l2: BlockID,
+    /// Timestamp of the L2 genesis block
+    pub l2_time: u64,
+    /// Extra data for the genesis block
+    pub extra_data: Option<Vec<u8>>,
+    /// Optional System configuration
+    #[serde(flatten)]
+    pub system_config: Option<SystemConfig>,
+}
```
refcell, line 161: Are we able to use alloy's genesis definition here?

https://github.com/alloy-rs/alloy/blob/main/crates/genesis/src/lib.rs#L27

refcell (commented):
Great work. I think if we're able to pull in alloy types here, we should be good to get this merged and continue iterating on what's in trunk.

bindings/rust/src/superchain.rs
merklefruit: I believe that struct holds info about the specific genesis block, whereas this struct is more general and describes an L2 chain genesis. Perhaps the opposite (aka defining ChainGenesis in op-alloy) would make sense?

bindings/rust/src/superchain.rs
```diff
@@ -0,0 +1,300 @@
...
+    /// standard OP Stack configuration and are considered "advanced".
+    Frontier = 1,
+    /// Standard chains don't have any customizations beyond the
+    /// standard OP Stack configuration and are considered "vanilla".
+    Standard = 2,
+}
+
+/// A chain configuration.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ChainConfig {
```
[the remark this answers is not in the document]
merklefruit: I believe alloy's ChainConfig is replicating this one in Geth: https://github.com/ethereum/go-ethereum/blob/64dccf7aa411c5c7cd36090c3d9b9892945ae813/params/config.go#L349

Seeing that you are re-defining it differently in superchain-registry's Go package here: https://github.com/ethereum-optimism/superchain-registry/blob/acf0eef75f26247d7e792ef3f661e54fa9dc73c4/superchain/superchain.go#L84

makes me think that they have been created separately voluntarily, otherwise you'd use that one in the Go package instead. I could be wrong tho!

bindings/rust/src/superchain.rs
merklefruit: But this type in the superchain-registry looks to match! 
https://github.com/ethereum-optimism/superchain-registry/blob/acf0eef75f26247d7e792ef3f661e54fa9dc73c4/superchain/superchain.go#L700

We can simply re-export it from this module then

bindings/rust/src/superchain.rs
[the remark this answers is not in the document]
refcell: Agreed thanks :)

bindings/rust/src/init.rs
```diff
@@ -0,0 +1,139 @@
...
+        //             entry_config.l1.public_rpc = ci_mainnet_rpc;
+        //         }
+        //     }
+        //     "sepolia" | "sepolia-dev-0" => {
+        //         if let Ok(ci_sepolia_rpc) = std::env::var("CIRCLE_CI_SEPOLIA_RPC") {
+        //             entry_config.l1.public_rpc = ci_sepolia_rpc;
+        //         }
+        //     }
+        //     _ => {}
+        // }
```
merklefruit, line 90: @refcell any tips on how to read env vars when running in no_std? this is the only thing left out because of it.

refcell: Can we just introduce an `std` feature flag here and run ci tests with std enabled?

merklefruit: yes good idea!

refcell (approved):
Looking good. I think it's reasonable to get this great start in trunk and keep iterating + adding tests.

clabby (approved):
Looks great!

Let's get final approval from @geoknee / @sbvegan, then I think we're good to merge :)

PR #260, opened 2024-06-05
Set pre-genesis hardfork times to zero when inheriting superchain-wide values

geoknee:
Closes https://github.com/ethereum-optimism/client-pod/issues/911

Do not merge until we have ensured the resulting diff to rollup configs will not cause DB rewinds in op-geth.

### DIFF
--- a/superchain/superchain.go
+++ b/superchain/superchain.go
@@ -156,22 +156,31 @@ func (c *ChainConfig) setNilHardforkTimestampsToDefaultOrZero(s *SuperchainConfi
 
 	for i := 0; i < reflect.Indirect(cVal).NumField(); i++ {
 		overridePtr := cVal.Field(i)
-		defaultPtr := sVal.Field(i)
-		if defaultPtr.IsNil() || !overridePtr.IsNil() {
-			// no change if override is set or default is unset
+		if !overridePtr.IsNil() {
+			// no change if override is set
 			continue
 		}
-		defaultValue := reflect.Indirect(defaultPtr).Uint()
-		if defaultValue >= *c.SuperchainTime {
-			// if hardfork activated after SuperchainTime...
-			if defaultValue > c.Genesis.L2Time {
-				// ...use default value if is after genesis
-				overridePtr.Set(defaultPtr)
-			} else {
-				// ...use zero if it is equal to or before genesis
-				overridePtr.Set(ptrZero)
-			}
+
+		defaultPtr := sVal.Field(i)
+		if defaultPtr.IsNil() {
+			// no change if default is unset
+			continue
 		}
+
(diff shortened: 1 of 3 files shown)


superchain/superchain.go
```diff
@@ -151,7 +152,14 @@ func (c *ChainConfig) setNilHardforkTimestampsToDefault(s *SuperchainConfig) {
 		if overrideValue.IsNil() &&
 			!defaultValue.IsNil() &&
 			reflect.Indirect(defaultValue).Uint() >= *c.SuperchainTime {
-			overrideValue.Set(defaultValue) // use default only if hardfork activated after SuperchainTime
+			// use default only if hardfork activated after SuperchainTime
```
sebastianst, line 155:
```suggestion
			// use default only if hardfork activated after SuperchainTime and after chain genesis
```

superchain/superchain.go
```diff
@@ -151,7 +152,14 @@ func (c *ChainConfig) setNilHardforkTimestampsToDefault(s *SuperchainConfig) {
 		if overrideValue.IsNil() &&
 			!defaultValue.IsNil() &&
 			reflect.Indirect(defaultValue).Uint() >= *c.SuperchainTime {
```
sebastianst, line 154: We reuse this uint value now two times, so maybe we define it like so
```suggestion
		if defaultUint := reflect.Indirect(defaultValue).Uint(); overrideValue.IsNil() &&
			!defaultValue.IsNil() &&
			defaultUint >= *c.SuperchainTime {
```
and then reuse it in the new `if`.

superchain/superchain.go
```diff
@@ -151,7 +152,14 @@ func (c *ChainConfig) setNilHardforkTimestampsToDefault(s *SuperchainConfig) {
...
-			overrideValue.Set(defaultValue) // use default only if hardfork activated after SuperchainTime
+			// use default only if hardfork activated after SuperchainTime
+			if reflect.Indirect(defaultValue).Uint() > c.Genesis.L2Time {
+				overrideValue.Set(defaultValue)
+			} else {
+				var zero uint64 = 0
+				ptrValue := reflect.ValueOf(&zero)
```
sebastianst, line 160: could move this above the `for` loop for reuse (and maybe then name the pointer `ptrZero`.

sebastianst (commented):
Nice change

geoknee:
Thanks for the feedback, incorporated here [eb46e2d](https://github.com/ethereum-optimism/superchain-registry/pull/260/commits/eb46e2d286f186aed79bd4584477ab0ae084da18)

geoknee:
We should merge this PR first https://github.com/ethereum-optimism/op-geth/pull/332

sebastianst (approved):

PR #265, opened 2024-06-10
Generate a `chainList.json` and `chainList.toml` file

geoknee:
Closes https://github.com/ethereum-optimism/client-pod/issues/839

Replaces `chainIds.json` with `chainList.json` and `chainList.toml` which contain richer information.

~Drive-by change: transcribe all code which generates json files and remove all Javascript code from the repo.~ reverted


superchain/configs/chainList.json
```diff
@@ -0,0 +1,130 @@
+[
+  {
+    "name": "Base Sepolia",
+    "identifier": "sepolia/base",
+    "chainId": 84532,
+    "public_rpc": "https://sepolia.base.org",
+    "explorer": "https://sepolia-explorer.base.org",
+    "superchain_level": 2
+  },
```
mds1, line 9: Have we considered approximately mirroring the JSON structure returned by https://chainid.network/chains.json or other similar endpoints for https://github.com/ethereum-lists/chains? For example, this would result in the following changes:
- `rpc` and `explorer` become arrays, which is nice so we don't enshrine a single provider
- Use the `parent` key to specify the appropriate L1
- All new new fields are left untouched

```json
  {
    "name": "Base Sepolia",
    "identifier": "sepolia/base",
    "chainId": 84532,
    "rpc": ["https://sepolia.base.org"],
    "explorers": ["https://sepolia-explorer.base.org"],
    "superchain_level": 2,
    "parent": { ... }
  },
```

geoknee: I tried that out here [30701d3](https://github.com/ethereum-optimism/superchain-registry/pull/265/commits/30701d3369cc2decfc839ec7e27cc630d13b388a) LMKWYT. 

My general comments are:
* we don't support multiple RPCs or explorers in the actual underlying registry structures at the moment (if we like this change, probably we should do that)
* we don't scrape any bridge URLs to put in the "parent" field at the moment
* we aren't using the chain agnostic identification scheme https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-2.md as the ethereum-list repo does
* I hardcoded the parent type as "L2" for now.

mds1: Sounds good. Not a blocker for me, just wanted to flag it as an option since it's possible e.g. MetaMask and others already have scripts to parse that JSON structure. 

I do think it's ok to make RPC and explorers arrays in the output though, even though it's not supported as such in the input. Will defer to you to resolve :)

superchain/internal/codegen/main.go
```diff
@@ -0,0 +1,120 @@
...
+	return filepath.Dir(currentFilePath)
+}
+
+func addresses() {
+	addressList := make(map[string]AddressList)
+	for id := range OPChains {
+		list := Addresses[id]
+		addressList[strconv.Itoa(int(id))] = *list
+	}
+	addressListBytes, err := json.MarshalIndent(addressList, "", " ")
```
sebastianst, line 109: This new address writer sets all non-existent contracts to the zero address. Non-existent contracts seem to have been omitted before, which I think is better?

superchain/internal/codegen/main.go
```diff
@@ -0,0 +1,120 @@
...
+func currentDir() string {
+	_, currentFilePath, _, ok := runtime.Caller(0)
+	if !ok {
+		fmt.Println("Error: Unable to determine the current file path")
+		os.Exit(1)
+	}
+	return filepath.Dir(currentFilePath)
+}
+
+func addresses() {
```
sebastianst, line 103: nit: give it some imperative name, like
```suggestion
func generateAddresses() {
```

superchain/configs/chainList.toml
```diff
@@ -0,0 +1,175 @@
+[[chains]]
+  name = "Base"
+  identifier = "mainnet/base"
+  chainId = 8453
+  rpc = ["https://mainnet.base.org"]
+  explorers = ["https://explorer.base.org"]
+  superchain_level = 2
+  [chains.parent]
+    type = "L2"
+    chain = "mainnet"
```
sebastianst, line 10: Interesting, didn't know that toml can be written with such indentation. The [official TOML website](https://toml.io/en/v1.0.0#array-of-tables) doesn't indent. But I guess since it's just produced by the toml go package, it's also correct.

sebastianst (commented):
I guess even though all js code is removed, we still want to generate the node package (CI job `publish-bot:`)?

.github/CODEOWNERS
```diff
@@ -15,6 +15,3 @@ Cargo.toml @ethereum-optimism/rust-reviewers
 /validation @ethereum-optimism/go-reviewers
 /superchain @ethereum-optimism/go-reviewers
 /add-chain @ethereum-optimism/go-reviewers
-
-## Solidity
-*.sol @ethereum-optimism/security-reviewers
```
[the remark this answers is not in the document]
geoknee: Correct.

geoknee:
> I guess even though all js code is removed, we still want to generate the node package (CI job `publish-bot:`)?

I suppose so -- I don't see any trigger for that workflow nor published package on NPM yet. Still, I put the manifest back in [bf4ceca](https://github.com/ethereum-optimism/superchain-registry/pull/265/commits/bf4ceca243427407909035575c6cc528b1a046af) and we can double check offline about the future of the npm package.

superchain/internal/codegen/main.go
geoknee: [9ba3c38](https://github.com/ethereum-optimism/superchain-registry/pull/265/commits/9ba3c38b2652497c3d2c8360d69fc3ef58bd25f7)[9ba3c38](https://github.com/ethereum-optimism/superchain-registry/pull/265/commits/9ba3c38b2652497c3d2c8360d69fc3ef58bd25f7)

superchain/internal/codegen/main.go
geoknee: I spent some time on this in the end. The zero addresses being written does stand out somewhat. It is actually out of scope for this PR, so I have reinstated the existing behaviour to unblock us.

I think it is worth coming back to, though. Thoughts and solutions I discarded for now:

1. Writing out zeros for "missing" addresses reflects the behaviour of the Go bindings. I wonder if we should use a pointer to an address inside `AddressList`, especially for contracts which we don't expect to always be there (i.e. optional fields). It seems to better reflect the underlying data -- especially if we ever got into a situation where a zero address was actually a meaningful value for us, it is a meaningful value in the EVM. I actually tried this type change out but it quickly snowballs and is a breaking change for downstream packages. 

2. The `omitempty` directive only works for "empty" fields which are being marshalled, and `[20]byte{}` is not considered empty. It is possible to get around this by writing custom marshalling code, but it is pretty nasty and lots of boilerplate.

.circleci/config.yml
```diff
@@ -98,13 +98,14 @@ jobs:
...
-      - image: cimg/node:18.18.2
+    executor:
+      name: go/default  # is based on cimg/go
+      tag: '1.21'
     steps:
       - checkout
       - run:
           name: Run codegen
-          command: pnpm codegen
+          command: go run superchain/internal/codegen/main.go
```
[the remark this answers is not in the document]
geoknee: I have added instructions to the readme, it should be invoked after running the add-chain tooling and before raising a PR with the new chain data.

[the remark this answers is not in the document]
geoknee: [b05ebc0](https://github.com/ethereum-optimism/superchain-registry/pull/265/commits/b05ebc09709939500849971cf2524cd5c01dc830)[b05ebc0](https://github.com/ethereum-optimism/superchain-registry/pull/265/commits/b05ebc09709939500849971cf2524cd5c01dc830)[b05ebc0](https://github.com/ethereum-optimism/superchain-registry/pull/265/commits/b05ebc09709939500849971cf2524cd5c01dc830)

superchain/internal/codegen/main.go
```diff
@@ -0,0 +1,120 @@
...
+	. "github.com/ethereum-optimism/superchain-registry/superchain"
+)
+
+type ChainEntry struct {
+	Name            string   `json:"name" toml:"name"`
+	Identifier      string   `json:"identifier" toml:"identifier"`
+	ChainId         uint64   `json:"chainId" toml:"chainId"`
+	RPC             []string `json:"rpc" toml:"rpc"`
+	Explorer        []string `json:"explorers" toml:"explorers"`
+	SuperchainLevel uint     `json:"superchain_level" toml:"superchain_level"`
```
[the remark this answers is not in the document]
geoknee: Good catch, this was an oversight on my part. Fixed it here [9a34302](https://github.com/ethereum-optimism/superchain-registry/pull/265/commits/9a343023a12f02756db60549b99337411d441b93)

README.md
```diff
@@ -89,7 +89,14 @@ Omit the `-run=` flag to run checks for all chains.
...
 
-### 5. Open Your Pull Request
+### 7. Run codegen
+This is a tool which will rewrite certain summary files of all the chains in the registry, including the one you are adding. The output will be checked in a continuous integration checks (it is required to pass):
+
+```
+sh scripts/codegen.sh
+```
+
+### 6. Open Your Pull Request
```
sebastianst, line 99: It's now 4. 7. 6. instead of 4. 5.

sebastianst (approved):

superchain/internal/codegen/main.go
sebastianst: ok bummer that it created so much complexity...

README.md
geoknee: [0fd0931](https://github.com/ethereum-optimism/superchain-registry/pull/265/commits/0fd093155a954bd274d400ea169071e6acd4ad61)

PR #276, opened 2024-06-16
refactor (validation module): centralize Retry logic using higher order generic helper fn

geoknee:


bitwiseguy (approved):
Nice improvement!

Issue #280, opened 2024-06-18
validation: remove dependency on `op-bindings`


geoknee:
These have actually been removed from the monorepo, so depending on them blocks us from updating the `validation->optimism` dependency.

We can move bindings that we need into the `validation` package itself, or adopt a more lightweight "cast call" approach to scraping data from chain.

geoknee:
Longer term we can
- strip down the bindings to only the necessary bits of code
- use a lighter-weight approach to scraping data from contracts:
e.g. create some more helpers like this: 

https://github.com/ethereum-optimism/superchain-registry/blob/4641467b9be406de390057778d1b35f8c8a90f70/validation/security-configs_test.go#L106-L107

PR #281, opened 2024-06-18
validation: check all predeploys owned by the L2 Proxy Admin

geoknee:
Towards https://github.com/ethereum-optimism/client-pod/issues/913

This PR:
* adds a check for "L2 Security Configs". This is checking that the L2 ProxyAdmin is in fact the admin of all "0x42..." prefixed predeploys (apart from GovernanceToken and WETH). 
* Overhauls the standard config files and types such that "roles" are defined universally, but "parameters" are defined per superchain
* allows some tests to run in parallel

I'm happy to split up the PR at the reviewer's request.


validation/predeploy-proxyadmin_test.go
```diff
@@ -0,0 +1,63 @@
...
+	proxyAdminAddr := predeploys.ProxyAdminAddr
+
+	proxyAdmin, err := bindings.NewProxyAdmin(proxyAdminAddr, client)
+	require.NoError(t, err)
+
+	getProxyAdmin := func(a common.Address) (common.Address, error) {
+		return proxyAdmin.GetProxyAdmin(&bind.CallOpts{}, a)
+	}
+
+	for k, p := range predeploys.Predeploys {
```
geoknee, line 32: This increases our dependence on the op-bindings package, which we want to avoid #280

validation/standard/standard-config-roles.toml
sebastianst, line 1: Instead of using nested mappings, we could also use toml [arrays and tables](https://toml.io/en/v1.0.0#array-of-tables), which might be a bit more scalable solution. Could be something like
```toml
[[roles]]
layer = 1
group = "universal"

[[roles.relation]]
name = "AddressManager"
method = "owner()"
value = "ProxyAdmin"

[[roles.relation]]
name = "SystemConfigProxy"
method = "owner()"
value = "SystemConfigOwner"

...

[[roles]]
layer = 2
group = "universal"

[[roles.relation]]
address = "0x4200000000000000000000000000000000000016"
method = "admin()"
value_address = "0x4200000000000000000000000000000000000018"

[[roles.relation]]
...
```

The parsing code can then easily convert this into the internal Go mappings.

validation/security-configs_test.go
```diff
@@ -81,16 +82,63 @@ func testSecurityConfigOfChain(t *testing.T, chainID uint64) {
...
 
-func TestSecurityConfigs(t *testing.T) {
+func TestL1SecurityConfigs(t *testing.T) {
 	isExcluded := map[uint64]bool{
 		11763072: true, // Base_devnet_0 (no AnchorStateRegistryProxy specified)
 	}
 	for chainID, chain := range OPChains {
 		t.Run(perChainTestName(chain), func(t *testing.T) {
+			chain := chain
+			chainID := chainID
```
sebastianst, line 92: I believe you'd want to move this above the `t.Run` if we're worried about reusing loop vars, because now you define it inside the `func` which is not guaranteed to run before the next loop iteration.
```suggestion
		chain, chainID := chain, chainID
		t.Run(perChainTestName(chain), func(t *testing.T) {
```
However, [Go 1.22. changed the behavior of loop vars](https://go.dev/blog/loopvar-preview) and they are now always a new instance and this isn't necessary any more. Are we using Go 1.22?

validation/security-configs_test.go
```diff
@@ -81,16 +82,63 @@ func testSecurityConfigOfChain(t *testing.T, chainID uint64) {
...
+	client, err := ethclient.Dial(chain.PublicRPC)
+	require.NoError(t, err, "Failed to connect to the Ethereum client at RPC url %s", chain.PublicRPC)
+	defer client.Close()
+	contractCallResolutions := standard.Config.Roles.L2.Universal
+
+	for contract, methodToOutput := range contractCallResolutions {
+		for method, output := range methodToOutput {
+			t.Run(fmt.Sprintf("%s.%s", contract, method), func(t *testing.T) {
+				method := method
+				output := output
```
sebastianst, line 113: same here

validation/security-configs_test.go
```diff
@@ -81,16 +82,63 @@ func testSecurityConfigOfChain(t *testing.T, chainID uint64) {
...
+func TestL2SecurityConfigs(t *testing.T) {
+	isExcluded := map[uint64]bool{
+		11155421: true, // sepolia-dev-0/oplabs-devnet-0   No Public RPC declared
+		11763072: true, // sepolia-dev-0/base-devnet-0     No Public RPC declared
+	}
+
+	for chainID, chain := range OPChains {
+		t.Run(perChainTestName(chain), func(t *testing.T) {
+			chain := chain
+			chainID := chainID
```
sebastianst, line 135: same here

validation/standard/standard-config-roles.toml
```diff
@@ -0,0 +1,50 @@
...
+"0x4200000000000000000000000000000000000010"."admin()" = "0x4200000000000000000000000000000000000018"
+"0x4200000000000000000000000000000000000011"."admin()" = "0x4200000000000000000000000000000000000018"
+"0x4200000000000000000000000000000000000012"."admin()" = "0x4200000000000000000000000000000000000018"
+"0x4200000000000000000000000000000000000013"."admin()" = "0x4200000000000000000000000000000000000018"
+"0x420000000000000000000000000000000000000F"."admin()" = "0x4200000000000000000000000000000000000018"
+"0x4200000000000000000000000000000000000015"."admin()" = "0x4200000000000000000000000000000000000018"
+"0x4200000000000000000000000000000000000000"."admin()" = "0x4200000000000000000000000000000000000018"
+"0x4200000000000000000000000000000000000014"."admin()" = "0x4200000000000000000000000000000000000018"
+"0x4200000000000000000000000000000000000017"."admin()" = "0x4200000000000000000000000000000000000018"
+"0x4200000000000000000000000000000000000018"."admin()" = "0x4200000000000000000000000000000000000018"
```
sebastianst, line 46: `...18` is its own admin?

geoknee: It is!

validation/security-configs_test.go
geoknee: Good call, fixed in [19b545c](https://github.com/ethereum-optimism/superchain-registry/pull/281/commits/19b545c3e6bcbe636d98defd6651570ccfe1f693). 

Interestingly, `go vet` did not pick up on this, I think it is a known issue that the static analysis for this problem is not perfect. Being above `t.Parallel` seemed enough to satisfy it, but I don't really understand why that should be. 

Anecdotally, I didn't actually see any problems yet with unexpected behaviour on these tests. Still, best to be safe.

validation/security-configs_test.go
geoknee: [19b545c](https://github.com/ethereum-optimism/superchain-registry/pull/281/commits/19b545c3e6bcbe636d98defd6651570ccfe1f693)

validation/security-configs_test.go
geoknee: [19b545c](https://github.com/ethereum-optimism/superchain-registry/pull/281/commits/19b545c3e6bcbe636d98defd6651570ccfe1f693)

sebastianst (approved):

Issue #282, opened 2024-06-18
Enforce consistent "quoting" in add-chain output


geoknee:
Existing YAML configs have the strings wrapped with quotation marks: 

https://github.com/ethereum-optimism/superchain-registry/blob/dbb44f112cefc6126779e8f22a908e0decedebb4/superchain/configs/mainnet/op.yaml#L13-L20

but the add-chain tooling outputs YAML without those marks: 

https://github.com/ethereum-optimism/superchain-registry/blob/dbb44f112cefc6126779e8f22a908e0decedebb4/add-chain/testdata/superchain/configs/sepolia/expected_baseline.yaml#L12-L19

We should enforce consistency one way or another.

PR #283, opened 2024-06-18
Wrap config hex values in double quotes before writing to yml

bitwiseguy:
Fixes https://github.com/ethereum-optimism/superchain-registry/issues/282


superchain/superchain.go
```diff
@@ -187,6 +188,8 @@ func (c *ChainConfig) setNilHardforkTimestampsToDefaultOrZero(s *SuperchainConfi
 // EnhanceYAML creates a customized yaml string from a RollupConfig. After completion,
 // the *yaml.Node pointer can be used with a yaml encoder to write the custom format to file
 func (c *ChainConfig) EnhanceYAML(ctx context.Context, node *yaml.Node) error {
+	ethAddressRegex := regexp.MustCompile(`^0x[a-fA-F0-9]+$`)
```
geoknee, line 191: Nit: maybe we should call this "hexStringRegex"? It will match the genesis hash, which is not an address.

geoknee (changes requested):

superchain/superchain.go
bitwiseguy: Good catch. Will push a fix

geoknee (approved):

PR #279, opened 2024-06-17
Implement new "add chain" flow

geoknee:
Closes https://github.com/ethereum-optimism/client-pod/issues/859

Changes: 
* new instructions for adding a standard chain
* modify the `add-chain` tooling accordingly
  * new CLI flag, triggers new `standard_chain_candidate` field to be written to config file
  * `promote-to-standard` subcommand 
* modify test helpers to apply tests to new categories of chain:
  * `frontier`, `frontier+ / standard_candidate`, `standard`
 * new "key handover check" test
   * applies only to standard chains

PR #300, opened 2024-06-20
validation: add standard multisig addresses for guardian and challenger

geoknee:
Towards https://github.com/ethereum-optimism/client-pod/issues/913

- [x] Note that many chains currently fail this check, so will need to be ~bumped down to "frontier".~ excluded for now


geoknee:
When https://github.com/ethereum-optimism/superchain-registry/pull/279 is merged, we should pull the key handover addresses into these config toml files (the test code can remain separate.

validation/standard/init.go
```diff
@@ -7,16 +7,35 @@ import (
...
+	}
+
+	Config.MultisigRoles["sepolia"] = new(MultisigRoles)
+	err = decodeTOMLFileIntoConfig("standard-config-roles-sepolia.toml", Config.MultisigRoles["sepolia"])
+	if err != nil {
+		panic(err)
+	}
+
+	Config.MultisigRoles["sepolia-dev-0"] = new(MultisigRoles)
+	err = decodeTOMLFileIntoConfig("standard-config-roles-sepolia-dev-0.toml", Config.MultisigRoles["sepolia-dev-0"])
```
bitwiseguy, line 38: To further cut down on redundant code related to `MultisigRoles` and `Params` you could loop through all network names like this:
```
	networks := []string{"mainnet", "sepolia", "sepolia-dev-0"}
	for _, network := range networks {
		Config.MultisigRoles[network] = new(MultisigRoles)
		err = decodeTOMLFileIntoConfig("standard-config-roles-"+network+".toml", Config.MultisigRoles[network])
		if err != nil {
			panic(err)
		}

		Config.Params[network] = new(Params)
		err = decodeTOMLFileIntoConfig("standard-config-params-"+network+".toml", Config.Params[network])
		if err != nil {
			panic(err)
		}
	}
```
And perhaps move the `panic` inside `decodeTOMLFileIntoConfig` to reduce lines even more

bitwiseguy (approved):
Left one comment for potential optimization. Other than that LGTM!

validation/standard/init.go
geoknee: [0a79b06](https://github.com/ethereum-optimism/superchain-registry/pull/300/commits/0a79b0608706551dee01073f16580ad75aaddf34)

Issue #285, opened 2024-06-07
SuperchainRegistry: modify repo structure


geoknee:
We need a compromise between the current structure, and the one[ suggested in the PRD. ](https://docs.google.com/document/d/13nFUdxksND4BzxRA-OdO6DpL888umoVgy56HikMsNW8/edit#heading=h.emx3tkspnffk)

This is because:
*  The repo already has some of the structures that the proposal has, with slightly different names etc.
*  We are constrained somewhat by what Go allows us to do (in terms of embedding files and so on)
* Some level of nesting is perhaps still desired, since we would like to keep the various superchains (mainnet, sepolia, sepoilia-dev-0) separate 

Here is my proposal: 
* Combine "addresses" and "genesis-system-configs" with top level config file.
*  Use TOML instead of yaml and json. 
* rename "configs" to "chains"

We should do this before early access launch, because after that we expect more chains to be added and it will be good to do this while things are quiet. We may want to close or merge any open PRs before doing this.

tessr:
Thank you @geoknee for putting this proposal together! I agree that it makes sense to find a middle ground between what exists today and what was written up in the PRD. Overall, your proposal sounds good but I have one requested change. Here's the rationale. 

From a Product perspective, the repo structure needs to prioritize two things:

1. The legibility of the chain list (answering the question _who is in the Superchain?_)
2. The legibility of the validation checks (answering the question _how do we define the standard config_?) 

Everything else can be an implementation detail and fit whatever format is easiest to maintain for Eng, but these two things should be stupid-easy to find and understand, even for someone who hasn't spent much time within the repo.

For (1), I think that `superchain/configs/chainList.json` is not easy enough to find today. I'd like to see that file as a top-level file, or maybe under `superchain` directly. If there are technical reasons why this isn't possible, we could put a link at the top of the readme or have some other kind of signpost, but I'd rather just have it front and center. I don't imagine that it will be obvious to look in `configs` for the list of members of the Superchain. 

For (2), I think your proposal is pretty much as legible as it gets! So this looks great to me. I _do_ want to just double check that we're good with governance ratifying a toml file rather than an executable set of tests -- I'll check this over Slack and keep you in the loop.

geoknee:
Thanks @tessr/

1. Happily there won't be any constraints on where to put the chain list file (it's an "output" file)
[...]

PR #305, opened 2024-06-21
Hoist chainList files to top level

geoknee:
Towards #https://github.com/ethereum-optimism/superchain-registry/issues/285


superchain/internal/codegen/main.go
```diff
@@ -70,12 +70,13 @@ func main() {
 	fmt.Printf("Found %d chains...\n", len(allChains))
 
 	currentDir := currentDir()
+	repositoryRoot := filepath.Join(currentDir, "../../../")
```
sebastianst, line 73: There's a more portable method to find the root, which works from any cwd, see [here](https://github.com/ethereum-optimism/optimism/blob/457f33f4fdda9373dcf2839619ebf67182ee5057/op-service/util.go#L126). But this works for now too.

sebastianst (approved):

superchain/internal/codegen/main.go
geoknee: Hmm looks like a useful tool -- but I don't think it would work out of the box in this repo (we have a `go.work` file at the top level, and several `go.mod` files one level down). 

We can revisit if it ends up being a pain to maintain this code, I think it will be rare that we will move things around.

PR #309, opened 2024-06-21
feat(bindings): Release PLZ

refcell:
**Description**

Introduces a `release-plz` github action that will open a pr to publish rust bindings crates.io updates. This PR is only opened when code in `bindings/rust-*` is touched.

The workflow closely tracks kona's [release-plz workflow](https://github.com/ethereum-optimism/kona/blob/main/.github/workflows/release.yaml).


clabby (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,36 @@
...
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - name: Commit Cargo.lock
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+          git add Cargo.lock
```
geoknee, line 30: We have two `Cargo.lock` files, once under `/bindings/rust-primitives` and one under `/bindings/rust-bindings`.  So I don't think this command will work. 

I'm not sure exactly what other changes you need to make to take account of the two crates. 

Also, I don't see this step in the release-plz instructions https://release-plz.ieni.dev/docs/github/quickstart, curious about why it is needed?

geoknee (changes requested):

.github/workflows/release.yml
refcell: Oh good catch!

geoknee: I think it would be ok to leave the steps in if you want, the explanatory comment is good though. 

Can we modify the step to reference the actual `cargo.lock` files? Otherwise, if we activate the step at some later point it will fail and need fixing?

geoknee (approved):

PR #313, opened 2024-06-24
feat:init java-bindings

### LINKED ISSUE #312: Superchain-registry: init java-binding for OP stack java components(hildr and op-besu)
This should be a sub-task to fix #36 .
We are refactoring hildr and op-besu, use an unified superchain configuration and structure can reduce our duplication of code. According to the design of rust-bindings, primitives data types can be used in an environment that does not rely on Jackson dependencies.

GrapeBaBa:
**Description**

Init java-bindings 

**Tests**

Please describe any tests you've added. If you've added no tests, or left important behavior untested, please explain why not.

**Additional context**

Add any other context about the problem you're solving.

**Metadata**

- Fixes #312


mslipper:
Hello! We're no longer exposing bindings of any kind directly from the SR. Please add these to a package of your own outside of this repository.

Issue #286, opened 2024-05-02
SuperchainRegistry: test read and write preserves `genesis.json` and `rollup.json` data semantics


geoknee:
We would like to enhance the `add-chain` tool to ensure:
* the tooling used by the registry to write compressed genesis information (i.e. `registry-data` tool in `op-chain-ops`)
* the tooling used to uncompress and rehydrate the genesis information (i.e. this invocation in `op-geth` https://github.com/ethereum-optimism/op-geth/blob/22bc5d43dee668bddb5b2df75103b3fdc20af4f8/cmd/utils/flags.go#L2215)

both work well enough to perfectly reproduce the input data (which is a raw `genesis.json` file and `rollup.json` file)

For more context see https://www.notion.so/oplabs/Superchain-Genesis-Hash-checks-60e173e9fb0c43e88a81d0666cb3d582

geoknee:
Let's concentrate on the `rollup.json` file first. We have an example of how to read the config out using an op-node function in the `generate-rollup-config` diff tool here https://github.com/ethereum-optimism/op-geth/blob/22bc5d43dee668bddb5b2df75103b3fdc20af4f8/cmd/utils/flags.go#L2215 . We also have input test data already for this file https://github.com/ethereum-optimism/op-geth/blob/22bc5d43dee668bddb5b2df75103b3fdc20af4f8/cmd/utils/flags.go#L2215 , which means the existing tests would already provide coverage for this new feature / check.

PR #316, opened 2024-06-24
Regenerate rollupConfig via op-node and check against local rollup.json

bitwiseguy:
Addresses the `rollup.json` check described in this issue: https://github.com/ethereum-optimism/superchain-registry/issues/286


add-chain/main.go
```diff
@@ -158,6 +161,26 @@ func entrypoint(ctx *cli.Context) error {
...
+	}
+	fmt.Printf("Attempting to read from %s\n", rollupConfigPath)
+	file, err := os.ReadFile(rollupConfigPath)
+	if err != nil {
+		return fmt.Errorf("failed to read from local rollup.json config file: %w", err)
+	}
+	var localRollupConfig *rollup.Config
+	if err = json.Unmarshal(file, &localRollupConfig); err != nil {
+		return fmt.Errorf("failed to unmarshal local rollup.json into rollup.Config struct: %w", err)
+	}
```
geoknee, line 178: There's potentially scope for DRYing this out a bit, I believe we are doing something similar to this inside of `constructChainConfig` on L148? I think the difference is that here you are unmarshalling into the op-node type instead of the superchain registry type. I suppose that makes sense if you are going to do a `reflect.DeepEqual` check.

add-chain/tmp.json
```diff
@@ -0,0 +1,2 @@
+&{Genesis:{L1:0x1fc104fe244042459314bc1d51a230320c64a3d47cd06b7998aa3f57b9b0fc79:5753349 L2:0x0ceefff2113580e4d665c324661e4982a824025bf97285cf09a52715b6f89083:0 L2Time:1713792864 SystemConfig:{BatcherAddr:0x27bf581015C64ec7C57C6E7b3117F35B5f6Dd706 Overhead:0x0000000000000000000000000000000000000000000000000000000000000834 Scalar:0x00000000000000000000000000000000000000000000000000000000000f4240 GasLimit:30000000}} BlockTime:2 MaxSequencerDrift:600 SeqWindowSize:3600 ChannelTimeout:300 L1ChainID:+11155111 L2ChainID:+42069 RegolithTime:0x1400058e1b8 CanyonTime:0x140003a4020 DeltaTime:0x140003a4028 EcotoneTime:<nil> FjordTime:<nil> InteropTime:<nil> BatchInboxAddress:0xFf00000000000000000000000000000000042069 DepositContractAddress:0x6C84AE91901CE8a4897187c33A0094a6e790f34d L1SystemConfigAddress:0x8787D3382F60d4e072b4276BBda065959195dB33 ProtocolVersionsAddress:0x0000000000000000000000000000000000000000 BlobsEnabledL1Timestamp:<nil> DAChallengeAddress:0x0000000000000000000000000000000000000000}
```
geoknee, line 1: Probably didn't mean to commit this?

add-chain/main.go
```diff
@@ -158,6 +161,26 @@ func entrypoint(ctx *cli.Context) error {
...
+
+	if !reflect.DeepEqual(opNodeRollupConfig, localRollupConfig) {
```
geoknee, line 180: We could use https://pkg.go.dev/github.com/google/go-cmp/cmp#Diff to get some more useful output in the case where the configs are not equal.

geoknee:
When I tried to run this, I got 

```
 failed when using op-node to generate rollup config: unknown chain ID: 42069
```

which makes sense if we are erroring before we have written the files into the registry (the op-node can only load files which have been embedded). What are your thoughts about working around this? Probably the check needs to run after the files have been written by the `add-chain` tool, I _think_ that would be sufficient.

bitwiseguy:
> When I tried to run this, I got
> 
> ```
>  failed when using op-node to generate rollup config: unknown chain ID: 42069
> ```
> 
> which makes sense if we are erroring before we have written the files into the registry (the op-node can only load files which have been embedded). Also seeing this error on CI.
> 
> What are your thoughts about working around this? Probably the check needs to run after the files have been written by the `add-chain` tool, I _think_ that would be sufficient.

The issue is that the op-node code imports the `superchain-registry/superchain` package, which contains actual chain config .yaml files from that package. But the `add-chain` tests use the mock `testdata` directory structure as the destination to write .yaml files. So the `rollup.LoadOPStackRollupConfig(rollupConfig.ChainID)` call will not see the test .yaml files generated during testing. 

What do you mean by "the check needs to run after the files have been written by the `add-chain` tool, I _think_ that would be sufficient"? Maybe we move the rollup.json sanity check from the `add-chain` main flow to a `add-chain` subcommand that is only run by CI (instead of getting triggered by the current `add-chain/e2e_test.go`?

add-chain/main.go
bitwiseguy: Since we have to move the check-rollup-config code outside of this main function (to solve import updates), this comment is no longer applicable

add-chain/testdata/monorepo/op-node/rollup_faultproofs.json
```diff
@@ -21,13 +21,13 @@
...
-    "l2_chain_id": 11155421,
+    "l2_chain_id": 4206903,
     "regolith_time": 0,
     "canyon_time": 0,
     "delta_time": 0,
     "batch_inbox_address": "0xff00000000000000000000000000000011155421",
-    "deposit_contract_address": "0x6509f2a854ba7441039fce3b959d5badd2ffcfcd",
-    "l1_system_config_address": "0x4af802b3010e07845b2b8c2250126e9ac0bdb6b9",
+    "deposit_contract_address": "0x76114bd29dFcC7a9892240D317E6c7C2A281Ffc6",
+    "l1_system_config_address": "0xa6b72407e2dc9EBF84b839B69A24C88929cf20F7",
```
geoknee, line 30: How did you choose these test data? Did you do a fresh L1 deployment?

add-chain/check-rollup-config.go
```diff
@@ -0,0 +1,56 @@
...
+		if err != nil {
+			return fmt.Errorf("op-node failed to generate rollup config: %w", err)
+		}
+		opts := cmp.Options{
+			cmpopts.IgnoreUnexported(big.Int{}),
+		}
+
+		// Exclude this field from check since it gets populated by op-node after rollup.json is created
+		localRollupConfig.ProtocolVersionsAddress = common.Address{}
+		opNodeRollupConfig.ProtocolVersionsAddress = common.Address{}
```
geoknee, line 47: I'm not entirely satisfied with this but even with this exception we are getting a lot of protection from other things not matching. We should create an issue and circle back to this.

If we wanted to follow the pattern with hardfork activation times, we would:
* unexport the ProtocolVersions field from SuperchainConfig type. 
* add it to the ChainConfig type and set it to the value from `rollup.json` during the `add-chain` flow
* have logic to conditionally hydrate the chain config type with the value from the superchain config type, depending on whether `superchain_time` is `nil` or not `nil` and whether ProtocolVersions is empty or not.
* have op-node read from the chain config type instead of the superchain config type (it now can't anyway because it is private). 

This way, the discrepancy in this new tool you have written would be resolved by new chains not having the `superchain_time` set and therefore op-node would not be hydrating a protocol versions address from the superchain.yaml file and it should match.

geoknee (approved):
Left a few comments. 

When we resolve the convos, I think it is worth merging this and circling back to the bigger open question. 

We also want to test this out thoroughly, and involve @sbvegan in that.

add-chain/testdata/monorepo/op-node/rollup_faultproofs.json
bitwiseguy: `l2_chain_id` needs to be unique compared to others in the superchain. Since this PR relies on the `add-chain` tests writing artifacts to the superchain package prior to `add-chain check-rollup-config` importing the superchain package and running its tests, I needed to change the mock data chain ids so that the test artifacts could all exist simultaneously.

bitwiseguy: Within the op-node code, the following rollup config address fields are set prior to writing the `rollup.json` file:
```		
DepositContractAddress: common.Address(addrs.OptimismPortalProxy),
L1SystemConfigAddress:  common.Address(addrs.SystemConfigProxy),
```
In rollup_faultproofs.json those address values are:
```
    "deposit_contract_address": "0x6509f2a854ba7441039fce3b959d5badd2ffcfcd",
    "l1_system_config_address": "0x4af802b3010e07845b2b8c2250126e9ac0bdb6b9",
```
However in the faultproofs .deploy file the addresses are:
```
  "OptimismPortalProxy": "0x76114bd29dFcC7a9892240D317E6c7C2A281Ffc6",
  "SystemConfigProxy": "0xa6b72407e2dc9EBF84b839B69A24C88929cf20F7",
```
I changed the values in rollup_faultproofs.json to match the ones in .deploy because I believe rollup_faultproofs.json was constructed in a faulty (pardon the pun) way that is not representative of a real user flow. So I believe this change fixes a bug in the mock test data

.gitignore
```diff
@@ -22,7 +26,7 @@ docs/
 .env*
 !.env.example
 !.env.test
-*.log
+*.lo
```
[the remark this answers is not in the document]
bitwiseguy: Good catch. Will fix

add-chain/e2e_test.go
```diff
@@ -102,8 +123,24 @@ func checkConfigYaml(t *testing.T, testName, chainName string) {
...
+	require.NoError(t, err, "failed to read testchain.yaml config file: %w", err)
 
 	require.Equal(t, string(expectedBytes), string(testBytes), "test .yaml contents do not meet expectation")
 }
+
+func cleanupTestFiles(t *testing.T, chainName string) {
+	paths := []string{
+		"../superchain/extra/addresses/sepolia/" + chainName + ".json",
+		"../superchain/extra/genesis-system-configs/sepolia/" + chainName + ".json",
+		"../superchain/configs/sepolia/" + chainName + ".yaml",
```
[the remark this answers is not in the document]
bitwiseguy: Used some constants here: https://github.com/ethereum-optimism/superchain-registry/pull/316/commits/abaf94be4641d9ad9ebfd66b413340b908b87a61

add-chain/check-rollup-config.go
bitwiseguy: Created a new issue for this follow-up effort: https://github.com/ethereum-optimism/superchain-registry/issues/334

PR #319, opened 2024-06-25
 Run `TestL1SecurityConfigs` only on standard/standard-candidate chains

geoknee:
This is not a requirement for frontier chains.


bitwiseguy (approved):
LGTM!

PR #320, opened 2024-06-25
Implement global chain ID check and use `chainShortName`

### LINKED ISSUE #160: adding a chain: allow filename to be distinct from chain name
Here's where that constraint comes from:

https://github.com/ethereum-optimism/superchain-registry/blob/730d61f58c4b107f96e65959d6a1895b5c48b60b/scripts/add-chain.sh#L59-L61

I think we _either_:

1. Allow the filename to be set by a new, separate environment variable
2. (My preference) Use the incoming "shortname" variable (see #153, this will be checked for global uniqueness) as the filename.

### LINKED ISSUE #299: SuperchainRegistry: Update chain ID check to check for global uniqueness
Currently chain ID checks only check uniqueness against chain IDs in the registry, but we'd like the check to be global instead, using the list from https://github.com/ethereum-lists/chains. Options for pulling the data:
- https://github.com/ethereum-lists/chains/tree/master/_data/chains is the folder with each chain's data
- https://chainid.network/chains.json is a JSON file generated from all that data. This file is currently ~570 kb
- https://chainid.network/chains_mini.json is a smaller JSON file generated from all that data (doesn't contain all fields). This file is currently ~373 kb

geoknee:
WIP

Closes #160 
Closes #299


geoknee:
```
        	Error Trace:	/Users/georgeknee/code/ethereum-optimism/superchain-registry/validation/uniqueness_test.go:114
        	Error:      	Not equal: 
        	            	expected: "race"
        	            	actual  : "dhobyghaut"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-race
        	            	+dhobyghaut
        	Test:       	TestChainsAreGloballyUnique
        	Messages:   	Local short chain name for 90001 does not match name from chainid.network
```

superchain/superchain.go
```diff
@@ -106,9 +107,6 @@ type ChainConfig struct {
 	// Superchain is a simple string to identify the superchain.
 	// This is implied by directory structure, and not encoded in the config file itself.
 	Superchain string `yaml:"-"`
-	// Chain is a simple string to identify the chain, within its superchain context.
-	// This matches the resource filename, it is not encoded in the config file itself.
-	Chain string `yaml:"-"`
```
geoknee, line 111: Removing this is a compile-time breaking change for
* op-geth
* op-node
which, even when patched, counts as a runtime breaking change for the `-network` and `-op-network` flags when starting these programs.

The shortest path to resolving this would be to keep the `Chain` property, and keep the code that sets it to the filename.

* New chains can have `Chain` = `ShortName` = file name
* existing chains keep `Chain` = file name, and get a new `ShortName` which isn't used for anything other than global chain id check
* we deprecate `Chain` and at some future point remove it and fix the downstream problems that will create.

validation/uniqueness_test.go
```diff
@@ -0,0 +1,122 @@
...
+	globalChainIds, err := getGlobalChains()
+	if err != nil {
+		t.Fatal(err)
+	}
+	localChainIds := make(chainIDSet)
+	localChainNames := make(chainNameSet)
+	localChainShortNames := make(chainShortNameSet)
+
+	for _, chain := range OPChains {
+		if globalChainIds[uint(chain.ChainID)] != nil {
```
geoknee, line 109: We need to enforce that the chain id is in the upstream repo.

geoknee:
In favour of #337

PR #330, opened 2024-06-26
Add data-availability_test.go

### LINKED ISSUE #293: SuperchainRegistry: increase validation coverage of standard config `parameters` to 100%
~This relies on the full config (defined here in the specs https://specs.optimism.io/protocol/configurability.html?highlight=standard#op-stack-configurability)  being added to `standard-config-mainnet.toml` (and similar files for testnets and frontier chains).~

The strategy will be to add each config parameter and a validation check that covers it, proceeding line by line until all parameters are in place and covered. 

Each parameter should be subject to a validation check.

Validation checks written for {Standard/Frontier} x {Mainnet, Testnet} chains, with minimal chain-specific exceptions . Covers all parameters in the standard config and copes with all supported contract versions and hardforks.

This may be easier to tackle if we do ethereum-optimism/superchain-registry#294 first, or else we can do a refactor after it is all done to tidy things up.  (Where possible, testing code should be deduplicated and refactored for maximum legibility. )

Edit: this ticket only covers "consensus parameters" and "policy parameters" from this doc https://www.notion.so/oplabs/OP-Stack-Configurability-from-a-SuperchainRegistry-validation-POV-9c335a936007456b89777f5a84fba58d

The remaining checks are covered by this ticket ethereum-optimism/client-pod#913

geoknee:
Closes #293

### DIFF
--- a/validation/data-availability_test.go
+++ b/validation/data-availability_test.go
@@ -1,0 +1,17 @@
+package validation
+
+import (
+	"testing"
+
+	. "github.com/ethereum-optimism/superchain-registry/superchain"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDataAvailability(t *testing.T) {
+	for _, chain := range OPChains {
+		t.Run(perChainTestName(chain), func(t *testing.T) {
+			RunOnStandardAndStandardCandidateChains(t, *chain)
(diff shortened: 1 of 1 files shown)


bitwiseguy (approved):
LGTM

PR #337, opened 2024-06-27
Implement global chain ID check and use chainShortName

geoknee:
Closes https://github.com/ethereum-optimism/superchain-registry/issues/160
Closes https://github.com/ethereum-optimism/superchain-registry/issues/299

## Upstream 
We are now storing a `shortName` for each chain, following EIP 3770, and harmonizing with https://github.com/ethereum-lists/chains

Logic to pass global chain ID check:
* Is there a chain in the upstream repo with this ID? If no, fail
* Does that chain have an identical `name` and `shortName`? If no, fail
* Does the RPC url supplied exist in the upstream repo for this chain ID? If no, fail
* Else pass

## Downstream
* Do not make any breaking changes
* In particular do not change the name of any of the config files. The name of the config file gets hydrated into a `chainConfig.Chain` string property which is used by op-geth and op-node to identify the chain to run on.
* We can circle back and tidy this up at a later point if we want to.


add-chain/e2e_test.go
```diff
@@ -121,15 +127,15 @@ func compareJsonFiles(t *testing.T, dirPath, testName, chainName string) {
...
+func checkConfigYaml(t *testing.T, testName, chainShortName string) {
 	expectedBytes, err := os.ReadFile("./testdata/superchain/configs/sepolia/expected_" + testName + ".yaml")
 	require.NoError(t, err, "failed to read expected.yaml config file: %w", err)
 
 	var expectedYaml map[string]interface{}
 	err = yaml.Unmarshal(expectedBytes, &expectedYaml)
 	require.NoError(t, err, "failed to unmarshal expected.yaml config file: %w", err)
 
-	testBytes, err := os.ReadFile(ymlConfigDir + chainName + ".yaml")
+	testBytes, err := os.ReadFile(ymlConfigDir + chainShortName + ".yaml")
```
bitwiseguy, line 138: Since this PR changes the filenames written, the filenames in `cleanupTestFiles` should also use the `chainShortName` instead of `chainName`

chainList.json
```diff
@@ -128,8 +128,8 @@
     }
   },
   {
-    "name": "OP-Sepolia",
-    "identifier": "sepolia/op",
+    "name": "OP Sepolia Testnet",
+    "identifier": "sepolia/opsep",
```
bitwiseguy, line 132: This name seems redundant since you have the `sepolia` prefix and the `sep` suffix

superchain/configs/mainnet/op.yaml
```diff
@@ -1,4 +1,5 @@
-name: OP-Mainnet
+name: OP Mainnet
+short_name: oeth
```
bitwiseguy, line 2: Should we use `opeth` here instead? I feel like `op` is a commonly used prefix for optimism across docs/repos

superchain/configs/mainnet/pgn.yaml
```diff
@@ -1,4 +1,5 @@
-name: PGN
+name: PGN (Public Goods Network)
+short_name: PGN
```
bitwiseguy, line 2: Should this be lowercase to be consistent with other chains?

bitwiseguy:
I know this PR didn't introduce the `chainList.json` and `chainList.toml` but as I'm reviewing I just noticed the amount of duplicated data in those two files. Is there any way to de-duplicate so changes such as the name/identifier fields updated in this PR do not have to be done twice?

chainList.json
```diff
@@ -1,7 +1,7 @@
 [
   {
-    "name": "OP-Mainnet",
-    "identifier": "mainnet/op",
+    "name": "OP Mainnet",
+    "identifier": "mainnet/oeth",
```
bitwiseguy, line 4: What is the difference between `identifier` used here and `chainShortName` introduced by this PR? Could they be combined into a single field?

validation/uniqueness_test.go
```diff
@@ -0,0 +1,196 @@
...
+		t.Fatal(err)
+	}
+	localChainIds := make(chainIDSet)
+	localChainNames := make(chainNameSet)
+	localChainShortNames := make(chainShortNameSet)
+
+	isExcluded := map[uint64]bool{
+		90001:    true, // sepolia/race, known chainId collision
+		11155421: true, // oplabs devnet 0, not in upstream repo
+		11763072: true, // base devnet 0, not in upstream repo
```
bitwiseguy, line 123: Are these tests checking the upstream repo? If not, are these last two exclusions necessary?

validation/uniqueness_test.go
```diff
@@ -0,0 +1,196 @@
...
+func (s chainShortNameSet) AddIfUnique(name string) error {
+	if s[name] {
+		return fmt.Errorf("Chain with short name %s is duplicated", name)
+	}
+	s[name] = true
+	return nil
+}
+
+func getGlobalChains() (map[uint]*uniqueProperties, error) {
+	chainListUrl := "https://chainid.network/chains_mini.json"
```
bitwiseguy, line 54: Can you add a comment giving some description about what this webpage is and why its a good source of truth for superchain-registry?

bitwiseguy (changes requested):
Overall look good but left a few comments/questions

superchain/configs/mainnet/op.yaml
geoknee: This is the short name already installed in the upstream ethereum-lists/chains repo. I agree that `op` is the better prefix -- we aren't using the short name downstream ourselves for anything at the moment. However, it is part of the upstream check so we need to store it.

superchain/configs/mainnet/pgn.yaml
geoknee: Again this is dictated by whatever is stored upstream. It can be changed upstream and then changed also here if desired.

chainList.json
geoknee: The identifier includes the superchain target. The chain short name does not in general (sometimes it also does, but not in any standardized way). 

Having said that, I think this should revert to using the existing names [020a0ec](https://github.com/ethereum-optimism/superchain-registry/pull/337/commits/020a0ec9346b705d0e0d02fdf2575676ba369579)

geoknee:
> I know this PR didn't introduce the `chainList.json` and `chainList.toml` but as I'm reviewing I just noticed the amount of duplicated data in those two files. Is there any way to de-duplicate so changes such as the name/identifier fields updated in this PR do not have to be done twice?

Both files are autogenerates by the `codegen` script. So there isn't really any work to do to update fields.

add-chain/e2e_test.go
geoknee: Good catch, [086a3d0](https://github.com/ethereum-optimism/superchain-registry/pull/337/commits/086a3d07a22bdb18a682c9f8e4cbe598ec3e796b)

chainList.json
geoknee: The name is redundant in this case, but not in all cases. I actually rolled back the changes to this file in the end.

add-chain/main.go
```diff
@@ -126,7 +136,12 @@ func entrypoint(ctx *cli.Context) error {
 		deploymentsDir = ctx.String(DeploymentsDirFlag.Name)
 	}
 
+	if chainShortName == "" {
```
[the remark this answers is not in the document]
geoknee: I will avoid messing with this now since we plan on tidying this up soon anyway #317.

chainList.json
geoknee: We will likely run into this in future again, but it may be something we can't really get around.

geoknee: There are two options here:
1. Lean into the _existing_ behaviour: have folks submit i) chain name ii) chain shortname and iii) "unqualified short identifier". The config yml filename uses the unqualified short identifier, e.g. "op" for both OP mainnet and OP sepolia. We inform contributors that this USI can be the same for testnets and mainnets. We will then avoid redundant identifiers.
2. Try to remove the filename as being semantically important and an extra degree of freedom. To avoid breaking changes, we only do this for new chains: set the filename to the chain short name. The identifier will then be built from the chain short name, and we are more likely to get redundant information in the fully qualified identifier above.

geoknee: There's also option:

0: remove `shortChainName` from this PR altogether. Then this PR has minimal changes and we can implement the other options at a later time if we want to.

bitwiseguy: Yes I'd vote option 0

geoknee: Implemented here [ea3c330](https://github.com/ethereum-optimism/superchain-registry/pull/337/commits/ea3c33047740b47b6dbdd1dd17fee838d04d0625)

I still used the "chain short name" terminology, since "unqualified short identifier" seemed a bit... much. But I made it clear in the .env.example file that it is not to be confused with the upstream chain short name.

validation/uniqueness_test.go
geoknee: [39cc30d](https://github.com/ethereum-optimism/superchain-registry/pull/337/commits/39cc30def0f43200e210ee63af6dfd215f7523e8)

superchain/extra/addresses/addresses.json
```diff
@@ -63,7 +72,10 @@
     "ProxyAdminOwner": "0xE75Cd021F520B160BF6b54D472Fa15e52aFe5aDD",
     "SystemConfigOwner": "0x23BA22Dd7923F3a3f2495bB32a6f3c9b9CD1EC6C",
     "Guardian": "0x7a50f00e8D05b95F98fE38d8BeE366a7324dCf7E",
-    "Challenger": "0x45eFFbD799Ab49122eeEAB75B78D9C56A187F9A7"
+    "Challenger": "0x45eFFbD799Ab49122eeEAB75B78D9C56A187F9A7",
+    "UnsafeBlockSigner": "0x93A14E6894eEB4FF6a373E1Ad4f498c3a207afe4",
+    "BatchSubmitter": "0x4e6BD53883107B063c502dDd49F9600Dc51b3DDc",
+    "Proposer": "0xe9e08A478e3a773c1B5D59014A0FDb901e6d1d69"
```
bitwiseguy, line 78: Are these changes intentionally included in this PR? They keep creeping into PRs :laughing:

geoknee: Fixed here [3bdd1b9](https://github.com/ethereum-optimism/superchain-registry/pull/337/commits/3bdd1b96a50f40aa2db0e6c20fabfadcb81948a3)

PR #338, opened 2024-06-27
Add a Makefile

geoknee:
Also fixes bug where we are not enforcing the javascript codegen in CI (means addresses.json is out of date).


Makefile
```diff
@@ -0,0 +1,53 @@
+### Auto-generated files
+codegen-all: go-codegen js-codegen
+
+go-codegen:
+	go run superchain/internal/codegen/main.go
+
+js-codegen:
+	node ./scripts/codegen.js
```
geoknee, line 8: This doesn't really need to be split out.

Makefile
```diff
@@ -0,0 +1,53 @@
...
+	go test ./superchain/... -v
+
+test-validation:
+	go test ./validation/... -v
+
+### Cleaning
+clean-add-chain:
+	rm -f superchain/configs/sepolia/testchain_*.yaml
+	rm -f superchain/extra/addresses/sepolia/testchain_*.json
+	rm -f superchain/extra/genesis-system-configs/sepolia/testchain_*.json
```
geoknee, line 34: Do we still need this?

geoknee (commented):
Also need to update the readme to use `make` commands!

How about filtering tests to a particular chain?

Makefile
```diff
@@ -0,0 +1,53 @@
...
+
+### Testing
+test-all: test-add-chain test-superchain test-validation
+
+test-add-chain:
+# We separate the first test from the rest because it generates artifacts
+# Which need to exist before the remaining tests run.
+	go test ./add-chain/... -run TestAddChain_Main -v
+	go test ./add-chain/... -run '[^TestAddChain_Main]' -v
+	make clean-add-chain
```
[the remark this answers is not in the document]
bitwiseguy: By default the make command will exit with a non-zero exit code (and stop execution) if any one of those commands fails

Makefile
bitwiseguy: I think that one is handy to have around

Makefile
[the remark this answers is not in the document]
geoknee: Thanks, will resolve this.

Makefile
```diff
@@ -0,0 +1,68 @@
...
+	# go run ./add-chain check-rollup-config
+	mkdir -p ./superchain/extra/genesis/$(SUPERCHAIN_TARGET)
+	cd $(MONOREPO_DIR) && go run ./op-chain-ops/cmd/registry-data \
+		--l2-genesis=$(GENESIS_CONFIG) \
+		--bytecodes-dir=$(ROOT_DIR)/superchain/extra/bytecodes \
+		--output=$(ROOT_DIR)/superchain/extra/genesis/$(SUPERCHAIN_TARGET)/$(CHAIN_SHORT_NAME).json.gz
+	make codegen
+
+promote-to-standard:
+	go run ./add-chain promote-to-standard --chain-id=${chain-id}
```
bitwiseguy, line 18: Should the arg be changed to `$(CHAIN_ID)` to read the env var?

.circleci/config.yml
```diff
@@ -24,25 +25,16 @@ jobs:
...
-      - run:
-          name: tidy validation module
-          command: go mod tidy
-          working_directory: validation
-      - run:
-          name: tidy add-chain module
-          command: go mod tidy
-          working_directory: add-chain
+          name: tidy
+          command: make tidy-all
```
bitwiseguy, line 37: Do you need to do `cp .env.example .env` before running the make commands in CI?

geoknee: Very likely, yes!

Makefile
```diff
@@ -0,0 +1,68 @@
...
+tidy-add-chain:
+	cd add-chain && go mod tidy
+
+tidy-superchain:
+	cd superchain && go mod tidy
+
+tidy-validation:
+	cd validation && go mod tidy
+
+### Removing a chain, example: make remove-chain superchain_target=sepolia chain=mychain
```
bitwiseguy, line 63: Instead of requiring two args for this command could you just pass in chain id?

geoknee:
We could also consider using `just` instead of `make` https://github.com/casey/just.

Makefile
geoknee: I don't think we have that in the `.env.example` currently (we read it from the rollup.json file during `add-chain`).

geoknee:
In favour of #366

PR #340, opened 2024-06-28
Update OP Mainnet SystemConfig Owner

maurelian:
**Description**

The owner of the OP Mainnet SystemConfig was updated [by this ops playbook](https://github.com/ethereum-optimism/superchain-ops/blob/main/tasks/eth/011-proto-ver%2Bownership-transfers/README.md#L8-L9) from the Foundation Operations Safe to the Foundation Upgrade Safe. These two Safe's have the same set of owners, this change is for operational purposes only.


bitwiseguy (approved):
LGTM

PR #344, opened 2024-06-28
Fix add-chain bug to robustly set filepaths

bitwiseguy:


add-chain/main.go
```diff
@@ -85,12 +86,14 @@ func entrypoint(ctx *cli.Context) error {
...
-	superchainRepoReadPath, err := os.Getwd()
-	if err != nil {
-		return fmt.Errorf("error getting current directory: %w", err)
+	// Get the current script filepath
+	_, thisFile, _, ok := runtime.Caller(0)
+	if !ok {
+		panic("error getting current filepath")
 	}
+	superchainRepoReadPath := filepath.Dir(thisFile)
 	superchainRepoWritePath := filepath.Dir(superchainRepoReadPath)
```
geoknee, line 95: Unless I am misunderstanding, this looks like
`superchainRepoReadPath = `superchain-registry/add-chain/`
`superchainRepoWritePath = `superchain-registry/`

Do these make sense?

geoknee (approved):

add-chain/main.go
bitwiseguy: Yea good catch. Fixed here: https://github.com/ethereum-optimism/superchain-registry/pull/344/commits/130f031d84d38c34d822c26bdd7c0bf902be8e00

PR #349, opened 2024-06-29
Refactor add-chain file/dir structure

bitwiseguy:
Refactors the `add-chain` package to more closely align with `ethereum-optimism/optimism/op-node` file and directory structure.

Refactors `add-chain` env vars to more closely align with `ethereum-optimism/optimism/op-node`. (Replaces https://github.com/ethereum-optimism/superchain-registry/pull/317)


add-chain/config/config.go
```diff
@@ -37,7 +37,7 @@ func (c *JSONChainConfig) VerifyPlasma() error {
 
 // constructChainConfig creates and populates a ChainConfig struct by reading from an input file and
 // explicitly setting some additional fields to input argument values
-func constructChainConfig(
+func ConstructChainConfig(
```
geoknee, line 40: Nit: godoc should have capitalized first word now, too.

add-chain/config/config.go
```diff
@@ -98,7 +98,7 @@ func constructChainConfig(
 
 // writeChainConfig accepts a rollupConfig, formats it, and writes some output files based on the given
 // target directories
-func writeChainConfig(
+func WriteChainConfig(
```
geoknee, line 101: here too

add-chain/flags/flags.go
```diff
@@ -0,0 +1,64 @@
+package flags
+
+import (
+	"github.com/urfave/cli/v2"
+)
+
+const EnvVarPrefix = "SCR"
+
+func prefixEnvVars(names ...string) []string {
```
geoknee, line 9: Curious why we accept a variadic slice of strings but only ever pass one string in? Is this because this is how it is done in the monorepo?

add-chain/main.go
```diff
@@ -92,60 +73,33 @@ func entrypoint(ctx *cli.Context) error {
 		return fmt.Errorf("failed to get superchain level: %w", err)
 	}
 
+	publicRPC := os.Getenv("SCR_PUBLIC_RPC")
+	sequencerRPC := os.Getenv("SCR_SEQUENCER_RPC")
+	explorer := os.Getenv("SCR_EXPLORER")
+	superchainTarget := os.Getenv("SCR_SUPERCHAIN_TARGET")
+	monorepoDir := os.Getenv("SCR_MONOREPO_DIR")
```
geoknee, line 80: Any reason we can't promote these to flags like the other variables? That way, we can still read them in from the environment (even the .env files), but also have the option to read in via command line flags. And it would simplify / unify the code a bit.

geoknee (approved):
Left a few comments, but otherwise looks good!

add-chain/flags/flags.go
bitwiseguy: Yes just used the same helper function as in op-node

add-chain/main.go
bitwiseguy: https://github.com/ethereum-optimism/superchain-registry/pull/349/commits/1ec8c9ead5f51af318146b9665bfab388185fee3

PR #354, opened 2024-07-02
Add check-genesis-config subcommand

bitwiseguy:


add-chain/testdata/superchain/extra/genesis-system-configs/sepolia/expected_zorasep.json
```diff
@@ -0,0 +1,8 @@
+{
+  "batcherAddr": "0x3Cd868E221A3be64B161D596A7482257a99D857f",
+  "overhead": "0x00000000000000000000000000000000000000000000000000000000000000bc",
+  "scalar": "0x00000000000000000000000000000000000000000000000000000000000a6fe0",
+  "gasLimit": 30000000,
+  "baseFeeScalar": 0,
+  "blobBaseFeeScalar": 0
```
bitwiseguy, line 7: I manually added these fields because they were non-existent previously.

add-chain/testdata/superchain/configs/sepolia/expected_zorasep.yaml
```diff
@@ -0,0 +1,21 @@
+name: testchain_zorasep
+chain_id: 4206904
+public_rpc: http://awe.some.rpc
+sequencer_rpc: http://awe.some.seq.rpc
+explorer: https://awesomescan.org
+
+superchain_level: 1
+standard_chain_candidate: true # This is a temporary field which causes most of the standard validation checks to run on this chain
+superchain_time: null # Missing hardfork times are NOT yet inherited from superchain.yaml
+contracts_version_tag: op-contracts/v1.3.0 # Multi-Chain Prep (MCP) https://github.com/ethereum-optimism/optimism/releases/tag/op-contracts%2Fv1.3.0
```
bitwiseguy, line 10: Diff from actual zorasep: I removed `""` from around `"op-contracts/v1.3.0"`

add-chain/cmd/check-genesis-config.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+
+		chainId := ctx.Uint64(flags.ChainIdFlag.Name)
+		gethGenesis, err := core.LoadOPStackGenesis(chainId)
+		if err != nil {
+			return fmt.Errorf("failed to load genesis via op-geth: %w", err)
+		}
+
+		opts := cmp.Options{
+			// ExcessBlobGas, BlobGasUsed: allow &0 == nil
```
geoknee, line 40: Why do we want to make an exception for this?

add-chain/testdata/superchain/configs/sepolia/expected_zorasep.yaml
geoknee: Is this a change we want to make elsewhere too? What problem did it cause?

add-chain/testdata/superchain/extra/genesis-system-configs/sepolia/expected_zorasep.json
geoknee: I think they didn't activate ecotone at genesis, so we wouldn't expect these fields to be present?

add-chain/cmd/check-genesis-config.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"github.com/ethereum-optimism/superchain-registry/add-chain/flags"
+	"github.com/ethereum/go-ethereum/core"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/urfave/cli/v2"
+)
+
+var CheckGenesisConfigCmd = cli.Command{
+	Name:  "check-genesis-config",
```
geoknee, line 18: Should we call this `check-genesis`? Is "genesis config" an established term?

add-chain/testdata/superchain/configs/sepolia/expected_zorasep.yaml
bitwiseguy: It caused the comparison on expected vs actual config .yaml file to fail [here](https://github.com/ethereum-optimism/superchain-registry/blob/main/add-chain/e2e_test.go#L164).

We currently don't treat strings consistently in the config .yaml file. We have a regex to put double quotes around contract addresses [here](https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/superchain.go#L201), but all other strings are expected to not have quotes. I imagine the actual zora sepolia config .yaml file was created before the `superchain.EnhanceYaml` function existed, so that's why it has double quotes around that address field

add-chain/testdata/superchain/extra/genesis-system-configs/sepolia/expected_zorasep.json
bitwiseguy: Those fields are required by the `superchain` package due to the way the struct is defined [here](https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/superchain.go#L56-L57). Should we make them optional by changing the json tags for those fields so that they include `omitempty`?

bitwiseguy:
@geoknee I also had to add this field to the `genesis_zorasep.json` file to get the tests to pass:
```
      "eip1559DenominatorCanyon": 250
```
This is because that value is hardcoded inside the op-geth `LoadOPStackChainConfig` function [here](https://github.com/ethereum-optimism/op-geth/blob/optimism/params/superchain.go#L79).

add-chain/cmd/check-genesis-config.go
bitwiseguy: This was a work-around that is no longer needed. Removed the exception: https://github.com/ethereum-optimism/superchain-registry/pull/354/commits/eaf83cece4b7612c39eeeb63fb1e1f3fd954f6f5

add-chain/testdata/superchain/extra/genesis-system-configs/sepolia/expected_zorasep.json
geoknee: Hmm I think perhaps we want them to be pointers, so that the default value is `nil`. Otherwise a missing field is indistinguishable from a field specifying `0`.

geoknee:
> @geoknee I also had to add this field to the `genesis_zorasep.json` file to get the tests to pass:
> 
> ```
>       "eip1559DenominatorCanyon": 250
> ```
> 
> This is because that value is hardcoded inside the op-geth `LoadOPStackChainConfig` function [here](https://github.com/ethereum-optimism/op-geth/blob/optimism/params/superchain.go#L79).

OK, does this imply that chains with a pre canyon genesis would fail the check? We might need to think about how to resolve this.

add-chain/testdata/superchain/extra/genesis-system-configs/sepolia/expected_zorasep.json
bitwiseguy: Needed to make those fields points and add the `omitempty` tag to get this to work: https://github.com/ethereum-optimism/superchain-registry/pull/354/commits/ce93f6f20cc5405c613827fbeba563925f2e02da

bitwiseguy:
> OK, does this imply that chains with a pre canyon genesis would fail the check? We might need to think about how to resolve this. Perhaps we only check that the generated fields have _at least_ all of the fields in the input file (but it is allowed to have more)?

The problem is that the field gets a default 0 value if it is not set so the cmp.Diff fails like this:
```
        	            	- 			EIP1559DenominatorCanyon: 0,
        	            	+ 			EIP1559DenominatorCanyon: 250,
```
Should we make a PR in the op-geth repo to change that field to a pointer and add the `omitempty` tag [here](https://github.com/ethereum-optimism/op-geth/blob/optimism/params/config.go#L441)?

add-chain/flags/flags.go
```diff
@@ -93,22 +99,14 @@ var (
...
-		Usage:   "Alternative to l2-genesis flag, if genesis-state is omitted. Path to block header at genesis",
-		EnvVars: prefixEnvVars("L2_GENESIS_HEADER"),
-	}
-)
+var L2GenesisHeaderFlag = &cli.PathFlag{
+	Name:    "l2-genesis-header",
+	Value:   "genesis-header.json",
+	Usage:   "Alternative to l2-genesis flag, if genesis-state is omitted. Path to block header at genesis",
+	EnvVars: prefixEnvVars("L2_GENESIS_HEADER"),
+}
```
geoknee, line 112: I don't know if we should do anything about this now, but I think if the command is invoked with this flag most likely the chain would fail the check-genesis command?

add-chain/e2e_test.go
```diff
@@ -131,6 +140,23 @@ func TestAddChain_CheckRollupConfig(t *testing.T) {
 	}
 }
 
+func TestAddChain_CheckGenesis(t *testing.T) {
+	t.Run("genesis_zorasep", func(t *testing.T) {
+		t.Parallel()
```
geoknee, line 145: Is this doing much for us, here?

add-chain/cmd/check-genesis.go
```diff
@@ -0,0 +1,54 @@
...
+		var localGenesis *core.Genesis
+		if err = json.Unmarshal(file, &localGenesis); err != nil {
+			return fmt.Errorf("failed to unmarshal local genesis.json into core.Genesis struct: %w", err)
+		}
+
+		chainId := ctx.Uint64(flags.ChainIdFlag.Name)
+		gethGenesis, err := core.LoadOPStackGenesis(chainId)
+		if err != nil {
+			return fmt.Errorf("failed to load genesis via op-geth: %w", err)
+		}
```
geoknee, line 36: Should we put a comment in, and/or a message if we get an error here, about requiring the chain to have been added already to the registry before running this command (in a separate `go run` invocation / build)?

add-chain/e2e_test.go
```diff
@@ -93,15 +102,15 @@ func TestAddChain_Main(t *testing.T) {
 
 	t.Run("compress-genesis", func(t *testing.T) {
 		// Must run this test to produce the .json.gz output artifact for the
-		// subsequent CheckGenesisConfig test
+		// subsequent CheckGenesis test
```
geoknee, line 105: Would this be more accurate?
```suggestion
		// subsequent TestAddChain_CheckGenesis invocation. 
```

geoknee (approved):

add-chain/cmd/check-genesis.go
```diff
@@ -0,0 +1,54 @@
...
+
+		// Exceptions for testing
+		if localGenesis.Config.Optimism != nil {
+			// this value is always hardcoded in op-geth to 250. Need to updated op-geth to
+			// conditionally set the value (if canyon is activated) to remove the exception
```
geoknee, line 41: TODO I'll file an issue and we can update this comment to point at it.

geoknee: https://github.com/ethereum-optimism/op-geth/issues/346

add-chain/e2e_test.go
bitwiseguy: No not at the moment. Will remove

add-chain/flags/flags.go
bitwiseguy: That portion of the code has not been tested. It is code that was copied over from the `optimism/op-chain-ops/registry-data`. Do we still want to support this alternative to to the `genesis.json` input? If so, do we have an example of what one of those inputs files would look like?

add-chain/cmd/check-genesis.go
bitwiseguy: Added more detailed error message

add-chain/flags/flags.go
geoknee: I don't know. Let's file an issue and follow up with @protolambda about whether we could streamline this / remove that code path.

PR #355, opened 2024-07-03
add-chain: add compress-genesis subcommand

bitwiseguy:
Moves the monorepo `op-chain-ops/cmd/registry-data` logic into the `add-chain compress-genesis` subcommand. This helps unify the location of the commands chain operators need to use in order to add a chain to the `superchain-registry`.

### DIFF
--- a/add-chain/cmd/compress-genesis.go
+++ b/add-chain/cmd/compress-genesis.go
@@ -1,0 +1,255 @@
+package cmd
+
+import (
+	"bytes"
+	"compress/flate"
+	"compress/gzip"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+
+	"github.com/ethereum-optimism/optimism/op-service/jsonutil"
+	"github.com/ethereum-optimism/superchain-registry/add-chain/flags"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/core"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/urfave/cli/v2"
+)
+
+var CompressGenesisCmd = cli.Command{
+	Name: "compress-genesis",
+	Flags: []cli.Flag{
+		flags.L2GenesisFlag,
+		flags.L2GenesisHeaderFlag,
+		flags.ChainShortNameFlag,
+		flags.SuperchainTargetFlag,
+	},
+	Usage: "Generate a single gzipped data file from a bytecode hex string",
+	Action: func(ctx *cli.Context) error {
+		// Get the current script filepath
+		_, thisFile, _, ok := runtime.Caller(0)
+		if !ok {
+			panic("error getting current filepath")
+		}
+		superchainRepoRoot := filepath.Dir(filepath.Dir(filepath.Dir(thisFile)))
+		superchainTarget := ctx.String(flags.SuperchainTargetFlag.Name)
+		if superchainTarget == "" {
+			return fmt.Errorf("must set superchain-target (SCR_SUPERCHAIN_TARGET)")
+		}
+		chainShortName := ctx.String(flags.ChainShortNameFlag.Name)
+		if chainShortName == "" {
+			return fmt.Errorf("must set chain-short-name (SCR_CHAIN_SHORT_NAME)")
+		}
+
+		zipOutputDir := filepath.Join(superchainRepoRoot, "/superchain/extra/genesis", superchainTarget, chainShortName+".json.gz")
+		genesisPath := ctx.Path(flags.L2GenesisFlag.Name)
+		if genesisPath == "" {
+			// When the genesis-state is too large, or not meant to be available, only the header data is made available.
+			// This allows the user to verify the header-chain starting from genesis, and state-sync the latest state,
+			// skipping the historical state.
+			// Archive nodes that depend on this historical state should instantiate the chain from a full genesis dump
+			// with allocation data, or datadir.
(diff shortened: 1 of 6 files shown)


add-chain/e2e_test.go
```diff
@@ -90,6 +90,24 @@ func TestAddChain_Main(t *testing.T) {
 			compareJsonFiles(t, "superchain/extra/genesis-system-configs/sepolia/", tt.name, tt.chainShortName)
 		})
 	}
+
+	t.Run("compress-genesis", func(t *testing.T) {
```
bitwiseguy, line 94: This test runs as part of the `TestAddChain_Main` because its outputs are used as inputs to the `add-chain check-genesis-config` command/test that is added in https://github.com/ethereum-optimism/superchain-registry/pull/354

add-chain/cmd/compress-genesis.go
```diff
@@ -0,0 +1,255 @@
...
+}
+
+type GenesisAccount struct {
+	CodeHash common.Hash                                          `json:"codeHash,omitempty"`
+	Storage  jsonutil.LazySortedJsonMap[common.Hash, common.Hash] `json:"storage,omitempty"`
+	Balance  *hexutil.Big                                         `json:"balance,omitempty"`
+	Nonce    uint64                                               `json:"nonce,omitempty"`
+}
+
+type Genesis struct {
```
bitwiseguy, line 202: Can we get rid of this struct and use the existing `core.Genesis` one instead? I'm wondering if this was necessary to avoid a circular dependency when the code lived inside the monorepo but maybe is no longer needed

add-chain/cmd/compress-genesis.go
```diff
@@ -0,0 +1,255 @@
...
+			panic("error getting current filepath")
+		}
+		superchainRepoRoot := filepath.Dir(filepath.Dir(filepath.Dir(thisFile)))
+		superchainTarget := ctx.String(flags.SuperchainTargetFlag.Name)
+		if superchainTarget == "" {
+			return fmt.Errorf("must set superchain-target (SCR_SUPERCHAIN_TARGET)")
+		}
+		chainShortName := ctx.String(flags.ChainShortNameFlag.Name)
+		if chainShortName == "" {
+			return fmt.Errorf("must set chain-short-name (SCR_CHAIN_SHORT_NAME)")
```
geoknee, line 46: and here

add-chain/cmd/compress-genesis.go
```diff
@@ -0,0 +1,255 @@
...
+	"github.com/ethereum-optimism/superchain-registry/add-chain/flags"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/core"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/urfave/cli/v2"
+)
+
+var CompressGenesisCmd = cli.Command{
```
geoknee, line 24: Do you think we need a separate sub command for this, or could it just run as a part of the main command?

add-chain/cmd/compress-genesis.go
geoknee: I don't think there was a circular dependency. 

I think this is a different type on purpose because it represents the _minified_ version of the genesis, where code is replaced with codehash or storage is replaced with storage hash. https://github.com/ethereum-optimism/superchain-registry/blob/852e9cfe836ecff7bbe3a5d771b67acb20558391/README.md#L79-L83

geoknee: Perhaps you meant the type in the `superchain` module?

https://github.com/ethereum-optimism/superchain-registry/blob/52c557aa401e6aac03f528adb5090e4b61ccdc00/superchain/superchain.go#L761

We might be able to deduplicate against this one, although there might be something about ordering fields that we lose by doing that.

add-chain/cmd/compress-genesis.go
bitwiseguy: This could run as part of the main command, it was more of a personal preference to use a sub command. However, for the current e2e tests we have several rollup.json examples and only one genesis.json. Its nice for testing to avoid having multiple of the ~8MB genesis.json files stored in the repo. Maybe if we add `compress-genesis` command to the main command we add a flag for `--skip-compress-genesis=true` for flexibility?

I also think its nice to not overload the main command with more responsibility. But I'm open to putting them together if you feel like it leads to a better user experience.

bitwiseguy:
> Shall we prepare the monorepo PR to remove

Monorepo PR created here: https://github.com/ethereum-optimism/optimism/pull/11075

add-chain/cmd/compress-genesis.go
bitwiseguy: Yea the superchain `Genesis` struct seems like the one we could deduplicate against. All of the fields seem identical except:
* superchain package reimplements its own types because it can't import go-ethereum (to avoid circular dependency)
  * `common.Hash` --> `Hash`
  * `common.Address` --> `Address`
  * `hexutil.Big` --> `HexBig`
  * `Alloc` is a `LazySortedJsonMap` vs a vanilla map. Not sure what the reasoning is for sorting the json map. Not sure its worth blocking this PR on it but maybe worth tracking down and optimizing by remove one of the redundant structs later?

geoknee: Perhaps the sorting helps with comparing the JSON files in a standard way? Not sure. 

Agree we shouldn't block this PR on it and we can return to it later.

add-chain/cmd/compress-genesis.go
geoknee: I don't feel strongly, sounds like there are some good reasons for keeping it as a subcommand. If we needed to add a flag when combining the commands, we aren't really simplifying overall.

PR #357, opened 2024-07-03
Centralize all validation exclusions

geoknee:
Should make them easier to audit. 

TODO: 
- [x] centralize remaining exclusions, may need to break some tests up

### DIFF
--- a/validation/exclusions_test.go
+++ b/validation/exclusions_test.go
@@ -64,4 +64,5 @@ var exclusions = map[string]map[uint64]bool{
 	"Uniqueness_Check": {
 		11155421: true, // oplabs devnet 0, not in upstream repo
 		11763072: true, // base devnet 0, not in upstream repo}
-	}}
+	},
+}
--- a/validation/exclusions_test.go
+++ b/validation/exclusions_test.go
@@ -56,4 +56,8 @@ var exclusions = map[string]map[uint64]bool{
 		8453:     true, // mainnet/base
 		11155421: true, // sepolia-dev-0/oplabs-devnet-0
 		11763072: true, // sepolia-dev-0/base-devnet-0
+	},
+	"Uniqueness_Check": {
+		11155421: true, // oplabs devnet 0, not in upstream repo
+		11763072: true, // base devnet 0, not in upstream repo}
 	}}
--- a/validation/exclusions_test.go
+++ b/validation/exclusions_test.go
@@ -1,0 +1,59 @@
+package validation
+
+import (
+	"regexp"
+	"testing"
+)
+
+func skipIfExcluded(t *testing.T, chainID uint64) {
+	for pattern := range exclusions {
+		matches, err := regexp.Match(pattern, []byte(t.Name()))
+		if err != nil {
+			panic(err)
+		}
+		if matches && exclusions[pattern][chainID] {
+			t.Skip("Excluded!")
+		}
+	}
+}
+
+var exclusions = map[string]map[uint64]bool{
+	"ChainID_RPC_Check": {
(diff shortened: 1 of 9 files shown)


validation/exclusions_test.go
```diff
@@ -0,0 +1,68 @@
...
+		11155421: true, // sepolia-dev-0/oplabs-devnet-0   (no public endpoint)
+		11763072: true, // sepolia-dev-0/base-devnet-0     (no public endpoint)
+	},
+	"L2OO_Params": {
+		10:        true, // OP mainnet - Upgraded to fault proofs
+		999999999: true, // sepolia/zora  Incorrect finalizationPeriodSeconds, 604800 is not within bounds [12 12]
+		1740:      true, // sepolia/metal Incorrect finalizationPeriodSeconds, 604800 is not within bounds [12 12]
+		919:       true, // sepolia/mode  Incorrect finalizationPeriodSeconds, 180 is not within bounds [12 12]
+		11155420:  true, // sepolia/op No L2OO because this chain uses Fault Proofs https://github.com/ethereum-optimism/superchain-registry/issues/219
+		11155421:  true, // oplabs-sepolia-devnet-0 No L2OO because this chain uses Fault Proofs https://github.com/ethereum-optimism/superchain-registry/issues/219
```
geoknee, line 35: Now these exclusions are in one place, it is easier to see how we can have fewer of them. If the standard config requires at least fault proofs, we ought to be able to drop this test entirely (and a bunch of other code besides).

validation/exclusions_test.go
```diff
@@ -0,0 +1,68 @@
...
+		84532:     true, // base-sepolia (incorrect challenger)
+		7777777:   true, // zora (incorrect challenger)
+		1750:      true, // metal (incorrect challenger)
+		919:       true, // mode sepolia (incorrect challenger)
+		999999999: true, // zora sepolia (incorrect challenger)
+		34443:     true, // mode (incorrect challenger)
+	},
+	"L2_Security_Config": {
+		11155421: true, // sepolia-dev-0/oplabs-devnet-0   No Public RPC declared
+		11763072: true, // sepolia-dev-0/base-devnet-0     No Public RPC declared
```
geoknee, line 50: We probably want a neater way of handling devnets, as these come up often in this exclusion list.

validation/validation_test.go
```diff
@@ -63,15 +63,15 @@ func testUniversal(t *testing.T, chain *ChainConfig) {
...
-		testDataAvailability(t, chain)
-		testResourceConfig(t, chain)
-		testL2OOParams(t, chain)
-		testGasLimit(t, chain)
-		testGasPriceOracleParams(t, chain)
+		t.Run("Data Availability", func(t *testing.T) { testDataAvailability(t, chain) })
+		t.Run("Resource Config", func(t *testing.T) { testResourceConfig(t, chain) })
+		t.Run("L2OO Params", func(t *testing.T) { testL2OOParams(t, chain) })
+		t.Run("Gas Limit", func(t *testing.T) { testGasLimit(t, chain) })
+		t.Run("GPO Params", func(t *testing.T) { testGasPriceOracleParams(t, chain) })
```
bitwiseguy, line 70: What is the advantage of wrapping these test functions with the `t.Run()` method?

bitwiseguy (approved):
Looks good. Just left one question

validation/validation_test.go
geoknee: This is the only way we can exclude a chain from some of these tests but not others. With the `.Run()` syntax, the test is executed as a _named_ sub test, which we can target in the `exclusions_test.go` file.

PR #358, opened 2024-07-08
validation,add-chain: Update dependencies on op-geth and optimism (monorepo)

### LINKED ISSUE #280: validation: remove dependency on `op-bindings`

geoknee:
Updates modules to point at latest op-geth and optimism (monorepo) modules. This is important for writing checks which validate against the latest data structures / helpers from those repositories. 

Some necessary changes to get these updates in:
* copy paste some bindings from the old version of `optimism` into this repo (op-bindings no longer exists in the monorepo)
* fix a bug to do with plasma config marshalling. 

Closes #280


superchain/superchain.go
```diff
@@ -127,7 +127,7 @@ func (c ChainConfig) Identifier() string {
 }
 
 type PlasmaConfig struct {
-	DAChallengeAddress *Address `json:"da_challenge_contract_address" yaml:"-"`
+	DAChallengeAddress *Address `json:"da_challenge_contract_address" yaml:"da_challenge_contract_address"`
```
geoknee, line 130: From the POV of the downstream packages, this is a required field if `PlasmaConfig` is not nil (in fact I was getting a nil pointer panic in op-node due to the bug of not storing this information). 

I believe we have some checks on this but they weren't sufficient to catch this bug.

PR #361, opened 2024-07-08
validation: add superchain config check

geoknee:
This serves as a check that the chain is actually (via an on chain call) in the superchain that it claims to be (via its config directory name).


refcell (approved):
LGTM nice add

PR #364, opened 2024-07-09
feat: Autogenerate Genesis System Config Aggregate

refcell:
**Description**

Autogenerates a json file at `extra/genesis-system-configs/genesisSystemConfigs.json` that aggregates all genesis system configs in subdirectories and maps chain id to the respective genesis system config. The final output looks something like the following:

```json
{
  "10": {
    "batcherAddr": "0x6887246668a3b87F54DeB3b94Ba47a6f63F32985",
    "overhead": "0x00000000000000000000000000000000000000000000000000000000000000bc",
    "scalar": "0x00000000000000000000000000000000000000000000000000000000000a6fe0",
    "gasLimit": 30000000
  },
  "11155420": {
    ...
  },
  ...
}
```

This allows genesis system configs to be consumed without file system directory traversal (needed for compile time `no_std` rust bindings). The chain id is found by comparing the filepath derived identifier to an identifier in the top-level `chainList.json` file. The binary can be run in the local dir path with `go run .`.


refcell:
We don't need this anymore since genesis system config is included in the superchain configs.

PR #367, opened 2024-07-09
Small tweaks to justfile

geoknee:


justfile
```diff
@@ -11,9 +11,10 @@ add-chain:
 # Promote a standard candidate chain to a standard chain, example: just promote-to-standard 10
 promote-to-standard CHAIN:
 	go run ./add-chain promote-to-standard --chain-id={{CHAIN}}
+	just codegen
 
 # Generate auto-generated files
-codegen:
+codegen: clean-add-chain
```
geoknee, line 17: This means `codegen` depends on `clean-add-chain`, so that will be executed first.

bitwiseguy (approved):
👍

PR #368, opened 2024-07-10
Reinsert metal sepolia and superlumio mainnet 

### LINKED ISSUE #287: SuperchainRegistry: Reinsert temporarily removed conduit chains to the registry
Potted context is as follows:

- When a chain is added to the superchain-registry, hardfork times are scraped from rollup.json.
- When the superchain-registry configs are pulled into op-node, the hardfork times need to be correct or else nodes which start up using the op-network will fork off.
- Incorrect hardfork times in the registry can also result in a validation check on the genesis block failing.
- We know that the hardfork times for mainnet/superlumio and sepolia/metal are incorrect because this check was failing
- Hardfork times can be queried from a running op-node https://docs.optimism.io/builders/node-operators/json-rpc#optimism_rollupconfig

From what I can gather from when conduit make this query it returns hardfork times which still appear to be incorrect.

https://github.com/ethereum-optimism/superchain-registry/pull/243

geoknee:
Replaces #250 

Closes #287


bindings/rust-bindings/src/lib.rs
```diff
@@ -54,7 +54,7 @@ mod tests {
     fn test_read_chain_configs() {
         let mainnet = SUPERCHAINS.get("mainnet").unwrap();
         assert_eq!(mainnet.config.name, "Mainnet");
-        assert_eq!(mainnet.chain_ids.len(), 8);
+        assert_eq!(mainnet.chain_ids.len(), 9);
```
geoknee, line 57: @refcell I think this is going to trip us up every time a new mainnet chain is added. Can we remove this assertion or replace it with something different?

superchain/configs/sepolia/metal.yaml
```diff
@@ -0,0 +1,25 @@
+name: Metal L2 Testnet
+chain_id: 1740
+public_rpc: https://testnet.rpc.metall2.com
+sequencer_rpc: https://testnet.rpc.metall2.com
+explorer: https://testnet.explorer.metall2.com
+
+superchain_level: 1
+standard_chain_candidate: true # This is a temporary field which causes most of the standard validation checks to run on this chain
+superchain_time: null # Missing hardfork times are NOT yet inherited from superchain.yaml
+contracts_version_tag: "op-contracts/v1.3.0" # Multi-Chain Prep (MCP) https://github.com/ethereum-optimism/optimism/releases/tag/op-contracts%2Fv1.3.0
```
geoknee, line 10: From @bitwiseguy : Do we need to harmonize the quoting of strings in this file with what we are doing in other yaml files and in our check-rollup-config command?

refcell (approved):

Issue #372, opened 2024-07-10
validtaion: check genesis predeploys against standard config


geoknee:
Predeploy spec: https://specs.optimism.io/protocol/predeploys.html

[Standard config spec states ](https://specs.optimism.io/protocol/configurability.html?highlight=op%20stack#op-stack-configurability)

> [Genesis state should have] Only standard predeploys and preinstalls, no additional state.

geoknee:
https://github.com/ethereum-optimism/superchain-registry/pull/373 is a sketch at a solution for this.

geoknee:
Q: with interoperable OP is it problematic if a chain enables the governance token on their chain?
A1: This OP token thing — there should be nothing deployed I think
A2: (Possibly) EAS predeploys — idk if chains other than OP Mainnet and Base have them. if they do, great, if not, I don’t think we need to require,

geoknee:
Actually here is the source, it is also in the specs https://github.com/ethereum-optimism/optimism/blob/d510910381d99660295f99e4eccd553df5829df6/op-e2e/actions/fjord_fork_test.go#L24-L29

geoknee:
Here are some results investigating the distribution of codehashes across the superchain registry. (For each predeploy implementation address, it shows a number of codehashes, each with a list of chains which use that implementation). 

https://github.com/ethereum-optimism/superchain-registry/blob/3ff4a2d6e69d248c36a9a1fee8a9b7df461d7710/validation/results.toml

BlocksOnAChain:
We should check with @blmalone about his approach for this check.

bitwiseguy:
We are planning to leverage @blmalone script. The work for this ticket will be to make sure the script is fully automated and integrate it into the superchain-registry CI

geoknee:
Latest plan here https://www.notion.so/oplabs/L2-Genesis-Predeploys-Verification-c1c6f0b5c38d406bbfe09854653dade4?pvs=4#fbe8b779b04d46b1b7d5b80f9c6f506e leaning on runbook here https://www.notion.so/oplabs/BHIC-L2-Genesis-Predeploy-Verification-Metal-Mode-Zora-f42ecf4dd0164b4d9303b933d9e98bf6#0a94fd37ca37474d998ab1668e655264

geoknee:
[...]

PR #373, opened 2024-07-10
sketch out genesis predeploy check

geoknee:
Towards #372 

To get this over the line, I think the blockers are as follows: 

* We need a good "ground truth" for what the standard predeploys actually are. I would have thought OP Mainnet, but then:
```
    superchain-genesis_test.go:75: 
                Error Trace:    /Users/georgeknee/code/ethereum-optimism/superchain-registry/validation/superchain-genesis_test.go:75
                                                        /Users/georgeknee/code/ethereum-optimism/superchain-registry/validation/validation_test.go:80
                Error:          Should be true
                Test:           TestValidation/OP_Mainnet_(10)/Standard_or_Standard_Candidate_Chain/Standard_Genesis_Predeploys
                Messages:       Genesis does not contain an account for 0x4200000000000000000000000000000000000019
```
* We need to fill out all of the expectations into a TOML file, read in and loop over them in this test
* Exclude (`exclusions_test.go`) or downgrade any chain failing the check


geoknee:
For OP Mainnet, or in fact for any chain, if the genesis allocs are not present we have the option to go on chain (L2) and query for the information.

validation/superchain-genesis_test.go
```diff
@@ -53,3 +59,142 @@ func testGenesisHashAgainstRPC(t *testing.T, chain *ChainConfig) {
...
+				continue
+			}
+			if chain.ChainID == 10 {
+				continue
+			}
+
+			g, err := LoadGenesis(chain.ChainID)
+			require.NoError(t, err)
+
+			ch := g.Alloc[MustHexToAddress(implementation)].CodeHash
```
geoknee, line 132: @blmalone

PR #376, opened 2024-07-10
Use blockspace charter terminology

### LINKED ISSUE #295: Use blockspace charter terminology in place of "standard config" terminology
And link from the readme to any BC material.

https://gov.optimism.io/t/season-6-introducing-blockspace-charters-superchain-first-governance/8133

> The set of validation declaration files we have defines the first blockspace charter (standard rollup charter). These are [some TOML files at present,](https://github.com/ethereum-optimism/superchain-registry/blob/main/validation/standard/standard-config-mainnet.toml) and I think it would nice if we think of those as the specification of the charter from the validation point of view.

geoknee:
Closes #295

### DIFF
--- a/validation/standard/readme.md
+++ b/validation/standard/readme.md
@@ -1,7 +1,7 @@
 # Standard Rollup Blockspace Charter definition files
 This directory contains a number of TOML files which declare various parameters, addresses, contract versions and other data which define a **standard chain** in the sense of the [Standard Rollup Charter](https://gov.optimism.io/t/season-6-draft-standard-rollup-charter/8135).
 
-Where necessary, distinct declarations are made for testnets and devnets to the configuration to vary to a greater extent.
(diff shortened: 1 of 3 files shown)


validation/standard/readme.md
```diff
@@ -0,0 +1,9 @@
+# Standard Rollup Blockspace Charter definition files
+This directory contains a number of TOML files which declare various parameters, addresses, contract versions and other data which define a **standard chain** in the sense of the [Standard Rollup Charter](https://gov.optimism.io/t/season-6-draft-standard-rollup-charter/8135).
+
+Where necessary, distinct declarations are made for testnets and devnets to the configuration to vary to a greater extent.
```
bitwiseguy, line 4: This sentence is a bit confusing to me. Suggest re-wording to something like this:

Where necessary, distinct declarations are made to allow testnets and devnets to have greater flexibility in their configurations.

PR #385, opened 2024-07-11
Bump devnet chains down to Frontier, prevent exclusions for Standard chains. 

geoknee:
Two changes: 

1. Devnet chains need to be excluded from many checks, mainly because: they do not tend to be in the upstream chain list repo, they do not tend to have public rpc endpoints, and they are generally configured in a loose way. So I don't think it makes sense for them to be anything other than Frontier chains. Testnet chains can still be standard / standard candidate, this allows us to go through the motions on testnets as it will be for mainnet.

2. I want to keep the list of exclusions for legacy chains short and prevent it from getting stale. So I would like to put in a mechanism (not 100% foolproof but pretty good) to prevent us excluding any chain from a check once it is a standard chain. In practice, I think this will force us to trim the exclusion list (e.g. when chains would pass the check but are still excluded, we just unexclude them and happy days).


validation/superchain-genesis_test.go
```diff
@@ -16,6 +16,11 @@ import (
 
 func testGenesisHash(t *testing.T, chainID uint64) {
 	skipIfExcluded(t, chainID)
+
+	if chainID == 10 {
+		t.Skip("OP Mainnet has a pre-bedrock genesis, so we exclude it from this check.")
+	}
```
geoknee, line 22: I think I finally understand why OP Mainnet doesn't work. It might not be because it has a pre-bedrock genesis, but that the compressed genesis file does contains an empty `alloc: {}` property. 

Likely added using the "alternative" add-chain path 
https://github.com/ethereum-optimism/superchain-registry/blob/9702f3ce28c79d3a9f0ec6f262eb6c2f997298f0/add-chain/cmd/compress-genesis.go#L51-L94

validation/superchain-genesis_test.go
```diff
@@ -16,6 +16,11 @@ import (
 
 func testGenesisHash(t *testing.T, chainID uint64) {
 	skipIfExcluded(t, chainID)
+
+	if chainID == 10 {
+		t.Skip("OP Mainnet has a pre-bedrock genesis (with an empty allocs object stored in the registry), so we exclude it from this check.")
+	}
```
bitwiseguy, line 22: Why is this exclusion called out separately whereas the other exclusions are defined in the exclusions_test.go file?

bitwiseguy (approved):
Looks good. Approved with one open question

validation/superchain-genesis_test.go
geoknee: This is because it is a really unique exceptional case, and the new `TestExclusions` test requires that no standard chain is excluded from any check. I tidied this up a bit [4c73866](https://github.com/ethereum-optimism/superchain-registry/pull/385/commits/4c73866516003dd9fd1da6ee9dabd96a064754e7).

Issue #347, opened 2024-06-28
`promote-to-standard` has unexpected behaviour on non-standard-candidate chains


geoknee:
* It should be a no-op on standard chains.

* It should only modify the intended config file fields, it should not modify the hardfork times for instance.

geoknee:
I think this command would benefit from some e2e tests. I noticed today that it also adds hardfork times to the config file, which I don't think we want.

PR #386, opened 2024-07-11
Use toml instead of yaml for chain config files

bitwiseguy:
## Description
Instead of using a combination of toml, yaml, and json config files, this PR combines the old config yaml, addresses json, and genesis-system-config json into a single toml file. This should make it easier to maintain the`superchain` package bindings going forward. There are already rust and go bindings for that package, but it will also be easier to add other languages in the future.

There are no breaking changes for downstream packages (i.e. op-geth and op-node) introduced by this PR. Instead, the `superchain` package `init()` function has just been updated so that it can parse the single toml config file and then populate the same data structures that were exported before.

## Additional Context
Fixes https://github.com/ethereum-optimism/superchain-registry/issues/285


.circleci/config.yml
```diff
@@ -21,11 +21,13 @@ commands:
   install-just:
     description: "Install just"
     steps:
-      - run: |
-          wget -qO - 'https://proget.makedeb.org/debian-feeds/prebuilt-mpr.pub' | gpg --dearmor | sudo tee /usr/share/keyrings/prebuilt-mpr-archive-keyring.gpg 1> /dev/null
-          echo "deb [arch=all,$(dpkg --print-architecture) signed-by=/usr/share/keyrings/prebuilt-mpr-archive-keyring.gpg] https://proget.makedeb.org prebuilt-mpr $(lsb_release -cs)" | sudo tee /etc/apt/sources.list.d/prebuilt-mpr.list
-          sudo apt update
-          sudo apt install just
+      - run:
+          name: "Install just"
```
bitwiseguy, line 25: This just made the step title easier to read in circle ci (instead of the long wget command being the name)

add-chain/cmd/convert-to-toml.go
```diff
@@ -0,0 +1,41 @@
...
+	"path/filepath"
+	"runtime"
+
+	"github.com/ethereum-optimism/superchain-registry/add-chain/config"
+	"github.com/ethereum-optimism/superchain-registry/add-chain/flags"
+	"github.com/ethereum-optimism/superchain-registry/superchain"
+	"github.com/urfave/cli/v2"
+)
+
+var ConvertToTOMLCmd = cli.Command{
```
bitwiseguy, line 14: This is a helper command to convert existing chain config .yaml files to .toml. Can be removed before this pr is merged

.circleci/config.yml
geoknee: Good call

superchain/superchain.go
```diff
@@ -76,57 +77,59 @@ const (
...
+	SequencerRPC string `yaml:"sequencer_rpc" toml:"sequencer_rpc"`
+	Explorer     string `yaml:"explorer" toml:"explorer"`
 
-	SuperchainLevel SuperchainLevel `yaml:"superchain_level"`
+	SuperchainLevel SuperchainLevel `yaml:"superchain_level" toml:"superchain_level"`
 
 	// If StandardChainCandidate is true, standard chain validation checks will
 	// run on this chain even if it is a frontier chain.
-	StandardChainCandidate bool `yaml:"standard_chain_candidate,omitempty"`
+	StandardChainCandidate bool `yaml:"standard_chain_candidate,omitempty" toml:"standard_chain_candidate,omitempty"`
```
geoknee, line 90: Can we remove all yaml struct tags? Is that worth doing, or not? (Genuine Q)

geoknee (commented):
This is looking pretty good to me so far, left a few comments.

superchain/superchain.go
```diff
@@ -173,6 +176,45 @@ func (c *ChainConfig) setNilHardforkTimestampsToDefaultOrZero(s *SuperchainConfi
 	}
 }
 
+func (c *ChainConfig) EnhanceTOML(ctx context.Context) (map[string]string, error) {
```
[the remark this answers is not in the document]
bitwiseguy: Good call. Will update that

superchain/superchain_test.go
```diff
@@ -189,24 +190,23 @@ func testNetworkUpgradeTimestampOffset(l2GenesisTime uint64, blockTime uint64, u
...
+superchain_config_addr = "0x02d91Cf852423640d93920BE0CAdceC0E7A00FA7"
+
+canyon_time = 1
+delta_time = 2
+ecotone_time = 3
+
+[l1]
+  chain_id = 314
+  public_rpc = "https://disney.com"
+  explorer = "https://disneyscan.io"
```
bitwiseguy, line 205: toml's standard ordering puts all top level fields first, then nested fields follow. So this raw test data needed to be modified (along with the superchain.toml files) to adhere to that standard. Otherwise the fields are all missing when we unmarshal to a struct

superchain/superchain.go
bitwiseguy: Yes we can and should after we're confident the toml config files contain all of the data read from the yaml versions, and we no longer need the yaml tags to help with the one-time conversion (via `add-chain convert-to-toml` temp subcommand)

superchain/superchain_test.go
geoknee: Not sure I follow: does this example not also have "all top level fields first, then nested fields follow"?

bitwiseguy: This example was changed so that it has all top level fields first. Previously it looked like this:
```
name: Mickey Mouse
l1:
  chain_id: 314
  public_rpc: https://disney.com
  explorer: https://disneyscan.io

protocol_versions_addr: "0x252CbE9517F731C618961D890D534183822dcC8d"
superchain_config_addr: "0x02d91Cf852423640d93920BE0CAdceC0E7A00FA7"

canyon_time: 1
delta_time: 2
ecotone_time: 3
fjord_time:
```
My comment was just explaining the reason for the diff

geoknee: Gotcha! Makes sense, thanks.

add-chain/cmd/promote-to-standard.go
```diff
@@ -37,8 +37,8 @@ var PromoteToStandardCmd = cli.Command{
 
 		superchainRepoPath := filepath.Dir(filepath.Dir(filepath.Dir(thisFile)))
 		targetDir := filepath.Join(superchainRepoPath, "superchain", "configs", chain.Superchain)
-		targetFilePath := filepath.Join(targetDir, chain.Chain+".yaml")
-		err := config.WriteChainConfig(*chain, targetFilePath)
+		targetFilePath := filepath.Join(targetDir, chain.Chain+".toml")
+		err := config.WriteChainConfigTOML(*chain, targetFilePath)
```
geoknee, line 41: Not a blocker for this PR and out of scope, but I would like to fix this to work in a more lightweight way. As is, it pulls hardfork overrides in to the leaf configs.

superchain/init.go
```diff
@@ -89,9 +69,8 @@ func init() {
 			}
 			superchainEntry.ChainIDs = append(superchainEntry.ChainIDs, chainConfig.ChainID)
 			OPChains[chainConfig.ChainID] = &chainConfig
-			Addresses[chainConfig.ChainID] = &addrs
-			GenesisSystemConfigs[chainConfig.ChainID] = &genesisSysCfg
-
+			Addresses[chainConfig.ChainID] = &chainConfig.Addresses
+			GenesisSystemConfigs[chainConfig.ChainID] = &chainConfig.Genesis.SystemConfig
```
geoknee, line 73: We could file an issue describing an obvious cleanup to remove these mappings (breaking change).

superchain/superchain.go
```diff
@@ -76,55 +78,57 @@ const (
...
+
+	Genesis ChainGenesis `yaml:"genesis" toml:"genesis"`
 
-	BlockTime           uint64 `yaml:"block_time"`
-	SequencerWindowSize uint64 `yaml:"seq_window_size"`
+	Addresses AddressList `toml:"addresses"`
 
 	// Optional feature
-	Plasma *PlasmaConfig `yaml:"plasma,omitempty"`
+	Plasma *PlasmaConfig `yaml:"plasma,omitempty" toml:"plasma,omitempty"`
```
geoknee, line 117: See other comment. I think we should place this above `Genesis`.

superchain/superchain.go
```diff
@@ -76,55 +78,57 @@ const (
...
+	HardForkConfiguration `yaml:",inline" toml:",inline"`
+
+	BlockTime           uint64 `yaml:"block_time" toml:"block_time"`
+	SequencerWindowSize uint64 `yaml:"seq_window_size" toml:"seq_window_size"`
+
+	Genesis ChainGenesis `yaml:"genesis" toml:"genesis"`
 
-	BlockTime           uint64 `yaml:"block_time"`
-	SequencerWindowSize uint64 `yaml:"seq_window_size"`
+	Addresses AddressList `toml:"addresses"`
```
geoknee, line 114: I suppose this data is now available both at the existing top level `Addresses` mapping _and_ at `config.Addresses`. We could probably "hide it" by not exporting it, but this would require a bit of extra work to get it to unmarshal properly. 

I think it could make sense though, to use the single unified input TOML file but umarshal it three times. Once into an unmodified `ChainConfig` (i.e. it does not have Addresses field), and then into the top level `Addresses` field and again into the top level `GenesisSystemConfigs` field. If we did that, we wouldn't be changing the API of the Go bindings at all.

I don't feel strongly though, I think longer term a single mapping for a single set of TOML files makes the most sense, and we should just go ahead and make the breaking change and tidy it up when we find the time. For now, this is probably an acceptable bit of duplication.

geoknee (commented):
Posting review comments so far, I need to take a closer look at the custom marshalling code next. Generally it appears to work and I think we're close to having this mergeable. Main outstanding thing is the questions around the `mapToAddressList` fn.

add-chain/cmd/convert-to-toml.go
bitwiseguy: Done here: https://github.com/ethereum-optimism/superchain-registry/pull/386/commits/347a835f491592b51b80bd6db63ec74a2554dfce

superchain/superchain.go
bitwiseguy: https://github.com/ethereum-optimism/superchain-registry/pull/386/commits/1f362c8813018ed8f50efdd0017a08c91d38bb22

add-chain/testdata/superchain/configs/sepolia/expected_plasma.toml
```diff
@@ -0,0 +1,53 @@
...
+  ProxyAdminOwner = "0x9663FF52F1666954551460FcF190b4E3DaECab37"
+  SuperchainConfig = "0xdf69B06cEB4A4b4Db3CC288840435e673ac4d893"
+  SystemConfigOwner = "0xe8D525b30dD4C406570d47396b5b237DB30F13F1"
+  SystemConfigProxy = "0x8787D3382F60d4e072b4276BBda065959195dB33"
+  UnsafeBlockSigner = "0x9431B143448A60A374ab196d0d7C0683c5369006"
+
+[plasma]
+  da_challenge_contract_address = "0x3333333333333333333300000000000000000000"
+  da_challenge_window = 5555555
+  da_resolve_window = 7777777
```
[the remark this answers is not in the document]
bitwiseguy: https://github.com/ethereum-optimism/superchain-registry/pull/386/commits/1f362c8813018ed8f50efdd0017a08c91d38bb22

add-chain/main.go
```diff
@@ -38,6 +38,7 @@ var app = &cli.App{
 		&cmd.CheckRollupConfigCmd,
 		&cmd.CompressGenesisCmd,
 		&cmd.CheckGenesisCmd,
+		&cmd.ConvertToTOMLCmd,
```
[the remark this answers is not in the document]
bitwiseguy: https://github.com/ethereum-optimism/superchain-registry/pull/386/commits/347a835f491592b51b80bd6db63ec74a2554dfce

superchain/superchain.go
```diff
@@ -171,105 +172,170 @@ func (c *ChainConfig) setNilHardforkTimestampsToDefaultOrZero(s *SuperchainConfi
...
-	// Check if context is done before processing
-	if err := ctx.Err(); err != nil {
-		return fmt.Errorf("context error: %w", err)
+func (c ChainConfig) MarshalTOML() ([]byte, error) {
+	// Uses []outField to deterministically set the order of the toml based on the order of fields
+	// in the ChainConfig struct. Otherwise the fields are ordered alphabetically
+	type outField struct {
+		key   string
+		value interface{}
 	}
```
geoknee, line 181: I understand that we can run into problems here if the `ChainConfig` struct ordering has unnested fields listed after nested fields, since the resulting file will have an incorrect information hierarchy.

Can we detect that inside this function and error or panic with a clear warning about how to fix this?

superchain/superchain.go
```diff
@@ -171,105 +172,170 @@ func (c *ChainConfig) setNilHardforkTimestampsToDefaultOrZero(s *SuperchainConfi
...
+			}
+		} else if fieldName == "Addresses" {
+			nested, err := field.Interface().(AddressList).MarshalTOML()
+			if err != nil {
+				return nil, err
+			}
+			nestedMap := make(map[string]interface{})
+			if err := toml.Unmarshal(nested, &nestedMap); err != nil {
+				return nil, err
+			}
```
geoknee, line 228: Why is this necessary (marshalling and then unmarshalling into a different type?

superchain/configs/sepolia/superchain.toml
```diff
@@ -0,0 +1,13 @@
+name = "Sepolia"
+protocol_versions_addr = "0x79ADD5713B383DAa0a138d3C4780C7A1804a8090"
+superchain_config_addr = "0xC2Be75506d5724086DEB7245bd260Cc9753911Be"
+
+canyon_time =  1699981200 # Tue 14 Nov 2023 17:00:00 UTC
+delta_time =   1703203200 # Fri 22 Dec 2023 00:00:00 UTC
+ecotone_time = 1708534800 # Wed 21 Feb 2024 17:00:00 UTC
+fjord_time =   1716998400 # Wed May 29 16:00:00 UTC 2024
```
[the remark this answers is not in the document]
bitwiseguy: Fixed

superchain/superchain.go
```diff
@@ -252,36 +294,36 @@ func (c *ChainConfig) EnhanceYAML(ctx context.Context, node *yaml.Node) error {
...
-	Challenger        Address `json:"Challenger"`
-	Proposer          Address `json:"Proposer"`
-	UnsafeBlockSigner Address `json:"UnsafeBlockSigner"`
-	BatchSubmitter    Address `json:"BatchSubmitter"`
+	SystemConfigOwner Address `json:"SystemConfigOwner" toml:"SystemConfigOwner"`
+	ProxyAdminOwner   Address `json:"ProxyAdminOwner" toml:"ProxyAdminOwner"`
+	Guardian          Address `json:"Guardian" toml:"Guardian"`
+	Challenger        Address `json:"Challenger" toml:"Challenger"`
+	Proposer          Address `json:"Proposer" toml:"Proposer"`
+	UnsafeBlockSigner Address `json:"UnsafeBlockSigner" toml:"UnsafeBlockSigner"`
```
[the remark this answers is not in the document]
bitwiseguy: Yes but will require more changes to downstream code that imports these fields and is currently tightly coupled to the key names

superchain/init.go
bitwiseguy: Filed issue here: https://github.com/ethereum-optimism/superchain-registry/issues/391

superchain/superchain.go
bitwiseguy: Agreed that the breaking change will be the best solution. Issue filed here to capture that work: https://github.com/ethereum-optimism/superchain-registry/issues/391

geoknee:
The `codegen` functionality / check will break when we delete the extra addresses json files. 

We should migrate `codegen.js` 

https://github.com/ethereum-optimism/superchain-registry/blob/5d2f856eb73fc6709489a28f21cf74475b6ba1af/scripts/codegen.js#L7-L11

to be performed by the existing Go codegen tool: 
https://github.com/ethereum-optimism/superchain-registry/blob/5d2f856eb73fc6709489a28f21cf74475b6ba1af/superchain/internal/codegen/main.go#L33-L34

I think we have all of the utility code we need to do that. 

Finally, we would trim the `node codegen.js` line from the justfile.

add-chain/cmd/promote-to-standard.go
geoknee: https://github.com/ethereum-optimism/superchain-registry/issues/347

superchain/superchain.go
bitwiseguy: I needed to create a generic `map[string]interface{}` version of the `AddressList` in order to append it to the `out []outField` slice, which is what is used to preserve the order of fields and the ensure correct formatting. Without this whole `else if fieldName == "Addresses"` block the formatting of the output toml file looks like this. Notice that `addresses` is not formatted as a proper toml tag:
```
        	            	+addresses = AddressManager = "0x3eb579b25F6b9547e0073c848389a768FD382296"
        	            	+AnchorStateRegistryProxy = "0x03b82AE60989863BCEb0BbD442A70568e5AefB85"
        	            	+BatchSubmitter = "0x19CC7073150D9f5888f09E0e9016d2a39667df14"
        	            	+Challenger = "0x8c20c40180751d93E939DDDee3517AE0d1EBeAd2"
```

superchain/superchain.go
```diff
@@ -24,7 +24,7 @@ var ErrEmptyVersion = errors.New("empty version")
 //go:embed configs
 var superchainFS embed.FS
 
-//go:embed extra/addresses extra/bytecodes extra/genesis extra/genesis-system-configs
+//go:embed extra/addresses extra/bytecodes extra/genesis
```
geoknee, line 27: Can we also drop `extra/addresses`?

bitwiseguy: https://github.com/ethereum-optimism/superchain-registry/pull/386/commits/2a13885ea768ac6d1c2a8b9ed66afcaed9159fa3

superchain/superchain.go
geoknee: OK. Shall we add a comment explaining this, at least?

superchain/superchain.go
bitwiseguy: Added code to detect this situation. Tested manually by moving some fields around. One tricky part is ignoring `inline` structs. https://github.com/ethereum-optimism/superchain-registry/pull/386/commits/fe0700ae674b787f3ece802bbc223edd71905be1

superchain/superchain.go
bitwiseguy: https://github.com/ethereum-optimism/superchain-registry/pull/386/commits/d00a3b0588294f9711f7c75c3b913338225a4c1a

refcell (approved):
Rust code GTG. go LGTM.

add-chain/testdata/superchain/configs/sepolia/expected_faultproofs.toml
```diff
@@ -0,0 +1,51 @@
+name = "testchain_faultproofs"
```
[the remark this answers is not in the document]
bitwiseguy: The fault proofs have additional contracts in the `[addresses]` field of the config so its checking that those are properly added

add-chain/main.go
```diff
@@ -135,8 +136,24 @@ func entrypoint(ctx *cli.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to read addresses from chain: %w", err)
+	}
+
+	if rollupConfig.Plasma != nil {
+		addresses["DAChallengeAddress"] = rollupConfig.Plasma.DAChallengeAddress.String()
+	}
+
+	addressList := superchain.AddressList{SystemConfigProxy: superchain.MustHexToAddress("0x0fe884546476dDd290eC46318785046ef68a0BA9")}
+	err = mapToAddressList(addresses, &addressList)
```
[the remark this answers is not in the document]
bitwiseguy: That was a leftover bit of test code. That value was actually being overwritten so it was inconsequential, but it is now removed as part of cleanup

geoknee (approved):

PR #389, opened 2024-07-12
make almost all env vars mandatory

geoknee:
and replace stubs with comments


geoknee:
@bitwiseguy would appreciate your help debugging CI failures on this branch. Any clues?

add-chain/flags/flags.go
```diff
@@ -20,42 +20,42 @@ var (
...
-		Required: false,
+		Required: true,
 	}
 	ChainShortNameFlag = &cli.StringFlag{
 		Name:     "chain-short-name",
 		Value:    "",
 		EnvVars:  prefixEnvVars("CHAIN_SHORT_NAME"),
 		Usage:    "Custom short name of the chain",
-		Required: false,
+		Required: true,
```
bitwiseguy, line 30: I don't think we should make these flags required here. As the code is currently written we have a top level entrypoint command that has these flags:
```
		Flags: []cli.Flag{
			flags.PublicRpcFlag,
			flags.SequencerRpcFlag,
			flags.ExplorerFlag,
			flags.SuperchainTargetFlag,
			flags.MonorepoDirFlag,
			flags.ChainNameFlag,
			flags.ChainShortNameFlag,
			flags.RollupConfigFlag,
			flags.DeploymentsDirFlag,
			flags.StandardChainCandidateFlag,
		},
```
If they are set to `Required`, that means they will be required for all subcommands too, which is not what we want since some of those are irrelevant to subcommands. Instead, we can rely on logic within the command functions themselves to determine what is required for that specific command. The way we are loading env vars from the .env file also complicates things due to the timing of when the env vars are set vs when the cli app attempts to validate all of the required flags

bitwiseguy: This is the source of the CI failure you are seeing on this branch

bitwiseguy: Yes I like that refactor idea. However I think there are still some instances where we'd want the same flag to be required on one subcommand but optional on another

geoknee: Yeah that's ok, we could possibly solve that with some copy+mutate or inheritance type thing on the flags. I can try and cook something up.

geoknee:
This is stale now.

PR #394, opened 2024-07-16
fix(bindings): Parse TOML Chain Configs

refcell:
**Description**

Fixes bindings to parse the `toml` chain config files.


bitwiseguy:
@refcell would you mind pulling in the latest commits from the `ss/toml-over-yaml` branch? The latest commit on that branch removed all of the `superchain/extras/addresses` and `superchain/extras/genesis-system-config` files. All of that data should now be pulled from the config .toml files. I just want to confirm the rust code is getting its data from the right spot

clabby (approved):
LGTM! Thanks for fixing.

Issue #371, opened 2024-07-10
validation: add drift detection on `SystemConfig.startBlock()`


geoknee:
We want to
* store the _original_ `startBlock` value in the superchain registry somewhere (this could be part of the add-chain tooling)
  * get the `Initialized` logs from the system config contract, and use the oldest block number
* have a validation check which queries the current value and ensures it is equal to the original value.

geoknee:
No mention of this in the BHIC spec. 

This is an example of something that can change in principle in future, so might be something (therefore) to check in the registry. 

However, if we can be sure that this parameter cannot change after key handover, does that mean we don't need to check it going forward?

geoknee:
There's a general question about how much we should be relying on BHIC check + key handover to ensure a chain is still "standard".

geoknee:
From my investigations, it seems that the l1 block time stored in the `rollup.json` file (and therefore read by `op-node`) is decided by when the getting-started script is run https://github.com/ethereum-optimism/optimism/blob/d510910381d99660295f99e4eccd553df5829df6/packages/contracts-bedrock/scripts/getting-started/config.sh#L26-L29, or even when the op-node genesis creation command is run: https://github.com/ethereum-optimism/optimism/blob/547ea72d9849e13ce169fd31df0f9197651b3f86/op-node/cmd/genesis/cmd.go#L189-L190.

geoknee:
The system config value [is documented as](https://github.com/ethereum-optimism/optimism/blob/d510910381d99660295f99e4eccd553df5829df6/packages/contracts-bedrock/src/L1/SystemConfig.sol#L77-L79)

_    /// @notice Storage slot for block at which the op-node can start searching for logs from._.

geoknee:
* The check in the registry should be that the off-chain start block value is earlier than or equal to the on chain value.
* Chains that have the off-chain value later than the on chain value will fail the test. They then need to manually alter their config (i.e. the rollup.json) to push it back earlier and resubmit the PR
* We probably need a bit more work to get drift detection in place for the on-chain value, we need to understand how the start block has changed on chain over time.
* Existing chains which fail will need to be excluded, or we communicate with them about updating their offchain config values.

bitwiseguy:
Existing chains that fail the offChain <= onChain start block check:
* OP mainnet
  * offChain: [17_422_590](https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/configs/mainnet/op.toml#L21)
[...]

PR #398, opened 2024-07-18
validation: start block RPC check

bitwiseguy:
## Description

Adds validation check to compare on-chain `SystemConfig.startBlock()` value with the one stored in the chain config file. They should match, else the validation for that chain fails.

### Additional context

Fixes https://github.com/ethereum-optimism/superchain-registry/issues/371


validation/start-block_test.go
sebastianst, line 1: More consistent to name this file `start_block_test.go`, as to not mix `-` and `_` in Go filenames.

validation/start-block_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	return val.Uint64(), nil
+}
+
+// checkForDepositEvents looks in the blocks between startBlock and endBlock (inclusive) for any TransactionDeposited
+// events emitted by the OptimismPortalProxy contract at portalAddress
+func checkForDepositEvents(client *ethclient.Client, portalAddress common.Address, startBlock uint64, endBlock uint64) ([]types.Log, error) {
+	// eventTopic for TransactionDeposited(address indexed from, address indexed to, uint256 indexed version, bytes opaqueData)
+	// - copied from validation/internal/bindings/optimism-portal.go
+	eventTopic := common.HexToHash("0xb3813568d9991fc951961fcb4c784893574240a28925604d09fc577c55bb7c32")
```
sebastianst, line 63: Confirmed that this is the right topic with `cast`:
```
cast sig-event "TransactionDeposited(address indexed from, address indexed to, uint256 indexed version, bytes opaqueData)"
0xb3813568d9991fc951961fcb4c784893574240a28925604d09fc577c55bb7c32
```

validation/start-block_test.go
```diff
@@ -0,0 +1,79 @@
...
+	require.NoError(t, err)
+
+	if offChainParam > onChainParam {
+		// Ensure there aren't any skipped deposits in the block gap
+		portalAddress := Addresses[chain.ChainID].OptimismPortalProxy
+		require.NotZero(t, portalAddress)
+
+		missedEvents, err := checkForDepositEvents(client, common.Address(portalAddress), onChainParam, offChainParam)
+		require.NoError(t, err)
+		require.Zero(t, len(missedEvents))
```
sebastianst, line 40:
```suggestion
		require.Len(t, missedEvents, 0)
```
and note that the `len` of a `nil` slice is defined as `0`.

validation/start-block_test.go
```diff
@@ -0,0 +1,79 @@
...
+	query := ethereum.FilterQuery{
+		FromBlock: big.NewInt(int64(startBlock)),
+		ToBlock:   big.NewInt(int64(endBlock)),
+		Addresses: []common.Address{portalAddress},
+		Topics:    [][]common.Hash{{eventTopic}},
+	}
+
+	missedEvents, err := client.FilterLogs(context.Background(), query)
+	if err != nil {
+		return []types.Log{}, err
```
sebastianst, line 75: unless necessary, I'd rather return
```suggestion
		return nil, err
```

validation/start-block_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	// Create a query
+	query := ethereum.FilterQuery{
+		FromBlock: big.NewInt(int64(startBlock)),
+		ToBlock:   big.NewInt(int64(endBlock)),
+		Addresses: []common.Address{portalAddress},
+		Topics:    [][]common.Hash{{eventTopic}},
+	}
+
+	missedEvents, err := client.FilterLogs(context.Background(), query)
```
sebastianst, line 73: don't we have a timeout defined in the validation checks that we can use here?

validation/start-block_test.go
```diff
@@ -0,0 +1,79 @@
...
+	require.NotEmpty(t, rpcEndpoint)
+
+	client, err := ethclient.Dial(rpcEndpoint)
+	require.NoErrorf(t, err, "could not dial rpc endpoint %s", rpcEndpoint)
+
+	systemConfigAddress := Addresses[chain.ChainID].SystemConfigProxy
+	require.NotZero(t, systemConfigAddress)
+
+	offChainParam := chain.Genesis.L1.Number
+	onChainParam, err := getStartBlockWithRetries(context.Background(), common.Address(systemConfigAddress), client)
```
sebastianst, line 30: would be nice to use a timeout here

sebastianst (commented):
looks great, just some nits

validation/start-block_test.go
bitwiseguy: I'm good with that proposed naming convention. However, there are many existing test files in this repo that use `-` instead of `_`. A few examples
* [gpo-params_test.go](https://github.com/ethereum-optimism/superchain-registry/blob/main/validation/gpo-params_test.go)
* [rollup-config_test.go](https://github.com/ethereum-optimism/superchain-registry/blob/main/validation/rollup-config_test.go)
* etc.

I propose we make that naming convention update in a follow-up pr so that we change all the files at the same time to keep them consistent

validation/start-block_test.go
bitwiseguy: Ahh good to know. I need to start using `cast` more often for these types of tasks

validation/start-block_test.go
bitwiseguy: Added one here https://github.com/ethereum-optimism/superchain-registry/pull/398/commits/9af8df9638afc1a86a6452329ca867fe99dbdf10

validation/start-block_test.go
bitwiseguy: There is not currently a parent ctx passed down to these validation checks. I created a local timeout of 10 seconds for these start block tests: https://github.com/ethereum-optimism/superchain-registry/pull/398/commits/9af8df9638afc1a86a6452329ca867fe99dbdf10

validation/start-block_test.go
sebastianst: I see, wasn't aware of this. I just had never seen `_` and `-` mixed use in Go filenames.

sebastianst (approved):

PR #399, opened 2024-07-18
validation: add gas token test

geoknee:
Changes: 
*  The `add-chain` package records the gas token address in config `.toml` files
* Generated chain list files include this information
* Standard chains are validated to _not_ have a custom gas token
* Since CGT is a beta feature, the on-chain interface does not yet support the queries we are making in the above steps. So we catch "execution reverted" and interpret that as "CGT disabled".


superchain/superchain.go
```diff
@@ -106,7 +106,8 @@ type ChainConfig struct {
 	SequencerWindowSize uint64 `toml:"seq_window_size"`
 
 	// Optional feature
-	Plasma *PlasmaConfig `toml:"plasma,omitempty"`
+	Plasma         *PlasmaConfig `toml:"plasma,omitempty"`
+	GasPayingToken *Address      `toml:"gas_paying_token, omitempty"` // Just metadata, not consumed by downstream OPStack software
```
sebastianst, line 110:
```suggestion
	GasPayingToken *Address      `toml:"gas_paying_token,omitempty"` // Just metadata, not consumed by downstream OPStack software
```
It might need to be this way. It's the usual way at least.

validation/gas-token_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// L1Block .isCustomGasToken() check
+	got, err := getBool("isCustomGasToken()", MustHexToAddress("0x4200000000000000000000000000000000000015"), client)
+	if !strings.Contains(err.Error(), "execution reverted") {
+		// Pre: Custom Gas Token feature. Reverting is acceptable.
+		require.NoError(t, err)
+	} else {
+		// Post: Custom Gas Token fearure. Must be set to false.
+		require.False(t, got)
+	}
```
sebastianst, line 40: Is this right? You're now asserting `got` to be false only if `execution reverted` is in the error, in all other cases you only assert the error to be nil.

Don't you want to make
* if `err != nil` then it must be "exec rev"
* else `got` must be false

same below for the `isCustomGasToken` check

validation/gas-token_test.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func getString(method string, contractAddress Address, client *ethclient.Client) (string, error) {
```
sebastianst, line 57: nit, maybe name this `getHexString` or so, because it's not actually a `string` getter (functions that return type `string`) but a bytes getter, which then returns the bytes converted to a hex string.

validation/gas-token_test.go
```diff
@@ -0,0 +1,105 @@
...
+		To:   &addr,
+		Data: crypto.Keccak256([]byte(method))[:4],
+	}
+
+	// Make the call
+	callContract := func(msg ethereum.CallMsg) ([]byte, error) {
+		return client.CallContract(context.Background(), msg, nil)
+	}
+
+	result, err := Retry(callContract)(callMsg)
```
sebastianst, line 86: you could extract this into a function and share it with `getString`.

validation/gas-token_test.go
```diff
@@ -0,0 +1,105 @@
...
+	// If the result is a string "0x1" or "0x0", convert to bool
+	if string(result) == "0x1" {
+		decodedBool = true
+	} else if string(result) == "0x0" {
+		decodedBool = false
+	} else {
+		return false, errors.New("Unexpected return value, cannot convert to bool")
+	}
+
+	return decodedBool, nil
```
sebastianst, line 104: Since you're not doing anything else with the bool, I'd shorten this to
```suggestion
	switch string(result) {
	case "0x1":
		return true, nil
	case "0x0":
		return false, nil
	default:
		return false, errors.New("unexpected non-bool return value")
	}
```

And note that Go error messages should in general [start small](https://go.dev/wiki/CodeReviewComments#error-strings).

superchain/superchain.go
geoknee: [96ebb0a](https://github.com/ethereum-optimism/superchain-registry/pull/399/commits/96ebb0a0a6b2dc505e805d956ddbd6ad3d680997)

validation/gas-token_test.go
geoknee: [4524bb2](https://github.com/ethereum-optimism/superchain-registry/pull/399/commits/4524bb26aecf72f0eaffb44eda464800363519ec)

validation/gas-token_test.go
geoknee: [4524bb2](https://github.com/ethereum-optimism/superchain-registry/pull/399/commits/4524bb26aecf72f0eaffb44eda464800363519ec)

validation/gas-token_test.go
geoknee: [2ed63de](https://github.com/ethereum-optimism/superchain-registry/pull/399/commits/2ed63de79b4269047f376a7a48f0c54414809dfb)

validation/gas-token_test.go
geoknee: [9f02a52](https://github.com/ethereum-optimism/superchain-registry/pull/399/commits/9f02a52aeb2d05b5fb00b8d44179b7d16c2bcc81)

add-chain/main.go
```diff
@@ -204,11 +204,14 @@ func getGasPayingToken(l1rpcURl string, SystemConfigAddress superchain.Address)
...
 	result, err := sc.GasPayingToken(&opts)
+
 	if strings.Contains(err.Error(), "execution reverted") {
 		// This happens when the SystemConfig contract
 		// does not yet have the CGT functionality.
 		return nil, nil
-	} else if err != nil {
+	}
+
+	if err != nil {
```
sebastianst, line 214: I think this was actually fine before.

sebastianst (approved):

add-chain/main.go
geoknee: Yeah its a minor one / style / personal taste. In a more complicated function, I like doing this because it can make it easier to read https://danp.net/posts/reducing-go-nesting/ .

PR #408, opened 2024-07-19
feat: Hydrated Superchain configs.toml

refcell:
**Description**

Re-organizes the `configs.toml` in `superchain/configs/` so that instead of just a list of `ChainConfig` objects, the file contains a list of superchain configuration structs, each containing a list of `ChainConfig`s that are part of that Superchain.

This avoids the need to read the `superchain.toml` configs in each of [`mainnet`, `sepolia`, sepolia-dev-0`] to get superchain config info including the protocol version address and superchain config address.

PR also updates rust bindings (significantly cleaner) and adds the optional superchain config address to the `RollupConfig` struct.

**New Config Layout**

```
Superchains
- Mainnet {
	Name: Mainnet
	Config: Superchain Config { ... }
	ChainConfigs
		- OP Mainnet Chain Config { ... }
		- Base Mainnet Chain Config { ... } 
		...
}
- Sepolia { ... }
- Sepolia-dev-0 { ... }
```


superchain/internal/codegen/main.go
```diff
@@ -100,21 +100,41 @@ func main() {
...
-		var buf bytes.Buffer
-		if err := toml.NewEncoder(&buf).Encode(Configs{Configs: cfgs}); err != nil {
-			panic(fmt.Errorf("failed to marshal toml: %w", err))
-		}
+  superchains := make([]Superchain, 0)
+  for _, sc := range Superchains {
+    chainConfigs := make([]ChainConfig, 0)
+    for _, chainId := range sc.ChainIDs {
+      chain := OPChains[uint64(chainId)]
+      if chain == nil {
```
geoknee, line 113: Is it more idiomatic to do
```suggestion
      chain, found := OPChains[uint64(chainId)]
      if !found {
```

justfile
```diff
@@ -15,7 +15,7 @@ promote-to-standard CHAIN:
 
 # Generate auto-generated files
 codegen: clean-add-chain
-	go run superchain/internal/codegen/main.go
+	CODEGEN=true go run superchain/internal/codegen/main.go
```
geoknee, line 18: Curious about why this is needed?

geoknee (approved):

justfile
refcell: Ah, so since the codegen is run in circleci, the configured rpcs are used instead of the default ones in the superchain init when the `Superchains` list is created. Since that list is used in the configs.toml generation to get the `SuperchainConfig` object (containing the rpcs), those rpcs are different when run locally without the [circle ci rpc envs](https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/init.go#L76-L77) set and remotely in circleci environment.

bindings/rust-bindings/src/superchain.rs
```diff
@@ -2,11 +2,23 @@
...
-    /// Chain configs.
+pub struct HydratedSuperchainConfigs {
+    /// A list of hydrated superchain configs.
+    pub superchains: Vec<HydratedSuperchainConfig>,
+}
+
+/// A hydrated Superchain Config.
+#[derive(Debug, Clone, Default, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct HydratedSuperchainConfig {
```
clabby, line 16: Is there a non-hydrated `SuperchainConfig` type I'm missing? If not, would pref we drop the `Hydrated` prefix to the type names.

clabby (approved):
Rust code lgtm. Small nit on the type names.

bindings/rust-bindings/src/superchain.rs
refcell: There is, see right below on line 20

PR #412, opened 2024-07-22
fix(bindings): Repair Releases

refcell:
**Description**

Fixes golang codegen to dup static file output to rust crate so it can be included in the crate upload.

Also cleans up crates and their dependencies quite a bit - removes a lot of cruft and splits code out nicely into different modules.

The ci release workflow should now work, will put that up in another pr.


clabby (approved):
Rust code lgtm, pending go reviewer's approval.

bindings/rust-bindings/etc/chainList.toml
```diff
@@ -0,0 +1,197 @@
+[[chains]]
```
bitwiseguy, line 1: Can we just use the existing `<repo-root>/chainList.toml` file instead of duplicating here?

superchain/configs/configs.toml
```diff
@@ -4,16 +4,16 @@
 
 [[superchains]]
```
bitwiseguy, line 5: This file is only currently used by the rust code. Could we remove this entire file (and the codegen code that creates it) if you prefer for the file to be written to the rust dir? Or do you think this file will be useful in the future for other languages to create their own bindings?

bitwiseguy (approved):

bindings/rust-bindings/etc/chainList.toml
refcell: No we can't unfortunately because then it won't be included in the crates release and when someone uses the crate, it won't build because the file is missing. I will continue trying to figure out a work around but this is the best solution I found after a few hours

superchain/configs/configs.toml
refcell: Ah that's a great idea actually, will do.

bindings/rust-bindings/etc/chainList.toml
bitwiseguy: Ok no worries. Doesn't need to block this pr, just a potential future cleanup

PR #414, opened 2024-07-23
editorconfig: standardize spacing in files

bitwiseguy:
By specifying standard white space settings we should avoid unnecessarily large diffs where devs with different settings overwrite spacing without changing code


refcell (approved):
Nice

Issue #410, opened 2024-07-19
display eth DA / alt DA status


tessr:
### Request

The Superchain Registry currently checks whether a chain uses Alt-DA or Eth DA. We should surface this in the machine/human readable files with a new `data_availability_type` field. This field can be set to either `eth-da` or `alt-da`.

For example:

`chainList.toml`: 

```toml
[[chains]]
  name = "OP Mainnet"
  identifier = "mainnet/op"
  chain_id = 10
  rpc = ["https://mainnet.optimism.io"]
  explorers = ["https://explorer.optimism.io"]
  superchain_level = 2
  data_availability_type = "eth-da"
  [chains.parent]
    type = "L2"
    chain = "mainnet"
```

`chainList.json`:

```json
  {
    "name": "OP Mainnet",
    "identifier": "mainnet/op",
    "chainId": 10,
    "rpc": [
      "https://mainnet.optimism.io"
    ],
    "explorers": [
      "https://explorer.optimism.io"
    ],
    "superchainLevel": 2,
    "data_availability_type":"eth-da",
    "parent": {
      "type": "L2",
      "chain": "mainnet"
    }
  },
```

`superchain/configs/op.toml`:
```toml
name = "OP Mainnet"
chain_id = 10
public_rpc = "https://mainnet.optimism.io"
sequencer_rpc = "https://mainnet-sequencer.optimism.io"
explorer = "https://explorer.optimism.io"
superchain_level = 2
standard_chain_candidate = false
data_availability_type = "eth-da"
```

This is **not** blocking the EA launch, but we should populate this data soon after launch because our partners expect to see it.

### Implementation Constraints

We are currently only checking the purple section of the following diagram. If a chain has set a `PlasmaConfig`, they are using Alt-DA Mode. Otherwise, they use Eth DA. 

<img src="https://github.com/user-attachments/assets/992803b9-03b6-4ae6-bb6d-da98e345bf62" width="600" />
<br /><br />

For the gov launch, we should also:

* Determine which DA provider Alt-DA chains are using (e.g., the blue boxes). See also https://github.com/ethereum-optimism/superchain-registry/issues/406
* Perform an onchain check to confirm that batcher transactions are posting calldata/blobdata that matches their stated DA type. See also https://github.com/ethereum-optimism/superchain-registry/issues/400

bitwiseguy:
Could we use this data structure:
```toml
[[chains]]
  name = "OP Mainnet"
[...]

PR #417, opened 2024-07-23
Add data_availability_type field to ChainConfig

bitwiseguy:
Fixes https://github.com/ethereum-optimism/superchain-registry/issues/410


add-chain/cmd/update-configs.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"path/filepath"
+	"runtime"
+
+	"github.com/ethereum-optimism/superchain-registry/add-chain/config"
+	"github.com/ethereum-optimism/superchain-registry/superchain"
+	"github.com/urfave/cli/v2"
+)
+
+var UpdateConfigsCmd = cli.Command{
```
bitwiseguy, line 12: We can delete this whole file if we want. But could be useful to automate tasks such as adding hardfork timestamps in the future

add-chain/config/config.go
```diff
@@ -15,31 +15,6 @@ import (
...
-		if c.PlasmaConfig.DAChallengeWindow == nil {
-			return fmt.Errorf("missing required field: da_challenge_window")
-		}
-		if c.PlasmaConfig.DAResolveWindow == nil {
-			return fmt.Errorf("missing required field: da_resolve_window")
-		}
-	}
-	return nil
-}
-
```
bitwiseguy, line 42: Was able to get rid of this extraneous `JSONChainConfig` struct by just transferring the same `json` tags to the `superchain.ChainConfig` struct. Also moved the `JSONChainConfig.VerifyPlasma()` method to `superchain.ChainConfig.CheckDataAvailability()`

superchain/configs/mainnet/op.toml
```diff
@@ -29,25 +29,25 @@ seq_window_size = 3600
     gasLimit = 30000000
 
 [addresses]
```
bitwiseguy, line 31: All of the `addresses` diffs shown in this PR are due to reordering the fields. Previously they were alphabetically ordered, but the toml package we are using orders them by their placement in the struct. The `just codegen` command run as part of CI confirms that none of the addresses written to `superchain/extra/addresses/addresses.json` have changed

superchain/superchain.go
```diff
@@ -117,6 +125,23 @@ type PlasmaConfig struct {
...
+		daType = AltDA
+		if c.Plasma.DAChallengeAddress == nil {
+			return "", fmt.Errorf("missing required plasma field: da_challenge_contract_address")
+		}
+		if c.Plasma.DAChallengeWindow == nil {
+			return "", fmt.Errorf("missing required plasma field: da_challenge_window")
+		}
+		if c.Plasma.DAResolveWindow == nil {
+			return "", fmt.Errorf("missing required plasma field: da_resolve_window")
+		}
```
axelKingsley, line 140: I'm not sure how completely you want to validate DA, but checking for the windows seems not needed.  

if `use_plasma` (the deprecated/legacy flag) is specified, the other fields *should also* be specified. Same with `PlasmaConfig`. What that means to me is that the enablement check should only be concerned with that top level. But I am out of my element a bit here, and maybe you do want to do these secondary checks.

axelKingsley (commented):
I see lots of other changes in the `toml` files besides DA, like `ProxyAdmin`.

Regarding the DA check specifically, it looks fine.

bitwiseguy:
> I see lots of other changes in the `toml` files besides DA, like `ProxyAdmin`.

@axelKingsley - the values of the addresses haven't changed, but the ordering of the fields in the toml files has. Explained a bit more here https://github.com/ethereum-optimism/superchain-registry/pull/417#discussion_r1688553825

superchain/superchain.go
bitwiseguy: We're not checking for a `use_plasma` field anymore. We are just checking that `plasma_config` is not nil in the `rollup.json` (e.g. [here](https://github.com/ethereum-optimism/superchain-registry/blob/main/add-chain/testdata/monorepo/op-node/rollup_plasma.json#L34)). If that parent field is present, then we check that the other fields exist. Happy to remove the two window field checks if there is a valid version of the `plasma_config` that does not need them

bitwiseguy: I added checks for the legacy plasma fields in the `rollup.json` file that is an input for the `add-chain` code. The `add-chain` code uses a consistent format (always the nested, dedicated `PlasmaConfig` struct instead of flattened fields) for plasma in its toml config file output

add-chain/cmd/update-configs.go
bitwiseguy: Leaving this in for future use

axelKingsley (approved):
My approval is based on the `CheckDataAvailability` function and surrounding code, and assumes the other changes are correct (as Sam and I talked about it offline)

refcell (approved):

Issue #400, opened 2024-07-19
validation: autodetect, store and continuously validate Data Availability mode. 


geoknee:
The Data Availability mode (Ethereum vs non Ethereum) and provider (Celestia, etc) are important pieces of information to attach to each chain in the registry. 

We are not able to infer the DA mode from `rollup.json` -- there are some environment variables for op-node and op-batcher which have useful information, but will not be able to reliably scrape these, either. We have a config field `PlasmaConifg` but it is not clear if this is actually the whole story with alt DA.

The remaining options are to:
* ask and trust the chain contributor to declare the information
* go on chain to detect and verify it. 

We will go with the second option. 

Definition of done:

* `add-chain` tooling uses an L1 RPC client to scrape the latest batcher transaction* for the chain being added. It decodes that and infers if the chain is using alt DA (i.e. the DA mode boolean)
  - ~IF the chain is using alt DA, we require them to have set an "ALT_DA_PROVIDER_BYTE`~ 
*  a new field is added to the `superchain.ChainConfig` type to record the boolean from the previous step, ~as well as the identifier byte if the DA mode is not ethereum. For now, the identifier byte can be stored in a field `SelfCertifiedDAProviderByte` to emphasise that it is not yet validated on chain.~
* a validation check is added, which again scrapes the latest batcher transaction*, decodes the DA mode, and
  - for all chains checks it is consistent with the declared value now stored in the chain config
  - for standard chains (and standard candidate chains), ensures the DA mode is Ethereum
* ~This list of bytes is represented in the `superchain-registry` https://github.com/ethereum-optimism/specs/discussions/135 as an enum~

---
 *Scraping latest batcher transaction
This is likely a useful reusable module for many validation checks, as well as the add-chain tooling. The idea is to do a binary search on the batcher signing account's nonce. This is a very quick way of finding the latest batcher transaction. The initial lookback window could be just the sequencing window. 

---
EDIT: cutting some things from scope, these will be dealt with in a follow up ticket

bitwiseguy:
@sebastianst as a first step, should we just check whether the `BatchSubmitter` address for a chain has sent a tx in the last `SequencerWindow` l1 blocks?

[...]

Issue #406, opened 2024-07-19
validation: detect and verify alt-da provider


geoknee:
If DA mode is not Ethereum

* create new file that stores mapping of alt-da byte to provider name
* add-chain: env-var to specify alt-da byte
* validation: check alt-da byte and verify it matches location where data is posted.

tessr:
The idea here is to handle the blue part of the Alt-DA validation check tree: 

![image](https://github.com/user-attachments/assets/5c5746fc-108b-4cdf-89a2-cc0509fffc82)

We could display it like this: 

`superchain/configs/mainnet/lyra.toml`

```toml
name = "Lyra Chain"
chain_id = 957
public_rpc = "https://rpc.lyra.finance"
sequencer_rpc = "https://rpc.lyra.finance"
explorer = "https://explorer.lyra.finance"
superchain_level = 1
standard_chain_candidate = false
data_availability_type = "eth-da" # covered in 410
data_availability_provider = "celestia" # this is the field to add and validate
```

I am also open to this being an enum.

PR #421, opened 2024-07-25
validation: verify eth-da via latest batcher tx

bitwiseguy:
## Description

Use the updated `op-service/eth` packages `CheckRecentTxs` to perform a binary search to find the most recent batcher tx. If the tx exists within the last `ChainConfig.SequencerWindowSize` blocks on mainnet, then the chain passes the test for `eth-da`. Only checking `eth-da` for now. 

## Additional context

Follow-up work will be to detect and verify `alt-da`, as described in https://github.com/ethereum-optimism/superchain-registry/issues/406

## Metadata

Closes https://github.com/ethereum-optimism/superchain-registry/issues/400


validation/data-availability-type_test.go
```diff
@@ -0,0 +1,45 @@
...
+	depth := chain.SequencerWindowSize
+	require.NotZero(t, depth)
+
+	// First attempt without needing an archive node
+	//  - full nodes keep the last 128 blocks but we use 120 to give a buffer
+	nonArchivalDepth := 120
+	_, found, err := eth.CheckRecentTxs(ctx, client, nonArchivalDepth, common.Address(batchSubmitterAddress))
+	if !found {
+		_, found, err = eth.CheckRecentTxs(ctx, client, int(depth), common.Address(batchSubmitterAddress))
+	}
```
sebastianst, line 42: what's the purpose of first only checking the non-archival range?

validation/data-availability-type_test.go
```diff
@@ -0,0 +1,45 @@
...
+	require.NoErrorf(t, err, "failed when checking chain for recent batcher txs from %s", batchSubmitterAddress)
```
sebastianst, line 43: I think we should actually find that tx and check that it was sent to the batcher inbox. That's an additional cheap sanity check we can do here.

validation/data-availability-type_test.go
bitwiseguy: Because locally I (and likely many other chain operators) do not have access to an archival node. However, our CI does

validation/data-availability-type_test.go
bitwiseguy: https://github.com/ethereum-optimism/superchain-registry/pull/421/commits/ce5c6a61bbf97927aed11245602756a001bc76bb

validation/data-availability-type_test.go
sebastianst: We can leave it in since it doesn't hurt. But we could also just require devs and CIs that run this code to have access to an archival node, as a failing test on a full node is now inconclusive. I see that this actually works well for high-throughput chains, where batcher txs are frequent and probably below the 120 L1 blocks limit. But for low-throughput chains, which may be a probable case for new SCR arrivals, it could be misleading.

Note that you're currently not inspecting the first `err` assignment, since if `!found` you retry again. You might want to at least log the first `err`, if entering the `if !found` branch and `err` is non-nil.

validation/data-availability-type_test.go
```diff
@@ -0,0 +1,66 @@
...
+	require.True(t, found)
+
+	// Fetch the block by number
+	block, err := client.BlockByNumber(context.Background(), new(big.Int).SetUint64(blockNum))
+	require.NoError(t, err)
+
+	// Iterate over the transactions in the block
+	for _, tx := range block.Transactions() {
+		signer := types.NewCancunSigner(tx.ChainId())
```
sebastianst, line 57: You could maybe use [`MakeSigner`](https://pkg.go.dev/github.com/ethereum/go-ethereum/core/types#MakeSigner) to be more forwards compatible.

validation/data-availability-type_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+	// First attempt without needing an archive node
+	//  - full nodes keep the last 128 blocks but we use 120 to give a buffer
+	nonArchivalDepth := 120
+	blockNum, found, err := eth.CheckRecentTxs(ctx, client, nonArchivalDepth, common.Address(batchSubmitterAddress))
+	if !found {
+		blockNum, found, err = eth.CheckRecentTxs(ctx, client, int(depth), common.Address(batchSubmitterAddress))
+	}
+	require.NoErrorf(t, err, "failed when checking chain for recent batcher txs from %s", batchSubmitterAddress)
+	require.True(t, found)
```
sebastianst, line 49:
```suggestion
	require.True(t, found, "failed to find recent batcher transaction")
```

sebastianst (commented):
Note this also needs a rust reviewer now.

bindings/rust-bindings/etc/configs.toml
```diff
@@ -676,7 +676,7 @@
       Challenger = "0x45eFFbD799Ab49122eeEAB75B78D9C56A187F9A7"
       Proposer = "0x0000000000000000000000000000000000000000"
       UnsafeBlockSigner = "0x0000000000000000000000000000000000000000"
-      BatchSubmitter = "0x0000000000000000000000000000000000000000"
+      BatchSubmitter = "0xdb80Eca386AC72a55510e33CF9CF7533e75916eE"
```
bitwiseguy, line 679: This address was omitted from the original config. Added it after the retrieving the value onchain. We need this value to be set in order to perform the da availability test added by this pr

validation/data-availability-type_test.go
bitwiseguy: https://github.com/ethereum-optimism/superchain-registry/pull/421/commits/74ff586420d2bcef55e81709902413f57b3b84a3

sebastianst (approved):

refcell (approved):
Nice!

PR #423, opened 2024-07-26
updating superchain registry

sbvegan:
**Description**

Removing superlumio and pgn from the registry:

- deleted their `superchain/configs/<superchain-target>/<network>.toml` files and then ran `just codegen`.
- deleted their `extra/genesis/<superchain-target>/<network>.json.gz`

Is there additional genesis clean up that needs to be done?


refcell (approved):
This LGTM, it doesn't look like there's any other work to do - I see no other mentions of PGN.

sbvegan:
I'm not sure why this check is failing. I'm not seeing any issues locally and I don't have the option to rerun the check.

PR #425, opened 2024-07-29
add-chain: fix issues found via binary walkthrough

bitwiseguy:
## Description

Fixes the following bugs:
1. support customGasToken contract set (which is different from faultProof and non-fault proof sets)
2. allow user to pass the full filepath of a deployments file instead of always expecting file named `.deploy`
3. wrap `SCR_CHAIN_NAME` in quotes to support names with spaces

## Additional changes
Added a test for legacy deployment files (one .json file for each contract instead of a single .deploy file). Previously this was untested code


add-chain/main.go
```diff
@@ -205,14 +214,12 @@ func getGasPayingToken(l1rpcURl string, SystemConfigAddress superchain.Address)
 
 	opts := bind.CallOpts{}
 	result, err := sc.GasPayingToken(&opts)
-
-	if strings.Contains(err.Error(), "execution reverted") {
-		// This happens when the SystemConfig contract
-		// does not yet have the CGT functionality.
-		return nil, nil
-	}
```
bitwiseguy, line 213: This caused the program to crash when `err == nil`

add-chain/addresses.go
```diff
@@ -142,68 +142,94 @@ func readAddressesFromJSON(contractAddresses map[string]string, deploymentsDir s
...
+			// Use legacy deployment artifact schema
+			fmt.Printf("failed to find .deploy file. Will look for legacy .json files")
+
+			if _, err := os.ReadFile(filepath.Join(deploymentsDir, FaultDisputeGame+".json")); errors.Is(err, os.ErrNotExist) {
+				contracts = append(contracts, faultProofContracts...)
+			} else if _, err := os.ReadFile(filepath.Join(deploymentsDir, AnchorStateRegistryProxy+".json")); errors.Is(err, os.ErrNotExist) {
+				contracts = append(contracts, customGasTokenContracts...)
+			} else {
+				contracts = append(contracts, L2OutputOracleProxy)
 			}
```
sebastianst, line 189: I'm a bit confused here. So if the `FaultDisputeGame+".json"` can _not_ be found, the FP contracts are appended to the `contracts` slice? And then again, if the `AnchorStateRegistryProxy+".json"` can _not_ be found, the CGT contracts are appended? Was the idea to negate `!` the `error.Is` checks? This would also better match the corresponding code below in [lines 217+](https://github.com/ethereum-optimism/superchain-registry/pull/425/files#diff-c8e7fef18ff265070f062af3aa6f25275760db61b16aac008aaca33c55e201acR217-R223).

Also, are FPs and CGTs mutually exclusive?

add-chain/addresses.go
```diff
@@ -142,68 +142,94 @@ func readAddressesFromJSON(contractAddresses map[string]string, deploymentsDir s
...
 
-		for _, name := range contracts {
-			address, err := addressList.AddressFor(name)
-			if err != nil {
-				return fmt.Errorf("failed to retrieve %s address from list: %w", name, err)
-			}
-			contractAddresses[name] = address.String()
+	if _, err = addressList.AddressFor(FaultDisputeGame); err == nil {
+		contracts = append(contracts, faultProofContracts...)
+	} else if _, err = addressList.AddressFor(AnchorStateRegistryProxy); err == nil {
```
sebastianst, line 219: so FDG and CGT are mutually exclusive?

add-chain/main.go
sebastianst: that makes sense 😅

superchain/init.go
sebastianst, line 1: what's with the whitespace diff in this file?

bitwiseguy: I believe the linter just changed this whitespace from spaces to tabs

add-chain/addresses.go
bitwiseguy: Good catch. That logic is backwards. We want to use the FP contract set if `FaultDisputeGame.json` CAN be found.

AFAIK a chain using FP cannot use CGT. At least for right now. Can you confirm this is correct @sbvegan?

sbvegan: Yeah my understanding is CGT chains do not support fault proofs. cc @tynes [to confirm ](https://github.com/ethereum-optimism/specs/discussions/140#discussioncomment-10191608)

sbvegan: confirmed that this is the case at the moment, once https://github.com/ethereum-optimism/optimism/pull/10780 is merged then we will support cgt with fp

add-chain/addresses.go
sbvegan: at the moment, once https://github.com/ethereum-optimism/optimism/pull/10780 is merged then we will support cgt with fp

add-chain/addresses.go
bitwiseguy: Instead of hardcoding the list of expected addresses to find in the .deploy (or individual .json files), https://github.com/ethereum-optimism/superchain-registry/pull/425/commits/2df905bb646a753ae3ef2c35d0f75865ee341df4 improves the `readAddressesFromJSON` function to read all available addresses. This means we no longer have to maintain that list and the logic around it as features and expected contract sets evolve. We can rely on downstream add-chain and validation tests to check for the existence of required contracts. The same commit also adds a test case for the legacy deployment file type (one .json file for each contract instead of all in a single .deploy file). Previously that was untested code.

superchain/init.go
sebastianst: Would be great to run the linter in CI, so we catch these things at the PR level

add-chain/addresses.go
```diff
@@ -15,154 +15,89 @@ type AddressData struct {
...
-)
-
-func readAddressesFromChain(addresses map[string]string, l1RpcUrl string, isFaultProofs bool) error {
+func readAddressesFromChain(addresses *superchain.AddressList, l1RpcUrl string, isFaultProofs bool) error {
 	// SuperchainConfig
-	address, err := castCall(addresses[OptimismPortalProxy], "superchainConfig()(address)", l1RpcUrl)
+	address, err := castCall(addresses.OptimismPortalProxy.String(), "superchainConfig()(address)", l1RpcUrl)
 	if err != nil {
-		addresses[SuperchainConfig] = ""
+		addresses.SuperchainConfig = superchain.Address{}
```
sebastianst, line 22: the default value is already a zero array, no need to set it explicitly

add-chain/addresses.go
```diff
@@ -15,154 +15,89 @@ type AddressData struct {
...
-	DelayedWETHProxy         = "DelayedWETHProxy"
-	DisputeGameFactoryProxy  = "DisputeGameFactoryProxy"
-	FaultDisputeGame         = "FaultDisputeGame"
-	MIPS                     = "MIPS"
-	PermissionedDisputeGame  = "PermissionedDisputeGame"
-	PreimageOracle           = "PreimageOracle"
-)
-
-func readAddressesFromChain(addresses map[string]string, l1RpcUrl string, isFaultProofs bool) error {
+func readAddressesFromChain(addresses *superchain.AddressList, l1RpcUrl string, isFaultProofs bool) error {
```
sebastianst, line 18: There's quite some repetitive boilerplate in this function, maybe it could be done shorter and easier extensible by iterating over a
```go
for _, x := range []struct{
  contract superchain.Address
  call     string
  target   *superchainAddress
  fp       *bool
  allowErr bool
}{
  {
    contract: addresses.OptimismPortalProxy
    call      "superchainConfig()(address)"
    target    &addresses.SuperchainConfig
  },
  // more entries to read
} {
  // call x.call on x.contract and save to *x.target
  // if x.fp non-nil, read entry only if *x.fp == isFaultProofs
  // if allowErr is true, don't return on error, just log
  // could also check if the target address is non-zero before setting, this way this can also be used for the guardian where the first entry could then have allowErr = true, so the second entry is only read if the first failed
  address, err := castCall(...)
}
```
I see that we'd need further logic for the batcher... maybe we also add an optional field `toAddress func(string) Address` and use this one instead of the default `superchain.MustHexToAddress` and then can put the batcher hash extraction code in there.

add-chain/addresses.go
```diff
@@ -15,154 +15,89 @@ type AddressData struct {
...
 	// SuperchainConfig
-	address, err := castCall(addresses[OptimismPortalProxy], "superchainConfig()(address)", l1RpcUrl)
+	address, err := castCall(addresses.OptimismPortalProxy.String(), "superchainConfig()(address)", l1RpcUrl)
```
sebastianst, line 20: Many `castCall` calls seem to now have an `Address` at hand, instead of the address `string`. If there aren't many invocations left to `castCall` that pass a string instead of an `Address`, maybe a good opportunity to clean it up and let `castCall`'s first argument be of type `Address`. It would be more safer anyways, since the argument _must_ be an address.

add-chain/addresses.go
```diff
@@ -15,154 +15,89 @@ type AddressData struct {
...
-		DisputeGameFactoryProxy,
-		FaultDisputeGame,
-		MIPS,
-		PermissionedDisputeGame,
-		PreimageOracle,
-	}
-
-	contracts := universalContracts
-
+func readAddressesFromJSON(addressList *superchain.AddressList, deploymentsDir string) error {
```
sebastianst, line 100: So downstream checks would detect any missing address?

add-chain/addresses.go
sebastianst: Nice improvement!
> We can rely on downstream add-chain and validation tests to check for the existence of required contracts.

So this is guaranteed for _every_ contract?

sebastianst (commented):
Nice improvement!

add-chain/addresses.go
sebastianst: Given urgency to get this in, let's do this in a follow up.

add-chain/addresses.go
sebastianst: Can also do this in a follow-up

superchain/init.go
bitwiseguy: Its already being run [here](https://github.com/ethereum-optimism/superchain-registry/blob/main/.circleci/config.yml#L47-L54). I'm guessing its not working correctly because we don't run the `git diff --exit-code` after running the linter

add-chain/addresses.go
bitwiseguy: https://github.com/ethereum-optimism/superchain-registry/pull/425/commits/0b48204ccb9226ea7a2a1e0442db21a5f3d54ca3

add-chain/addresses.go
bitwiseguy: Created a new issue to capture this: https://github.com/ethereum-optimism/superchain-registry/issues/427

add-chain/addresses.go
bitwiseguy: https://github.com/ethereum-optimism/superchain-registry/pull/425/commits/df412f3da8785483b5dabb20388190bc0663c376

superchain/init.go
bitwiseguy: https://github.com/ethereum-optimism/superchain-registry/pull/425/commits/a9968585226688bc6e50ef63c59d17eef2ec61f6

add-chain/addresses.go
bitwiseguy: This is not guaranteed for every contract, but we do want to allow frontier chains to join the registry with non-standard contract sets. We rely on validation tests, especially the [standard tests](https://github.com/ethereum-optimism/superchain-registry/blob/main/validation/validation_test.go#L47-L63) to verify that contracts are correctly configured based on our expectations of the features on a standard chain.

sebastianst (approved):

PR #431, opened 2024-08-01
fix(codegen): Order by Name

refcell:
**Description**

Internal codegen previously generated the rust-bindings `configs.toml` and ordered items using the L1 chain id. This pr updates it to use the `name` string for sorting, fixing the flake where sorting changes in CI due to duplicate L1 chain ids in various configs.

This fixes the flake seen in https://github.com/ethereum-optimism/superchain-registry/pull/423.

### DIFF
--- a/bindings/rust-bindings/etc/configs.toml
+++ b/bindings/rust-bindings/etc/configs.toml
@@ -14,70 +14,15 @@
       explorer = "https://etherscan.io"
 
   [[superchains.chains]]
-    name = "OP Mainnet"
-    chain_id = 10
-    public_rpc = "https://mainnet.optimism.io"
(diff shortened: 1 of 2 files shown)


Inphi (approved):

bitwiseguy (approved):
The chain name _should_ be globally unique but we don't actually enforce that as far as I know. The l2 `chain_id` _must_ be globally unique and this is enforced before the chain is added to the superchain-registry so it might be better to sort by that

refcell:
> The chain name _should_ be globally unique but we don't actually enforce that as far as I know. The l2 `chain_id` _must_ be globally unique and this is enforced before the chain is added to the superchain-registry so it might be better to sort by that

Good suggestion, I've updated it with this logic.

clabby (approved):

PR #433, opened 2024-08-02
Add granite activation dates

mbaxter:
**Description**

Add granite activation dates.


mbaxter:
@axelKingsley @sebastianst - can you confirm these dates look good?

refcell (approved):
Approving for rust-reviewers

superchain/configs/sepolia/op.toml
```diff
@@ -10,6 +10,7 @@ canyon_time = 1699981200 # Tue 14 Nov 2023 17:00:00 UTC
 delta_time = 1703203200 # Fri 22 Dec 2023 00:00:00 UTC
 ecotone_time = 1708534800 # Wed 21 Feb 2024 17:00:00 UTC
 fjord_time = 1716998400 # Wed 29 May 2024 16:00:00 UTC
+granite_time = 1723132800 # Thu Aug 8 16:00:00 UTC
```
sebastianst, line 13: We usually want to give at least 1 week notice for Testnet releases. This would only be 6 days, and it's already Friday and we would still need to pull this into op-geth and op-node, and run on a devnet for a few hours at least, before we roll this out to testnet and make a release.

sebastianst: Aug 12 works.

mbaxter: updated

superchain/configs/sepolia/superchain.toml
```diff
@@ -6,6 +6,7 @@ canyon_time =  1699981200 # Tue 14 Nov 2023 17:00:00 UTC
 delta_time =   1703203200 # Fri 22 Dec 2023 00:00:00 UTC
 ecotone_time = 1708534800 # Wed 21 Feb 2024 17:00:00 UTC
 fjord_time =   1716998400 # Wed 29 May 2024 16:00:00 UTC
+granite_time = 1723478400 # Mon Aug 12 16:00:00 UTC
```
sebastianst, line 9: The activation time should only be set in the `superchain.toml` and then pulled into all other chain's configurations automatically. We only specify in individual chain configs if we need to _override_ the value from the superchain config.

sebastianst:
There's a missing `createTimestampComment` entry for `granite_time` in `GenerateTOMLComments` in `superchain.go`. But this shouldn't cause the Rollup Config Diff check to not show a diff. The granite time is correctly added to the `HardForkConfiguration` struct.

mbaxter:
Moving this back to draft while we resolve the issue with the chain config diff.

sebastianst:
@mbaxter after https://github.com/ethereum-optimism/optimism/pull/11339 got merged, you can update the `validation` go module's monorepo dependency to latest `develop` and then the diff should show the newly expected granite timestamps.

sebastianst (approved):
Pre-approving so this can be merged once the monorepo dependency of `validation` is updated to latest monorepo and the GH action diff shows the newly expected granite timestamps.

Inphi (approved):
LGTM. Diff check failing is expected given that we're adding new activation time for the hf. The diff checks confirm all chain configs are updated to include `granite_time`.

PR #436, opened 2024-08-05
Set granite mainnet activation date

mbaxter:
## Description

Set granite mainnet activation date.

### DIFF
--- a/superchain/configs/mainnet/superchain.toml
+++ b/superchain/configs/mainnet/superchain.toml
@@ -6,6 +6,7 @@ canyon_time =  1704992401 # Thu 11 Jan 2024 17:00:01 UTC
 delta_time =   1708560000 # Thu 22 Feb 2024 00:00:00 UTC
 ecotone_time = 1710374401 # Thu 14 Mar 2024 00:00:01 UTC
 fjord_time =   1720627201 # Wed 10 Jul 2024 16:00:01 UTC
+granite_time = 1724428800 # Fri 23 Aug 2024 16:00:00 UTC
 
 [l1]
   chain_id = 1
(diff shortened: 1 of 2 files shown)


sebastianst (approved):

superchain/configs/mainnet/superchain.toml
```diff
@@ -6,6 +6,7 @@ canyon_time =  1704992401 # Thu 11 Jan 2024 17:00:01 UTC
 delta_time =   1708560000 # Thu 22 Feb 2024 00:00:00 UTC
 ecotone_time = 1710374401 # Thu 14 Mar 2024 00:00:01 UTC
 fjord_time =   1720627201 # Wed 10 Jul 2024 16:00:01 UTC
+granite_time = 1724428800 # Fri 23 Aug 2024 16:00:00 UTC
```
refcell, line 9: The reason why CI is failing is because the `granite_time` is set in the `superchain.toml` but not in any of the chain configs. Is this the intended timestamp for all chains in the superchain or just op mainnet? If so, we should add this line to each of the chain configs in the `mainnet/` dir. I'm not sure if there's a command to automatically do this. I think it would be neat to introduce this command or subcommand if it doesn't exist in a follow on pr.

If not, we should just add this hardfork timestamp to `op.toml` instead of `superchain.toml` I believe.

refcell (approved):

superchain/configs/mainnet/superchain.toml
mbaxter: Discussed offline.  The hf times are inherited by other chains that set a superchain_time field.  For details see the [readme here](https://github.com/ethereum-optimism/superchain-registry/blob/192aaad0d363ff7e269a16eb688066772d3b6796/README.md?plain=1#L90).   

There is also a plan to make this validation check report a status rather than failing to avoid confusion.

PR #438, opened 2024-08-06
Add l1 contract bytecode validation 

### LINKED ISSUE #405: bytecode validation for L1 contracts
at the moment we are checking the contract version, but that is not a final solution for this check.
So we want to check contract logic and do the bytecode validation for L1 contracts we are using.

geoknee:
Closes #405 

Open questions:
* should we keep the existing `.version()` check? It is more informative when a chain fails to print out the expected and actual versions
* are we 100% sure the bytecodes are static across all chains in the superchain, post MCP? They are different across superchains (I needed a different list of bytecode hashes for sepolia versus mainnet). 
* are there contracts which previously we were unable to check the version of, but we can now check the bytecode (hash) of?


geoknee:
> * should we keep the existing `.version()` check? It is more informative when a chain fails to print out the expected and actual versions

Almost certainly yes IMO.

validation/standard/standard-versions-sepolia.toml
```diff
@@ -0,0 +1,14 @@
+["op-contracts/v1.4.0"] # Fault Proofs https://github.com/ethereum-optimism/optimism/releases/tag/op-contracts%2Fv1.4.0
```
geoknee, line 1: I think due to the "proxy" issue, we may end up only needing one file (not split by mainnet and sepolia).

geoknee: These bytecodes are not currently correct, we need to get them from OP Mainnet.

geoknee:
Update: so far on this PR I failed to properly get the bytecode of the _implementation_ contract sitting behind the proxy.

validation/superchain-version_test.go
```diff
@@ -72,54 +70,48 @@ func getContractVersionsFromChain(list AddressList, client *ethclient.Client) (C
...
-	isemver, err := bindings.NewISemver(addr, client)
-	if err != nil {
-		return "", fmt.Errorf("%s: %w", addr, err)
-	}
-	version, err := Retry(isemver.Version)(&bind.CallOpts{
-		Context: ctx,
-	})
+// getBytecodeHash will get the hash of the bytecode of a contract at a given address.
+func getBytecodeHash(ctx context.Context, addr common.Address, client *ethclient.Client) (string, error) {
+	code, err := client.CodeAt(ctx, addr, nil)
```
geoknee, line 105: We need to change this so that _if_ the contract is proxied, we are getting the code of the _implementation_ contract, not the proxy.

geoknee: You can get the address of the proxy admin like this: 
```go
	Addresses[chain.ChainID].ProxyAdmin 
```

vdamle:
Closing this PR as https://github.com/ethereum-optimism/superchain-registry/pull/490 is meant to address this.

Issue #429, opened 2024-08-01
rename plasma --> alt-da


bitwiseguy:
The name `plasma` has been deprecated and we should use `alt-da` instead. This means we should replace all references to `plasma` across the superchain-registry, monorepo, op-node, etc. 

Since `alt-da` is still a beta feature, we can break backwards compatibility, but should work with devrel to come up with a way to migrate chains that are already using `plasma`

PR #439, opened 2024-08-06
Rename plasma to altDA

bitwiseguy:
Fixes https://github.com/ethereum-optimism/superchain-registry/issues/429


.editorconfig
```diff
@@ -5,10 +5,14 @@ root = true
 charset = utf-8
 end_of_line = lf
 indent_size = 2
-indent_style = space
+indent_style = tab
```
refcell, line 8: Can we keep this as spaces?

refcell (approved):
Rust changes look good.

.editorconfig
bitwiseguy: The justfile requires tab otherwise the recipes are considered invalid. I can just add a separate section for the justfile and then change the default back to space

bitwiseguy: Done here https://github.com/ethereum-optimism/superchain-registry/pull/439/commits/05657800bd6f2964e71971aed6a350727880a12d

add-chain/cmd/check-rollup-config.go
```diff
@@ -43,6 +43,7 @@ var CheckRollupConfigCmd = cli.Command{
 		localRollupConfig.ProtocolVersionsAddress = common.Address{}
 		localRollupConfig.ChannelTimeoutGranite = 50
 		opNodeRollupConfig.ProtocolVersionsAddress = common.Address{}
+		localRollupConfig.ChannelTimeoutGranite = 50
```
geoknee, line 46: This doesn't look quite right?

justfile
```diff
@@ -28,11 +28,12 @@ test-all: test-add-chain test-superchain test-validation
 test-add-chain:
 	# We separate the first test from the rest because it generates artifacts
 	# Which need to exist before the remaining tests run.
-	TEST_DIRECTORY=./add-chain go run gotest.tools/gotestsum@latest --format testname -- -run TestAddChain_Main
-	TEST_DIRECTORY=./add-chain go run gotest.tools/gotestsum@latest --format testname -- -run '[^TestAddChain_Main]'
+	TEST_DIRECTORY=./add-chain go run gotest.tools/gotestsum@latest --format testname -- -count=1 -run TestAddChain_Main
+	TEST_DIRECTORY=./add-chain go run gotest.tools/gotestsum@latest --format testname -- -count=1 -run '[^TestAddChain_Main]'
```
geoknee, line 32: Out of scope for this PR: We should ensure any validation check with external dependencies uses `-count=1` too (to avoid caching of test results).

justfile
```diff
@@ -28,11 +28,12 @@ test-all: test-add-chain test-superchain test-validation
...
 
 # Test all Go code in the superchain module
 test-superchain:
+	just clean-add-chain
```
geoknee, line 36: We are using this syntax elsewhere to declare one just command as a dependency in the definition of another:
```suggestion
test-superchain: clean-add-chain
```

geoknee (commented):
The rollup config diff tool is failing, and nowadays that failure is not expected/acceptable in general.

I think the problem is that the tool cannot generate configs on `main`, since the tool only checks out the `superchain` folder from `main` and retains the rest of the code from the PR branch.  That might be an easy fix -- @bitwiseguy do you think you could look into it?

geoknee (approved):

PR #447, opened 2024-08-07
update diff tools to not fail, but post/update a comment instead

geoknee:
Tested out here #443.

Towards #284


.github/workflows/genesis-diff.yml
```diff
@@ -33,7 +33,48 @@ jobs:
...
+      uses: actions/github-script@v6
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        script: |
+          const issue_number = context.payload.pull_request.number;
+          const run_id = context.runId;
+          const repo = context.repo.repo;
+          const owner = context.repo.owner;
+          const run_url = `https://github.com/${owner}/${repo}/actions/runs/${run_id}`;
+          const body = `Detected nonzero diff in genesis. This PR modifies the genesis data of at least one chain in the registry, which may or may not have been intended. It could affect whether a node started up with this data can successfully sync with the wider network. You can see the detail [here](${run_url}).`;
```
sebastianst, line 52:
```suggestion
          const body = `Detected nonzero diff in genesis. This PR modifies the genesis data of at least one chain in the registry, which may or may not have been intended. You can see the detail [here](${run_url}).`;
```
I would remove that one sentence, a consumer of this bot message probably doesn't need an explanation here, so better to keep it short.

.github/workflows/rollup-config-diff.yml
```diff
@@ -33,7 +33,48 @@ jobs:
...
       run: |
         ./scripts/diff-configs.sh ./generate-rollup-config/output-base ./generate-rollup-config/output-HEAD
+      continue-on-error: true
 
+    - name: Check command status and comment on PR
+      if: steps.run_command.outcome == 'failure'
+      uses: actions/github-script@v6
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        script: |
```
sebastianst, line 46: this looks like the same script as above. is there a way to deduplicate?

.github/workflows/genesis-diff.yml
geoknee: [b92e4ef](https://github.com/ethereum-optimism/superchain-registry/pull/447/commits/b92e4ef81b6002f51cdc20f1140fa6553b931f8b)

.github/workflows/rollup-config-diff.yml
geoknee: Yes, I made a reusable action to deduplicate the code [3d49b7f](https://github.com/ethereum-optimism/superchain-registry/pull/447/commits/3d49b7f8c232c0c07820d2bb1e77662eddf1ad49)

sebastianst (approved):

.github/workflows/rollup-config-diff.yml
sebastianst: nice

PR #451, opened 2024-08-08
rename `documentation` to `docs`

tessr:
Removes `docs/` from the gitignore so we can use that as the directory for our documentation.


geoknee (approved):
Looks good. One thing is that it seems we have no link into the glossary from the top level readmes?

tessr:
Ah yes, good call -- I'll add a link (though I'm sure inquiring minds will be able to find it regardless).

Issue #458, opened 2024-08-09
Do not use overrides in `LoadOPStackRollupConfig`


geoknee:
All information should be stored in and read from the superchain registry. 

* update the op-node `LoadOPStackRollupConfig` function
* update the add-chain module to capture the data
* possibly update existing config toml files to backport missing data

We can lean on the rollup config diff checker tool to help guide us with this (we should not have any net diff in the configs being exported, I think). 

My preference would be to keep `LoadOPStackRollupConfig` very transparent and backport any necessary data into the individual config files in the registry.

bitwiseguy:
Completed via [this monorepo pr](https://github.com/ethereum-optimism/optimism/pull/11459) and [this superchain-registry pr](https://github.com/ethereum-optimism/superchain-registry/pull/469).

PR #459, opened 2024-08-09
superchain: add MaxSequencerDrift, ChannelTimeout fields to config

bitwiseguy:
## Description
The following fields were previously not included in the .toml config files:
* `MaxSequencerDrift`
* `ChannelTimeoutBedrock`
* `ChannelTimeoutGranite`

This meant we relied on the hardcoded values in op-node to configure these params at runtime. We would like to allow chain operators some flexibility to configure these fields when adding their chain to the registry.

This PR adds the default values to all existing chain configs:
* `max_sequencer_drift = 600`
* `channel_timeout_bedrock = 300`
* `channel_timeout_granite = nil`

## Additional Context
This pr also cleans up the add-chain code so we unmarshal into the same struct instance that we end up writing to the .toml files. This reduces complexity of copying fields from one struct to another.

Complementary monorepo pr: https://github.com/ethereum-optimism/optimism/pull/11427

Fixes https://github.com/ethereum-optimism/superchain-registry/issues/458


add-chain/testdata/superchain/configs/sepolia/expected_zorasep.toml
```diff
@@ -8,7 +8,9 @@ standard_chain_candidate = true # This is a temporary field which causes most of
 batch_inbox_addr = "0xCd734290E4bd0200dAC631c7D4b9E8a33234e91f"
 block_time = 2
 seq_window_size = 3600
+max_sequencer_drift = 600
 data_availability_type = "eth-da"
+channel_timeout_bedrock = 300
```
vdamle, line 13: out of curiosity, why are we not including granite channel timeout here and in others configs that follow?

superchain/superchain.go
```diff
@@ -102,8 +102,12 @@ type ChainConfig struct {
 
 	BlockTime            uint64           `toml:"block_time" json:"block_time"`
 	SequencerWindowSize  uint64           `toml:"seq_window_size" json:"seq_window_size"`
+	MaxSequencerDrift    uint64           `toml:"max_sequencer_drift" json:"max_sequencer_drift"`
 	DataAvailabilityType DataAvailability `toml:"data_availability_type"`
 
+	ChannelTimeoutBedrock *uint64 `toml:"channel_timeout_bedrock" json:"channel_timeout"`
```
geoknee, line 108: Did you intend to use different keys for toml and json here?

superchain/configs/sepolia-dev-0/base-devnet-0.toml
```diff
@@ -10,9 +10,12 @@ canyon_time = 1698436800 # Fri 27 Oct 2023 20:00:00 UTC
 delta_time = 1706555000 # Mon 29 Jan 2024 19:03:20 UTC
 ecotone_time = 1706634000 # Tue 30 Jan 2024 17:00:00 UTC
 fjord_time = 1715961600 # Fri 17 May 2024 16:00:00 UTC
+granite_time = 1723046400 # Wed 7 Aug 2024 16:00:00 UTC
```
geoknee, line 13: I think this is probably an unintended change. The technique of overwriting the toml using the Go bindings (themselves mostly the result of unmarshalling the toml) has the effect of picking up hardfork times from superchain.toml and dumping them in here.

geoknee: Same for a few other files, too.

add-chain/cmd/update-configs.go
```diff
@@ -25,6 +25,10 @@ var UpdateConfigsCmd = cli.Command{
 				panic(err)
 			}
 
+			chain.MaxSequencerDrift = 600
+			fifty := uint64(50)
+			chain.ChannelTimeoutBedrock = &fifty
```
geoknee: See below comment about use of this tool. 

Also general question about this tool: do we think it is useful enough to keep in the repo? Seems more like one-time maintenance code to me so might be a good candidate for removal.

superchain/superchain.go
bitwiseguy: Yes. The `rollup.json` uses a different tag for the same field. See here: https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/types.go#L85

add-chain/testdata/superchain/configs/sepolia/expected_zorasep.toml
bitwiseguy: For chains that haven't activated the `granite` hardfork yet, the `channel_timeout_granite` setting is not yet applicable. In an effort to keep the config files clean, we typically exclude fields that are not applicable. If/when `granite` is activated for a given chain, we should add that field to the chain's config file.

superchain/configs/sepolia-dev-0/base-devnet-0.toml
bitwiseguy: Are you saying the `granite_time` should not be added at this time? Or it will never be added? IMO its better to explicitly set the timestamp in a chain's config file instead of relying on downstream software to configure the chain for you. Its more difficult to understand when a chain's hardfork will activate (or was activated) if its not part of its config file

add-chain/cmd/update-configs.go
bitwiseguy: I've used this tool at least three times in the past few weeks to add new fields or update existing fields. Its much easier and more scalable to use the tool than manually adding/editing fields in every chain config in the registry. Maybe we can just improve the tool, and I can remove this diff to the tool in this pr, instead of removing it completely? Or is there a different way you have in mind to apply new settings to many chains at once?

bitwiseguy:
Closing in favor of https://github.com/ethereum-optimism/superchain-registry/pull/469 since this one needed lots of reverted work. Was easier to start with a fresh pr

PR #453, opened 2024-08-08
Introduce `validate-modified-chains` `just` command

geoknee:
* run as separate step in CI (ultimately this should become a separate job)
* remove validation suite from unit tests command

This enables us to avoid validation failures (due to external factors) for chain A from blocking unrelated PRs (such as adding chain B, or updating the readme). 

This was tested on https://github.com/ethereum-optimism/superchain-registry/pull/452, where you can see only the modified chain got validated. 

Towards #428

Issue #428, opened 2024-08-01
ci: reroute validation errors/alerts caused by 3rd party components


bitwiseguy:
Currently the validation checks run as part of CI jobs that block PRs from being merged if there are failures. Also, the hourly CI job runs these same checks and reports failures to the `#notify-ci-failures` channel, which is generally used to detect failures of `ethereum-optimism` code, not external 3rd party software. Open PRs should not be blocked by issues with chains that are irrelevant to the PR.

We should make the following changes:
1. Create a new, independent validation CI job that runs every hour and routes its notifications to a new slack channel
2. Adjust the current "open PR" CI job so that it only runs validation checks on chains whose config changes as part of that PR (i.e. a new chain added, or existing chain modified). We should be able to run `git diff` to detect which chains have been changed, then leverage the `just validate <chain-id>` command to validation only those chains.

Separate standard-candidate chain validation checks:
* Set of validation checks they must pass to join the registry - should run in CI and block pr merge if there are failures
* Separate set of validation checks that inform what they need to fix before becoming standard
  * perhaps manually triggered github action

geoknee:
We also only need to run genesis checks as a one off when the chain is added, there isn't much point running drift detection on this.

geoknee:
Design doc in progress here https://www.notion.so/oplabs/Superchain-Registry-Check-Routing-0f4f734c159a48c3abfc0d6a82f07faf

geoknee:
An early step towards closing this ticket would be:
* add a `just ` command for running _only_ unit tests. This can use a regex to filter out validation checks
* change the CI config so that unit tests run in a different step to validation checks

geoknee:
A subsequent step would be to:
* detect which chains were modified using `git diff`
* only run validation checks for those chains

PR #464, opened 2024-08-12
CI: New job to run periodic chain validation and minor refactoring

vdamle:
**Description**

* Add a new job and slack notification target to run validation of all chains on a periodic basis.
* Move the step to validate modified chains into it's own job and add that job to the `pr-checks` list.
* Alter existing cron job to run this new job instead of golang-test, which runs all unit tests.

**Tests**

No new tests are being added/removed in this change.

**Additional context**

This is a follow up to the changes introduced in https://github.com/ethereum-optimism/superchain-registry/pull/453.

**Metadata**

[Reference issue](https://github.com/ethereum-optimism/superchain-registry/issues/428)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ commands:
     parameters:
       channel:
         type: string
-        default: C03N11M0BBN
+        default: C03N11M0BBN # to slack channel `notify-ci-failures`
     steps:
       - slack/notify:
           channel: << parameters.channel >>
@@ -24,7 +24,7 @@ commands:
     parameters:
       channel:
         type: string
-        default: C03N11M0BBN # TODO - modify this to a new slack channel
+        default: C07GQQZDW1G # to slack channel `notify-superchain-validation-failures`
     steps:
       - slack/notify:
           channel: << parameters.channel >>
@@ -124,9 +124,6 @@ jobs:
     steps:
       - checkout
       - install-just
-      - run:
-          name: Install gotestsum
-          command: go install gotest.tools/gotestsum@latest
       - install-foundry
       - run:
           name: run validation checks
@@ -140,9 +137,6 @@ jobs:
     steps:
       - checkout
       - install-just
-      - run:
-          name: Install gotestsum
-          command: go install gotest.tools/gotestsum@latest
       - install-foundry
       - run:
           name: run validation checks on all chains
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -19,6 +19,18 @@ commands:
           event: fail
           template: basic_fail_1
           branch_pattern: main
+  notify-failures-on-validate-all:
+    description: "Notify Slack on any chain validation failures"
+    parameters:
+      channel:
+        type: string
+        default: C03N11M0BBN # TODO - modify this to a new slack channel
```
vdamle, line 27: TODO - **change the slack channel id once a new one is created**

.circleci/config.yml
```diff
@@ -170,7 +215,7 @@ jobs:
 workflows:
   hourly:
     jobs:
-      - golang-test:
+      - golang-validate-all:
```
bitwiseguy, line 218: I think we'll probably want an additional hourly job, instead of replacing the existing one:
1. We can leave the existing one that runs `golang-test` and sends failures to `notify-ci-failures` (which is how we get notifications if something is wrong with our code)
2. Create the new `golang-validate-all` job, as you have done here, which sends failures to `notify-chain-validate-failures`, indicating there is an issue with an external chain

.circleci/config.yml
```diff
@@ -102,10 +118,39 @@ jobs:
...
+    shell: /bin/bash -eo pipefail
+    executor:
+      name: go/default  # is based on cimg/go
+      tag: '1.21'
+    steps:
+      - checkout
+      - install-just
+      - run:
+          name: Install gotestsum
+          command: go install gotest.tools/gotestsum@latest
```
geoknee, line 132: This can probably be removed #456

.circleci/config.yml
geoknee: (Optional suggestion) I wonder if it is worth factoring this out into a reusable step, taking the slack channel ID as a parameter (since that is the only difference between this step and the `notify-failues-on-main` step.

.circleci/config.yml
vdamle: Sounds good, i've now got both `golang-test` and `golang-validate-all` to run hourly, with notifications going to separate slack channels.

.circleci/config.yml
vdamle: cleaned up all invocations of `Install gotestsum`

geoknee (approved):

PR #468, opened 2024-08-13
Introduce TestPromotion

### LINKED ISSUE #467: report full standard validation results for standard candidate chains to new slack channel
These are useful to get from time to time, for an update on the question “what blocks this chain from becoming standard”?

Implementation sketch:
* get a list of chains which are standard_candidates (inspect chain.StandardChainCandidate)
* just promote-to-standard chainID, or you could [duplicate the Go code which does that](https://github.com/ethereum-optimism/superchain-registry/blob/0cebefcddd72dc646d0c919e3983b6e65726a8af/add-chain/cmd/promote-to-standard.go#L27-L31)
* invoke the full set of standard validation checks
* report to new slack channel on failure (but we need a way to separate / indicate this is a special failure, maybe a new slack channel would be fine)

geoknee:
- Add a new command and test to invoke promotion checks on promotion candidates
- Add a nightly job to run this test on all promotion candidates and send a notification of failures to slack.
- minor refactoring to clean up circle CI config.

Closes #467


geoknee:
It's critical that this test is only run in its own isolated test invocation, because it mutates the chain metadata on the fly. We could use [`TestMain` ](https://medium.com/goingogo/why-use-testmain-for-testing-in-go-dafb52b406bc)to get better control over when tests run, if that makes sense.

.circleci/config.yml
```diff
@@ -7,24 +7,11 @@ orbs:
   rust: circleci/rust@1.6.1
 
 commands:
-  notify-failures-on-main:
+  notify-failures-to-slack:
```
geoknee, line 10: Nit: shall we still have "on main" in the name, since that is a hardcoded branch pattern?

validation/security-configs_test.go
```diff
@@ -45,7 +45,8 @@ var checkResolutions = func(t *testing.T, r standard.Resolutions, chainID uint64
 			got, err := getAddress(method, contractAddress, client)
 			require.NoErrorf(t, err, "problem calling %s.%s (%s)", contract, method, contractAddress)
 
-			assert.Equal(t, want, got, "%s.%s = %s, expected %s (%s)", contract, method, got, want, output)
+			// Use assert.True here for a concise output of failures, since failure info is sent to a slack channel
+			assert.True(t, want == got, "%s.%s = %s, expected %s (%s)", contract, method, got, want, output)
```
geoknee, line 49: Do you think we could roll this out to any other locations where we are comparing byte slices?

I wonder if it might actually be just as well to do 

```go
if want != got {
  t.Errorf("%s.%s = %s, expected %s (%s)", contract, method, got, want, output)
}
```
This will give us the most useful output only.

validation/promotion_test.go
```diff
@@ -0,0 +1,24 @@
...
+	. "github.com/ethereum-optimism/superchain-registry/superchain"
+)
+
+func TestPromotion(t *testing.T) {
+	for _, chain := range OPChains {
+		chain := chain
+		t.Run(perChainTestName(chain), func(t *testing.T) {
+			t.Parallel()
+			if chain.StandardChainCandidate {
+				chain.StandardChainCandidate = false
```
bitwiseguy, line 16: What's the purpose of this line? Might be good to add a comment to this function that explains what the function does, which may explain this change here

geoknee:
```suggestion
         // promote the chain to standard 
         // by mutating the chainConfig
				chain.StandardChainCandidate = false
```

.circleci/config.yml
```diff
@@ -140,7 +128,22 @@ jobs:
...
+      tag: '1.21'
+    steps:
+      - checkout
+      - install-just
+      - install-foundry
+      - run:
+          name: run validation checks
+          command: just promotion-test
+      - notify-failures-on-main:
+          channel: C07GZVCCUS0 # to slack channel `notify-superchain-promotion-test-failures`
```
bitwiseguy, line 146: Nit: could probably change the channel name to `notify-superchain-promotion-failures` (i.e. drop `test`) to reduce length and match naming convention of `notify-superchain-validation-failures`

bitwiseguy (approved):
Approved with a few minor comments

.circleci/config.yml
vdamle: restored.

validation/security-configs_test.go
vdamle: Done, as discussed.

.circleci/config.yml
vdamle: Good catch! Updated channel name and comment.

PR #474, opened 2024-08-13
Teeny README updates

pifafu:
# Changes
This PR updates the README with a banner and makes some small edits to copy.

### **[View file](https://github.com/ethereum-optimism/superchain-registry/blob/546854a97e26ca4f8acfa4fb753b488e09483e6e/README.md)**.

---

| Dark mode | Light mode
:---|:---
CleanShot 2024-08-14 at 11 56 35@2x | CleanShot 2024-08-14 at 11 56 48@2x


README.md
```diff
@@ -1,4 +1,4 @@
-# superchain-registry
+<img width="3200" alt="Frame 2 (5)" src="https://github.com/user-attachments/assets/088a3923-762f-41fb-bd0a-975dea417e7f"><br/>
```
protolambda, line 1: Can the file be put into the `.github` directory and referenced that way? The `github.com/user-attachments` asset link might break. Also updating the `alt` image text while here would be nice.

protolambda (commented):
LGTM, although

README.md
sebastianst: That would indeed be nice to fix! Otherwise great banner ✨🔴⏡🔴✨

pifafu: Done! https://github.com/ethereum-optimism/superchain-registry/pull/474/commits/775076127721b6551eb31cfe268ec61382739710

pifafu:
I've made my final updates. Updated the asset so that it still works in gh dark mode and uploaded to .github:

| Dark mode | Light mode
:---|:---
CleanShot 2024-08-14 at 11 56 35@2x | CleanShot 2024-08-14 at 11 56 48@2x

protolambda (approved):
Approving to unblock this PR. The image / text tweaks are quite nice.

However, if I can suggest one last change, it would be nice if instead of a 4MB image we use a compressed lower-resolution version of the asset. GitHub does not render the readme header to be that big anyway. And now the page loads slow because of the large image.

pifafu:
@protolambda Good call, I've reduced it to 421 KB!

Issue #430, opened 2024-08-01
superchain_level should match decentralization stage


bitwiseguy:
Currently we have the following categories:
* `superchain_level` = 1 : stage 0
* `superchain_level` =2 : stage 1

We should decrement the `superchain_level` by 1 so that they match, and we avoid confusion.

PR #480, opened 2024-08-15
repo: decrement superchain level

bitwiseguy:
Request from pm to decrement the superchain_level for all chains such that it more closely aligns with the decentralization level used by l2beat

Fixes https://github.com/ethereum-optimism/superchain-registry/issues/430


superchain/superchain.go
```diff
@@ -60,8 +60,8 @@ type HardForkConfiguration struct {
 type SuperchainLevel uint
 
 const (
-	Standard SuperchainLevel = 2
-	Frontier SuperchainLevel = 1
+	Standard SuperchainLevel = 1
+	Frontier SuperchainLevel = 0
```
geoknee, line 64: We did well here to define these constants, this diff is much smaller than it would need to be if we had sprinkled `0`s and `1`s everywhere!

geoknee (approved):

refcell (approved):

PR #484, opened 2024-08-16
wip: updates to genesis predeploy validation

bitwiseguy:


validation/genesis/genesis-predeploy_test.go
```diff
@@ -145,100 +151,107 @@ func testGenesisPredeploys(t *testing.T, chain *ChainConfig) {
...
+	var cd *ContractData
+	if ok {
+		// it is a predeploy, we need to perform masking in order to validate it
+		filePath := path.Join(contractsDir, "forge-artifacts", artifactName+".sol", artifactName+".json")
+		data, err := os.ReadFile(filePath)
+		// Check if the error is due to the file not existing
+		if err != nil {
+			if os.IsNotExist(err) {
+				t.Logf("File not found: %s", filePath)
+				return
```
bitwiseguy, line 191: I think this is ok because the set of hardcoded `predeployArtifactNames` probably doesn't apply to all chains, especially ones created long ago. But would be good to verify this. I was getting errors about `EAS.json` and `SchemaRegistry.json` files not existing. 

Also in the `monorepo-outputs.sh` script we are always generating outputs based on mainnet. Should we have an arg so it uses the op sepolia deploy-config if we are validating a sepolia chain?

geoknee (approved):

validation/genesis/genesis-predeploy_test.go
geoknee: It might be that these contracts got moved into a legacy folder or something: if they are present in the synthetic genesis file, I would expect the source files + deployment artifacts to exit somewhere...

It might be that the address is just getting 1 wei allocated or something, and no actual code. So we can definitely have an escape hatch if there is empty bytecode in the synthetic genesis.

We should add a comment here for now and we can circle back and check.

bitwiseguy: I put an escape hatch [here](https://github.com/ethereum-optimism/superchain-registry/pull/484/files#diff-0951f9e907e0bc03da7d9405099e27b8b3b006040f37ef12112c1cd0a7ff6b42R173-R176) for the case where the code hash is zero. Not sure if that fully covers it though

bitwiseguy: Also, should [this line](https://github.com/ethereum-optimism/superchain-registry/pull/484/files#diff-0951f9e907e0bc03da7d9405099e27b8b3b006040f37ef12112c1cd0a7ff6b42R130) always use a synthetic version of `mainnet`? Or should it use a synthetic version of `sepolia` if we are validating a `sepolia` chain?

validation/genesis/monorepo-outputs.sh
```diff
@@ -10,6 +10,7 @@ go_version=$(grep -m 1 '^go ' ${monorepo_dir}/go.mod | awk '{print $2}')
 
 # Source the gvm script to load gvm functions into the shell
 set +e
+source ~/.gvm/scripts/gvm || exit 1
```
bitwiseguy, line 13: I had to put this line back in to get it to run locally

validation/genesis/genesis-predeploy_test.go
geoknee: > Also, should [this line](https://github.com/ethereum-optimism/superchain-registry/pull/484/files#diff-0951f9e907e0bc03da7d9405099e27b8b3b006040f37ef12112c1cd0a7ff6b42R130) always use a synthetic version of `mainnet`? Or should it use a synthetic version of `sepolia` if we are validating a `sepolia` chain?

We could try that, I'm not sure if it will make much difference?

justfile
```diff
@@ -46,11 +46,11 @@ validate-modified-chains REF:
 
 # Run validation checks for chains with a name or chain ID matching the supplied regex, example: just validate 10
 validate CHAIN_ID:
-	TEST_DIRECTORY=./validation go run gotest.tools/gotestsum@latest --format testname -- -run=TestValidation/{{CHAIN_ID}} -count=1
+	TEST_DIRECTORY=./validation go run gotest.tools/gotestsum@latest --format testname -- -run='TestValidation/.+\({{CHAIN_ID}}\)$' -count=1
 
 # Run genesis validation (this is separated from other validation checks, because it is not a part of drift detection)
 validate-genesis CHAIN_ID:
-	TEST_DIRECTORY=./validation/genesis go run gotest.tools/gotestsum@latest --format testname -- -run=TestGenesisPredeploys/{{CHAIN_ID}} -v
+	TEST_DIRECTORY=./validation/genesis go run gotest.tools/gotestsum@latest --format testname -- -run='TestGenesisPredeploys/.+\({{CHAIN_ID}}\)$' -v -count=1
```
geoknee, line 53: Let's get these changes into `main` (along with the accompanying changes in the circleci config).

PR #485, opened 2024-08-19
Try validating worldcoin sepolia

geoknee:
Produces an almost perfect match
Screenshot 2024-08-19 at 10 53 12

This diff is outside of `allocs`, so doesn't really block genesis predeploy validation. Filed this issue to track the problem, we do need to fix it  #487


.circleci/config.yml
```diff
@@ -12,19 +12,6 @@ commands:
...
-          channel: << parameters.channel >>
-          event: fail
-          template: basic_fail_1
-          branch_pattern: main
-  notify-failures-on-validate-all:
-    description: "Notify Slack on any chain validation failures"
-    parameters:
-      channel:
-        type: string
-        default: C07GQQZDW1G # to slack channel `notify-superchain-validation-failures`
```
geoknee, line 27: These changes were unintended and should be fixed.

Issue #478, opened 2024-08-14
Capture genesis creation metadata into the registry


geoknee:
- `deploy-config.json`
- monorepo commit hash where genesis command was invoked
- command invocation and parameters for the genesis generation command

So that we can validate the genesis for standard chains.

We may as well store it for all chains, in case they want to become standard in the future.

geoknee:
Proposed changes:

## `optimism` monorepo
The genesis creation script should store a new file `genesis-creation-metadata.json` (or, could we just a new field to the `genesis.json` file?)
 Inside, it lists:
*  `commitHash: "abcde"`.
* `invocationCommand: ["go", "run", "cmd/main.go", "genesis l2"]` 
(or the forge route like from this doc https://docs.optimism.io/builders/chain-operators/deploy/genesis)
* we already have `deploy-config.json` file hanging around

## `superchain-registry`
* add a mechanism where we pull in the commitHash and invocation command from above and store it in the chain config
* need to support legacy chains who didn't have the benefit of the changes above. They will need to be able to specify the missing data manually, most likely in the `.env` file.

bitwiseguy:
If we store the monorepo `commitHash` in the superchain-registry, will it be able to derive all the other information it needs? Just curious how the `invocationCommand` and `deploy-config.json` information would help create the compiled bytecode needed for the validation check.

From the superchain-registry side, this is how I imagine the bytecode recreation would work
1. Fetch the monorepo `commitHash` that we will store in superchain-registry (in the chain config .toml?)
2. Use the `commitHash` to checkout that specific version of the monorepo
3. Run `forge build` at that point in the monorepo
4. Fetch the contract bytecode from the forge-artifacts json file for each contract
5. Compare that bytecode to what is onchain to ensure they match

Where/how does the `invocationCommand` and `deploy-config.json` come into play?

geoknee:
Let me try to clarify this a bit. I think there are the following two approaches to genesis validation:

## "compile & ignore immutables"
This is exactly as you describe in your steps 1-5. I think you are correct that an invocationCommand is not required for this approach (although, we still ideally would have the necessary commands to install dependencies and compile, in case `pnpm install && forge build` does not work for older commits). We also do not need the `deploy-config.json` for this approach.

This is the initial approach we are taking on #475 .

The downsides are: we ignore immutable variables and have to inspect account balances, maintain the list of standard predeploys and storage variables in a more ad hoc fashion too.

## regenerate genesis exactly
[...]

PR #489, opened 2024-08-19
[ Feature Branch ] Genesis allocs validation

### LINKED ISSUE #372: validtaion: check genesis predeploys against standard config

### LINKED ISSUE #486: Pretty print the diff when a chain fails `Genesis.allocs` validation
If we end up implementing the validation via a simple bytes / string comparison only, then we want to avoid printing obscure / unreadable diffs to the console.

geoknee:
* Modifies `add-chain` flow to store sufficient metadata to validate the allocs in a chain's genesis file
* Adds a `genesis` package under the `validation` module, which contains a test to perform full genesis allocs validation
  * It does this by cloning the monorepo, checking out the declared commit, building and generating the l2 genesis using the deploy-config as input. The test can handle multiple different node versions, go versions, and genesis command invocations
  * This full check is only triggered if the validation metadata for a given chain is modified
  * The check is **seperate** from the other validation checks, so has its own triggers and conditions
  * For each chain modified in this way, we generate a dynamic CI configuration to run the validations concurrently, and then report a single "fan-in" result requiring all to pass
* Adds a "lite" check for each chain, run as a part of the regular validation suite, which requires the chain has a "genesis creation commit" stored for it. 
* Adds validation metadata for several chains already in the registry, as well as some not yet in the registry (this allows us to save that metadata so we can use it later). 

Closes #486 
Closes #372 
Towards #478


.circleci/config.yml
```diff
@@ -60,6 +60,22 @@ commands:
...
+            apt-get -yq install curl git mercurial make binutils bison gcc build-essential bsdmainutils
+            bash < <(curl -s -S -L https://raw.githubusercontent.com/moovweb/gvm/master/binscripts/gvm-installer)
+            source /root/.gvm/scripts/gvm
+  validate-genesis:
+    description: "Runs genesis validation checks"
+    steps:
+      - run:
+          name: Clone monorepo
+          command: mkdir ../optimism-temporary && git clone https://github.com/ethereum-optimism/optimism.git ../optimism-temporary
+      - run: just validate-genesis ""
```
geoknee, line 78:
```suggestion
      - run: just validate-genesis "*"
```

validation/genesis/genesis-predeploy_test.go
```diff
@@ -0,0 +1,195 @@
...
+	if false {
+		executeCommandInDir(t, monorepoDir, exec.Command("rm", "-rf", "node_modules"))
+		executeCommandInDir(t, contractsDir, exec.Command("rm", "-rf", "node_modules"))
+	}
+
+	// install dependencies
+	// TODO we expect this step to vary as we scan through the monorepo history
+	// so we will need some branching logic here
+	// executeCommandInDir(t, contractsDir, exec.Command("pnpm", "install", "--no-frozen-lockfile"))
+	executeCommandInDir(t, contractsDir, exec.Command("yarn", "install", "--no-frozen-lockfile"))
```
geoknee, line 77: We may need to use `nvm` depending on the version of node we need to build the monorepo.

validation/genesis/validation-inputs/34443/meta.toml
```diff
@@ -0,0 +1,13 @@
+genesis_creation_commit = "3eda4cd594ba6409584eb6ef95c341d98419e392"
+genesis_creation_command = [
```
geoknee, line 2: We can probably store this as a single string.

validation/genesis/monorepo-outputs.sh
```diff
@@ -0,0 +1,25 @@
+set -e
+
+echo "Inferring and selecting correct Node version"
+export NVM_DIR="$([ -z "${XDG_CONFIG_HOME-}" ] && printf %s "${HOME}/.nvm" || printf %s "${XDG_CONFIG_HOME}/nvm")"
+[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh" # This loads nvm
+nvm use
+
+echo "Running install command"
+eval $1
```
bitwiseguy, line 9: It seems risky to blindly execute a shell command that a user is passing in and has complete control over formatting. It might be better for the user to provide an enum-type value, which our code uses to select from a list of known, predefined commands to execute

validation/genesis/validation-inputs/34443/meta.toml
```diff
@@ -0,0 +1,14 @@
+genesis_creation_commit = "3eda4cd594ba6409584eb6ef95c341d98419e392"
+monorepo_build_command = "yarn install --no-frozen-lockfile"
+genesis_creation_command = """\
```
bitwiseguy, line 3: Perhaps this could be changed to `contracts_file_type` and the valid values are just `["single", "multiple"]` (indicating single .deploy file or one .json file for every contract). I think the other variables involved in this command like the `chainID` could be derived by the go code and passed into this shell script

validation/genesis/validation-inputs/10/meta.toml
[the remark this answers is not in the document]
geoknee: This is a very interesting suggestion. I'd love to get into the pros and cons, but I think we can probably proceed with this PR as is and revisit your idea.

validation/genesis/genesis-predeploy_test.go
```diff
@@ -0,0 +1,230 @@
...
+// REQUIREMENTS:
+// pnpm and yarn, so we can prepare https://codeload.github.com/Saw-mon-and-Natalie/clones-with-immutable-args/tar.gz/105efee1b9127ed7f6fedf139e1fc796ce8791f2
+func testGenesisPredeploys(t *testing.T, chain *ChainConfig) {
+	chainId := chain.ChainID
+
+	vis, ok := ValidationInputs[chainId]
+
+	if !ok {
+		t.Skip("WARNING: cannot yet validate this chain (no validation metadata)")
+	}
```
geoknee, line 50: TODO we should update this before merging the PR (the test should just fail if the metadata is not present).

validation/genesis/foundry-config.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/packages/contracts-bedrock/foundry.toml b/packages/contracts-bedrock/foundry.toml
```
geoknee, line 1: I don't think this is used anymore, it could be removed.

geoknee:
Action plan:

1. Merge the stack of PRs into this one. 
2. Rebase this PR
3. Confirm all known chains still validate correctly
4. Separate out any new chains being added on this PR, to get a streamlined version which only commits 
  i. the `add-chain` code
  ii. the validation code itselg
  iii. the validation input files (`meta.toml` and `deploy-config.toml`) 
 5. We save the new chain information onto another feature branch which we can continue to use to validate all of the chains we care about until they have been merged to `main` in the usual way

.circleci/continue_config.yml
```diff
@@ -0,0 +1,278 @@
+version: 2.1
```
geoknee, line 1: Could we make review easier by using `git mv`?

.circleci/continue_config.yml
```diff
@@ -0,0 +1,278 @@
...
+orbs:
+  go: circleci/go@1.8.0
+  node: circleci/node@5.2.0
+  slack: circleci/slack@4.10.1
+commands:
```
geoknee, line 6: Is there a way to break up this large file? Perhaps using orbs?

validation/genesis-allocs_test.go
```diff
@@ -0,0 +1,21 @@
...
+func testGenesisAllocsMetadata(t *testing.T, chain *ChainConfig) {
+	// This tests asserts that the a genesis creation commit is stored for
+	// the chain. It does not perform full genesis allocs validation.
+	// Full genesis allocs validation is run as a one-off requirement when
+	// chains are added to the registry and/or when the genesis creation commit itself
+	// is changed (and may be re-run at other choice moments).
+	// Therefore the presence of a genesis creation commit ensures that full genesis
+	// validation has been performed on this chain.
+	// This test is lightweight and can be run continually.
+	require.NotEmpty(t, genesis.ValidationInputs[chain.ChainID].GenesisCreationCommit)
```
geoknee, line 20: This tries to ensure that the information necessary for validating the genesis of the chain continues to exist over time.

add-chain/flags/flags.go
```diff
@@ -70,6 +70,18 @@ var (
...
+		EnvVars:  prefixEnvVars("DEPLOY_CONFIG"),
+		Usage:    "Filepath to deploy-config json input file",
+		Required: true,
+	}
+	GenesisCreationCommit = &cli.StringFlag{
+		Name:     "genesis-creation-commit",
+		EnvVars:  prefixEnvVars("GENESIS_CREATION_COMMIT"),
+		Usage:    "Commit in the https://github.com/ethereum-optimism/optimism/ repo at which the chain's genesis was created",
+		Required: true,
+	}
```
bitwiseguy, line 84: It would be a much better user experience, and less error-prone, if this commit was written into the deploy-config or rollup.json file. If we put the onus on the user to remember which commit they used, we are much more likely to have issues than if we automatically record that information via the tooling the user already needs to use (i.e. `op-node genesis l2` command)

add-chain/main.go
```diff
@@ -230,3 +261,32 @@ func getGasPayingToken(l1rpcURl string, SystemConfigAddress superchain.Address)
...
+}
+
+func writeGenesisValidationMetadata(commit string, targetDir string) error {
+	// Define default metadata params:
+	// These may not be sufficient to make the genesis validation work,
+	// but we address that with some manual trial-and-error intervention
+	// involving OPLabs engineers after the add-chain command runs.
+	const defaultNodeVersion = "18.12.1"
+	const defaultMonorepoBuildCommand = "pnpm"
+	const defaultGenesisCreationCommand = "opnode1"
```
bitwiseguy, line 280: Is there any explanation about what `opnode1` and `opnode2` mean?

add-chain/flags/flags.go
geoknee: Agreed that we should have the genesis creation tool output the commit. I'm not sure whether one of the existing files or a new one would be best.

A real small lift would be to print a message: "the commit you used just now might be important for future validation --- remember it!". Not an ideal solution, but maybe not a crazy stop gap until we get something better in place?

There are also plans afoot to approach this in a way which doesn't require such complex validation https://github.com/ethereum-optimism/specs/issues/350. So if that happens, the need to remember the commit would (probably?) go away.

add-chain/main.go
geoknee: I should add some. Essentially, the genesis creation command was previously entirely handled by an op-node subcommand, and there were (at least) two different APIs to that command (with respect to command line args). More recently, the command also involves using a `foundry` script, so I anticipate we may need to add a `foundry` option to this file.

add-chain/main.go
```diff
@@ -230,3 +261,32 @@ func getGasPayingToken(l1rpcURl string, SystemConfigAddress superchain.Address)
...
+	return os.WriteFile(path.Join(targetDir, "deploy-config.json"), data, os.ModePerm)
+}
+
+func writeGenesisValidationMetadata(commit string, targetDir string) error {
+	// Define default metadata params:
+	// These may not be sufficient to make the genesis validation work,
+	// but we address that with some manual trial-and-error intervention
+	// involving OPLabs engineers after the add-chain command runs.
+	const defaultNodeVersion = "18.12.1"
+	const defaultMonorepoBuildCommand = "pnpm"
```
bitwiseguy, line 279: Similar to the monorepo commit hash, could these two fields be derived and written into one of the pre-existing inputs to the `add-chain` tooling (i.e. the deploy-config or rollup.json file)? That way we are much more likely to be both backwards and forwards compatible, without requiring manual intervention

add-chain/e2e_test.go
```diff
@@ -172,7 +172,6 @@ func checkConfigTOML(t *testing.T, testName, chainShortName string) {
 func cleanupTestFiles(t *testing.T, chainShortName string) {
 	paths := []string{
 		addressDir + chainShortName + ".json",
-		genesisSystemConfigDir + chainShortName + ".json",
```
bitwiseguy, line 175: Should we remove the test artifacts written to `validationInputsDir` as well?

validation/genesis/genesis-allocs_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+	// Exit with the result of the tests
+	os.Exit(exitVal)
+}
+
+func TestGenesisAllocs(t *testing.T) {
+	for _, chain := range OPChains {
+		if chain.SuperchainLevel == Standard || chain.StandardChainCandidate {
+			t.Run(PerChainTestName(chain), func(t *testing.T) {
+				// Do not run in parallel
```
bitwiseguy, line 63: Might be good to explain why we don't want to run in parallel. Assuming its because we don't want to checkout a different version of the monorepo while another test is running

add-chain/flags/flags.go
vdamle: Not sure if I have suggested this in one of the discussions - the genesis creation tool could create a git tag with a "well known" format for the commit ex: `chainName-networkName-deploy`. Once that is done, it removes the need to remember the commit and any tools that run in the future could rely on the well known name to determine the previous point in time code.

.circleci/continue_config.yml
```diff
@@ -0,0 +1,278 @@
...
+      - install-gvm
+      - install-nvm
+      - run: just validate-genesis-allocs <<parameters.chainid>>
+  genesis-allocs-all-ok:
+    docker:
+      - image: cimg/go:1.22.6-node
+    resource_class: medium
+    steps:
+      - checkout
+      - run: echo "Job 1 is running"
```
vdamle, line 237: Is this a triage debug? if so, could we remove this?

geoknee: Yeah 100%

validation/genesis/config.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/op-chain-ops/genesis/config.go b/op-chain-ops/genesis/config.go
+index 4c813acd4..42aa795ed 100644
+--- a/op-chain-ops/genesis/config.go
++++ b/op-chain-ops/genesis/config.go
+@@ -588,7 +588,7 @@ func NewDeployConfig(path string) (*DeployConfig, error) {
+ 	}
+ 
+ 	dec := json.NewDecoder(bytes.NewReader(file))
+-	dec.DisallowUnknownFields()
++	// dec.DisallowUnknownFields()
```
vdamle, line 10: do we need to leave this in commented out?

.circleci/continue_config.yml
```diff
@@ -0,0 +1,278 @@
...
+      - install-just
+      - run:
+          name: Run codegen
+          command: just codegen
+      - run:
+          name: check git tree is clean
+          command: git diff --exit-code
+  golang-validate-genesis-allocs:
+    docker:
+      - image: cimg/go:1.22.6-node
```
bitwiseguy, line 218: Will using go v1.22.6 cause any issues since the go.mod/go.work files specify v1.21?

add-chain/e2e_test.go
geoknee: Ideally yes, it's a bit difficult because the cleanup function doesn't have the chain id in scope at the moment. Will take another look to see if I can refactor it.

validation/genesis/config.patch
geoknee: Nope, the patch could just do the deletion.

.circleci/continue_config.yml
geoknee: Good catch -- I don't think it will cause issues but I will adjust this if I can to make sure.

.circleci/continue_config.yml
geoknee: I tried rewriting the history this way, but GitHub does still not show a useful diff. Instead, the reviewer can check out this branch and run `git diff :.circleci/continue_config.yml main:./.circleci/config.yml`, this gives a pretty clean diff.

.circleci/continue_config.yml
geoknee: This looks more complex using circleci than it is with github actions, so I will leave this for future work.

add-chain/flags/flags.go
geoknee: @vdamle that is an interesting suggestion, I suppose one downside is that the monorepo needs to permit a large number of contributors to push git tags, and the list of tags will get pretty long. I think the plan of asking operators to launch off one of a small list of "released" commits would be a better solution, and also reduces the work it will take to validate the chain in future.

geoknee: Whatever approach we take in the future, we are stuck with the legacy behaviour now for quite a few chains -- so I don't think there is much to do on this PR to address the problem.

add-chain/main.go
geoknee: In principle yes, although I think we are more likely to explore more radical overhauls of the genesis creation path which will make this unnecessary. I'd be a bit nervous about modifying the existing deploy-config.json file, as it is one of the inputs to the genesis creation tool and it seems an antipattern to modify the input files.

bitwiseguy (approved):
🚀

PR #490, opened 2024-08-19
L1 contract bytecode validation

vdamle:
Validate contract bytecode hashes for proxy & non-proxy contracts.

* if proxy contract, fetch implementation address and the bytecode of the implementation contract.
* if not a proxy contract, get the bytecode of the address.
* Since fault proof based contracts use immutables, their bytecode will be different on-chain depending on instantiation parameters. There will be a follow-on enhancement to mask immutables' bytecode before comparing against what's stored in the superchain registry.

Fixes: https://github.com/ethereum-optimism/superchain-registry/issues/405

Supersedes https://github.com/ethereum-optimism/superchain-registry/pull/438


validation/standard/standard-versions.toml
```diff
@@ -7,8 +7,4 @@ optimism_mintable_erc20_factory = "1.9.0"
 optimism_portal = "3.10.0"
 system_config = "2.2.0"
 protocol_versions = "1.0.0"
-anchor_state_registry = "1.0.0"
-delayed_weth = "1.0.0"
-dispute_game_factory = "1.0.0"
-fault_dispute_game = "1.2.0"
```
geoknee, line 13: I think we can keep these?

validation/uniqueness_test.go
```diff
@@ -94,7 +94,10 @@ func getGlobalChains() (map[uint]*uniqueProperties, error) {
 		for i, url := range chain.RPC {
 			normalizedURL, err := normalizeURL(url)
 			if err != nil {
-				return nil, err
+				// one of the chains is now throwing the error
+				// parse "https://hk.p.bifrost-rpc.liebi.com, wss://hk.p.bifrost-rpc.liebi.com/ws": invalid character " " in host name
+				continue
+				//return nil, err
```
geoknee, line 100: Do we want to commit this workaround? I believe the underlying problem is resolved.

validation/superchain-version_test.go
```diff
@@ -107,6 +122,63 @@ func getContractVersionsFromChain(list AddressList, client *ethclient.Client) (C
 	return cv, nil
 }
 
+// getContractBytecodeHashesFromChain pulls the appropriate bytecode from chain
+// using the supplied client (calling the version() method for each contract). It does this concurrently.
```
geoknee, line 126: I think this comment needs updating (copypaste error).

validation/superchain-version_test.go
```diff
@@ -30,29 +56,18 @@ func testContractsMatchATag(t *testing.T, chain *ChainConfig) {
 
 	versions, err := getContractVersionsFromChain(*Addresses[chain.ChainID], client)
 	require.NoError(t, err)
-	_, err = findOPContractTag(versions)
+	_, err = findOPContractTagInVersions(versions)
+	require.NoError(t, err)
+
+	bytecodeHashes, err := getContractBytecodeHashesFromChain(chain.ChainID, *Addresses[chain.ChainID], client)
+	require.NoError(t, err)
+	_, err = findOPContractTagInByteCodeHashes(L1ContractBytecodeHashes(bytecodeHashes))
```
geoknee, line 64: Is this type conversion necessary?

validation/superchain-version_test.go
```diff
@@ -107,6 +122,63 @@ func getContractVersionsFromChain(list AddressList, client *ethclient.Client) (C
...
+		}
+		results.Store(contractName, r)
+		wg.Done()
+	}
+
+	wg := new(sync.WaitGroup)
+
+	for _, contractName := range contractsToCheck {
+		contractAddress, err := list.AddressFor(contractName)
+		//log.Printf("contract address: %s", contractAddress)
```
geoknee, line 146: ✂️

validation/superchain-version_test.go
```diff
@@ -123,6 +195,67 @@ func getVersion(ctx context.Context, addr common.Address, client *ethclient.Clie
...
+}
+
+// getBytecodeHash will get the hash of the bytecode of a contract
+//   - at a given address, if the contract is not a proxy contract
+//   - at the proxy implementation contract's address, if the contract is a proxy contract (we currently use the name suffix to determine
+//     whether the contract is a proxy or not)
+func getBytecodeHash(ctx context.Context, chainID uint64, contractName string, targetContractAddr common.Address, client *ethclient.Client) (string, error) {
+
+	addrToCheck := targetContractAddr
+	proxyContract := strings.Contains(strings.ToLower(contractName), "proxy")
```
geoknee, line 241: Nit: I think you might be able to use `strings.HasSuffix()` here.

validation/superchain-version_test.go
```diff
@@ -181,39 +314,71 @@ func findOPContractTag(versions ContractVersions) ([]standard.Tag, error) {
...
+	prettyStandard, err := json.MarshalIndent(standard.BytecodeHashes, "", " ")
+	if err != nil {
+		return matchingTags, err
+	}
 
-			if field.String() == "" {
-				// Ignore any empty strings, these are treated as "match anything"
-				continue
-			}
+	err = fmt.Errorf("contract versions %s do not match any standard op-contracts tag %s", pretty, prettyStandard)
```
geoknee, line 341: Should we state "bytecode hashes" instead of "contract versions" here?

validation/superchain-version_test.go
```diff
@@ -13,12 +14,37 @@ import (
...
+		"OptimismPortalProxy",
+		"SystemConfigProxy",
+		// "AnchorStateRegistryProxy",
+		// "DelayedWETHProxy",
+		// "DisputeGameFactoryProxy",
+		// "FaultDisputeGame",
+		// "MIPS",
+		"PermissionedDisputeGame",
+		"PreimageOracle",
+	}
```
geoknee, line 45: Longer term, a unified list of contracts to check the version and bytecode for makes sense. However, given that we are not yet tackling all the Fault Proof contracts I think it makes sense to keep two lists, or possibly better one "full" list, and then a list of `contractsToSkip` for the bytecode checks (with a comment and link to a github issue tracking the work that is still to be done).

geoknee (changes requested):

validation/uniqueness_test.go
vdamle: Dang, forgot to clean this up. will remove

validation/superchain-version_test.go
vdamle: Not needed, actually don't remember why I had to add it in the first place.

validation/superchain-version_test.go
vdamle: Ah, the `contractsToSkip` sounds like a good idea. will tweak.

validation/standard/standard-versions.toml
vdamle: There was an error with anchor state registry mismatch for OP sepolia. I see this [thread](https://oplabs-pbc.slack.com/archives/C073P71SGP2/p1723932925971519?thread_ts=1723932048.355599&cid=C073P71SGP2) where you've mentioned the right way to address this. I've added an exclusion for Sepolia in `Standard_Contract_Versions`.

validation/standard/standard-versions.toml
```diff
@@ -11,4 +11,4 @@ anchor_state_registry = "1.0.0"
 delayed_weth = "1.0.0"
 dispute_game_factory = "1.0.0"
 fault_dispute_game = "1.2.0"
-permissioned_dispute_game = "1.2.0"
+permissioned_dispute_game = "1.2.0"
```
geoknee, line 14: Nit: let's make sure the file ends with a newline.

validation/superchain-version_test.go
```diff
@@ -13,12 +14,45 @@ import (
...
+
+	contractsToSkip = []string{
+		"AnchorStateRegistryProxy",
+		"DelayedWETHProxy",
+		"DisputeGameFactoryProxy",
+		"FaultDisputeGame",
+		"MIPS",
+	}
```
geoknee, line 53: Ideally these contracts are declared to be skipped _only_ in the scope of the bytecode hash check, and not in the scope of the version check.

geoknee (approved):

validation/exclusions_test.go
```diff
@@ -61,6 +61,7 @@ var exclusions = map[string]map[uint64]bool{
 		999999999: true, // sepolia/zora         MCP (at time of writing)
 		11155421:  true, // sepolia-dev0/oplabs-devnet-0
 		11763072:  true, // sepolia-dev0/base-devnet-0
+		11155420:  true, // sepolia/op (contract versions upgraded with gov approval pending)
```
geoknee, line 64: This is causing unit test failures, because we have a "meta test" which prevents us making exclusions for standard chains. We most likely need to take another look at how we want to deal with this -- but I think we can revert this change and fix the problems separately (it shouldn't block this PR).

Issue #493, opened 2024-08-20
Perform L1 Bytecode checking for contracts with immutables


geoknee:
These contracts are not checked for bytecode match 
```
		"AnchorStateRegistryProxy",
		"DelayedWETHProxy",
		"DisputeGameFactoryProxy",
		"FaultDisputeGame",
		"MIPS",
```
because of their use of immutables, we cannot use a static bytecode hash across all chains to do the validation. 

Suggested fix is to declare the `immutableReferences` for the contract version in question (we get this from the relevant op-contracts tag compiler artifacts) as well as the hash of the _masked_ bytecode (masking here means replacing all immutable references with zeros. Then, the validation check pulls the bytecode from chain, masks it using the immutable references, hashes the result and compares against the expectation.

vdamle:
I'd like to note that the original description mentioned `"DisputeGameFactoryProxy"` as a contract that contains immutables. However, the contract does not have immutables (amended description to remove this contract)

PR #497, opened 2024-08-21
Handle immutable vars when verifying L1 contract bytecodes

vdamle:
Values assigned to immutable vars at initialization are stored in bytecode. The coordinates of such stored immutable values are in the compiled contract AST/json (available in the monorepo). The same is now hardcoded here for contracts that have immutables. The coordinates are used to mask the assigned value of the immutable vars in the bytecode, so that any hash that is computed on a deployed bytecode uses the same pre-image.

Fixes: https://github.com/ethereum-optimism/superchain-registry/issues/493


validation/superchain-version_test.go
```diff
@@ -268,7 +246,19 @@ func getBytecodeHash(ctx context.Context, chainID uint64, contractName string, t
 		return "", fmt.Errorf("%s: %w", addrToCheck, err)
 	}
 
-	return crypto.Keccak256Hash(code).Hex(), nil
+	// if the contract is known to have immutables, setup the filterer to mask the bytes which contain the variable's value
+	bytecodeImmutableFilterer, err := initBytecodeWithImmutables(code, contractName)
+	// the contract bytecode does not have immutables, use the deployed bytecode as-is to calculate hash
+	if err != nil {
+		return crypto.Keccak256Hash(code).Hex(), nil
+	}
```
geoknee, line 254: I'm not sure that using an error to communicate "does not have immutables" is the right pattern, since there are a couple of different errors that can be returned?

How about we either:
1. return a `hashImmutables` boolean as well as the code and the error, and use that in the control flow below to conditionally mask the bytecode
2. have the `.maskBytecode()` method work even if there are no immutable references (perhaps it operates on an empty slice) and just call it unconditionally (could rename it if we go this route).

validation/immutable-references.go
```diff
@@ -0,0 +1,58 @@
...
+
+// ContractASTsWithImmutableReferences stores the `immutableReferences`, copied from the compiled contract AST (from the combined JSON
+// artifact from the monorepo. We do this because the contracts and compiled artifacts are not available in the superchain
+// registry. Ex: ethereum-optimism/optimism/packages/contracts-bedrock/forge-artifacts/MIPS.sol/MIPS.json
+var ContractASTsWithImmutableReferences = map[string]string{
+	"AnchorStateRegistryProxy": `"immutableReferences":{"92829":[{"start":387,"length":32},{"start":828,"length":32},{"start":2296,"length":32}]}`,
+	"DelayedWETHProxy":         `"immutableReferences":{"97827":[{"start":831,"length":32},{"start":4133,"length":32}]}`,
+	"FaultDisputeGame":         `"immutableReferences":{"92829":[{"start":387,"length":32},{"start":828,"length":32},{"start":2296,"length":32}]}`,
+	"MIPS":                     `"immutableReferences":{"85798":[{"start":178,"length":32},{"start":1771,"length":32}]}`,
+}
```
geoknee, line 16: If possible, it would be great to store these in the `standard` package in a toml file, as we do for the versions and bytecodes themselves.

Furthermore, these references are _per op-contracts tag_, so we need to reflect that in the underlying data structures. So these need to be in a table under `op-contracts/v1.4.0`, with space for different references for future versions.

validation/immutable-references.go
```diff
@@ -0,0 +1,58 @@
...
+type ImmutableReference struct {
+	Start  int `json:"start"`
+	Length int `json:"length"`
+}
+
+// Define a struct to represent the structure of the JSON data
+type DeployedBytecode struct {
+	Bytecode            []byte                          `json:"object"`
+	ImmutableReferences map[string][]ImmutableReference `json:"immutableReferences"`
+}
```
geoknee, line 30: Nit: I know this is how the data is arranged in the solidity compiler output. However, I think a name like `bytecodeAndImmutableReferences` or something like that might be more informative?

validation/immutable-references.go
geoknee: Bonus points if it becomes clear to whoever reads the toml file, that the validation is masking the immutables (so is in some sense a slightly weaker validation than it is for other contracts which don't have immutables).

superchain/superchain.go
```diff
@@ -479,6 +479,13 @@ func canonicalizeSemver(version string) string {
 	return version
 }
 
+type ContractBytecodeImmutables struct {
+	AnchorStateRegistry string `toml:"anchor_state_registry,omitempty"`
+	DelayedWETH         string `toml:"delayed_weth,omitempty"`
+	FaultDisputeGame    string `toml:"fault_dispute_game,omitempty"`
+	MIPS                string `toml:"mips,omitempty"`
+}
+
```
geoknee, line 488: Nit: can we move this type to `validation/standard/init.go`, and also the `L1ContractBytecodeHashes` type?

validation/standard/standard-immutables.toml
```diff
@@ -0,0 +1,5 @@
...
+mips = '{"92829":[{"start":387,"length":32},{"start":828,"length":32},{"start":2296,"length":32}]}'
```
geoknee, line 5: Nit, can you tweak your editor settings so that it formats files with a newline at the end? Just prevents noise where this little character appears and disappears...

validation/immutable-references.go
```diff
@@ -0,0 +1,85 @@
...
+
+// Define a struct to represent the structure of the JSON data
+type BytecodeAndImmutableReferences struct {
+	Bytecode            []byte                          `json:"object"`
+	ImmutableReferences map[string][]ImmutableReference `json:"immutableReferences"`
+}
+
+// LoadImmutableReferences parses standard-immutables.toml and stores it in a map. Needs to be invoked one-time only.
+func LoadImmutableReferences() {
```
geoknee, line 35: Feels like this ought to be done in `init.go`?

superchain/superchain.go
```diff
@@ -479,6 +479,13 @@ func canonicalizeSemver(version string) string {
 	return version
 }
 
+type ContractBytecodeImmutables struct {
+	AnchorStateRegistry string `toml:"anchor_state_registry,omitempty"`
+	DelayedWETH         string `toml:"delayed_weth,omitempty"`
+	FaultDisputeGame    string `toml:"fault_dispute_game,omitempty"`
+	MIPS                string `toml:"mips,omitempty"`
```
geoknee, line 486: Perhaps add a comment here explaining the immutable info is encoded as a raw stringified JSON string but stored in a toml file (to make maintenance easier as well as readability).

validation/standard/standard-immutables.toml
```diff
@@ -0,0 +1,5 @@
+["op-contracts/v1.4.0"] # Fault Proof contract AST https://github.com/ethereum-optimism/optimism/releases/tag/op-contracts%2Fv1.4.0
```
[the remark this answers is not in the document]
geoknee: And also that the immutables themselves will vary from chain to chain, so we just ignore them in the validation.

validation/superchain-version_test.go
```diff
@@ -146,7 +146,7 @@ func getContractBytecodeHashesFromChain(chainID uint64, list AddressList, client
 			continue
 		}
 		wg.Add(1)
-		go getBytecodeHashAsync(chainID, contractAddress, results, contractName, wg)
+		getBytecodeHashAsync(chainID, contractAddress, results, contractName, wg)
```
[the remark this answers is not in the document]
vdamle: unintended, fixed

validation/standard/standard-immutables.toml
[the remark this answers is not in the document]
vdamle: Done

validation/immutable-references.go
```diff
@@ -0,0 +1,85 @@
...
+			}
+		}
+	}
+	if bytecodeImmutables != nil {
+		s := reflect.ValueOf(bytecodeImmutables).Elem()
+		for i := 0; i < s.NumField(); i++ {
+			name := s.Type().Field(i).Name
+			value := string(reflect.ValueOf(*bytecodeImmutables).Field(i).String())
+			ContractASTsWithImmutableReferences[name] = value
+		}
```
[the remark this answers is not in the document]
vdamle: Yeah, that's a good idea! can work as a follow up.

superchain/superchain.go
vdamle: Done.

superchain/superchain.go
vdamle: done

geoknee (approved):
LGTM!

PR #498, opened 2024-08-21
Fix mainnet granite activation time

Inphi:
Fix the mainnet granite activation time. It should be set to 11 Sep 2024 16:00:01 UTC.


sebastianst (approved):
checked the timestamp ✅

refcell (approved):

Issue #494, opened 2024-08-20
Allow testnets to have contract deployments greater than standard (even if not yet governance approved) 


geoknee:
Testnets can specify a minimum version (e.g., the latest gov-approved version) but should be allowed to be on any later version.

We are sailing validation failures presently for OP Sepolia, because we rollout proposed contracts there before the governance vote. So we need to build the validation checks against that pattern.

geoknee:
Via #508

PR #508, opened 2024-08-26
permit testnets and devnets to use newer contract versions

tessr:
Addresses #494

### DIFF
--- a/validation/superchain-version_test.go
+++ b/validation/superchain-version_test.go
@@ -10,6 +10,7 @@ import (
 	"testing"
 
 	. "github.com/ethereum-optimism/superchain-registry/superchain"
+	spr "github.com/ethereum-optimism/superchain-registry/superchain"
 	"github.com/ethereum-optimism/superchain-registry/validation/internal/bindings"
 	"github.com/ethereum-optimism/superchain-registry/validation/standard"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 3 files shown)


validation/superchain-version_test.go
```diff
@@ -10,6 +10,7 @@ import (
 	"testing"
 
 	. "github.com/ethereum-optimism/superchain-registry/superchain"
+	spr "github.com/ethereum-optimism/superchain-registry/superchain"
```
geoknee, line 13: We should be able to remove the `spr` import, we are already importing this package with the `.` import magic value.

validation/superchain-version_test.go
```diff
@@ -382,7 +389,18 @@ func checkMatch(s, c reflect.Value) bool {
...
+				return false
+			}
+
+			// testnets are permitted to have contract versions that are newer than what's specified in the standard config
+			// testnets may NOT have contract versions that are older.
+			min := spr.CanonicalizeSemver(field.String())
+			current := spr.CanonicalizeSemver(c.Field(i).String())
+			if semver.Compare(min, current) > 0 {
+				return false
+			}
```
geoknee, line 402: This logic seems right when we are comparing semantic version strings, but this helper is also used when comparing bytecode hashes, where the semver comparison doesn't make sense. 

To fix that, we can either:
1. Store bytecode hashes for anticipated versions of contracts which maybe rolled out to testnets
2. Skip the bytecode hash check for testnets.

tessr: I think (2) probably makes the most sense. We're not performing BHIC for testnets so I think we can skip the bytecode checking here too.

geoknee (approved):

PR #510, opened 2024-08-26
Store Optimism Config in config toml files

### LINKED ISSUE #487: Stop hardcoding EIP1559Elasticity (and store in the registry)
Chains which use a value other than 6 will fail the automated checks in `add-chain` and cannot be added to the registry (at present). 

op-geth function which defaults this to 6:

https://github.com/ethereum-optimism/op-geth/blob/0f5b9dcfd2ac66f6fd8faae526b1549721f5f392/params/superchain.go#L78-L81

op-node function which defaults to 10 but allows the deploy config to override it: 

https://github.com/ethereum-optimism/optimism/blob/2d32f8bfea15ad131bc7b98f732886b5bb9dcff6/op-chain-ops/genesis/genesis.go#L38-L41

geoknee:
Closes #487

* Updates the `add-chain` tool to start storing the `Optimism` field from the Genesis `Config` field in the registry. ~I chose to store this in the compressed genesis files themselves, since that is where the information lives in a "full" genesis file (the one in the registry is heavily deduplicated, this PR is effectively undoing some of that deduplication to allow chains more freedom to vary these parameters).~ The data is now in the top level config TOML files.
* Migrates existing ~compressed genesis files~ config giles to achieve a net zero diff. I did this by writing the previously default values into these genesis files
* One exception is zora sepolia, we happen to know the committed genesis information is not 100% correct and I used this opportunity to fix that (we had another test which happened to catch this). 
* Another case is base sepolia, this had an override in op-geth which I have now removed and placed in the registry instead.
* Note that the diffbot runs on this PR (`Rollup Config and Genesis Diff / compute-genesis-diff / compute-config-diff-and-report-with-pr-comment`). A comment shows any diff this PR introduces (a lack of a comment means we are not changing the effective genesis of any existing chain). 

See https://github.com/ethereum-optimism/op-geth/pull/371


add-chain/cmd/compress-genesis.go
```diff
@@ -116,6 +117,7 @@ var CompressGenesisCmd = cli.Command{
 
 		// convert into allocation data
 		out := Genesis{
+			Config:        &superchain.GenesisConfig{Optimism: (*superchain.OptimismConfig)(genesis.Config.Optimism)},
```
sebastianst, line 120: why the explicit typing?

superchain/superchain.go
```diff
@@ -483,7 +483,18 @@ type GenesisAccount struct {
...
+	EIP1559Elasticity        uint64  `json:"eip1559Elasticity"`
+	EIP1559Denominator       uint64  `json:"eip1559Denominator"`
+	EIP1559DenominatorCanyon *uint64 `json:"eip1559DenominatorCanyon,omitempty"`
+}
+type GenesisConfig struct {
+	// Optimism config, nil if not active
+	Optimism *OptimismConfig `json:"optimism,omitempty"`
+}
 type Genesis struct {
+	Config GenesisConfig `json:"config"`
```
sebastianst, line 497: Why don't we directly add the `*OptimismConfig` as a field here? To better match json hydration?

sebastianst (approved):

add-chain/cmd/compress-genesis.go
geoknee: This is a bit of a gnarly line :-) Improved readability here [27d4033](https://github.com/ethereum-optimism/superchain-registry/pull/510/commits/27d4033e7924e8eaae815a1e025590262576a2a6).

protolambda (dismissed):
Genesis files should never change. The optimism config is likely to change. These two things should be separated.

geoknee:
Note that since we modified `op.toml`, we are triggering validation for OP Mainnet, which is failing at present due to the fallback to the permissioned dispute game.

superchain/configs/sepolia/zora.toml
```diff
@@ -29,6 +29,11 @@ data_availability_type = "eth-da"
     overhead = "0x00000000000000000000000000000000000000000000000000000000000000bc"
     scalar = "0x00000000000000000000000000000000000000000000000000000000000a6fe0"
     gasLimit = 30000000
+  [genesis.config]
+    [genesis.config.optimism]
+      eip1559_elasticity = 6
+      eip1559_denominator = 50
+      eip1559_denominator_canyon = 250
```
geoknee, line 36:
```suggestion
  [genesis.config]
    [genesis.config.optimism]
      eip1559_elasticity = 6
      eip1559_denominator = 50
```

superchain/superchain.go
```diff
@@ -31,12 +31,23 @@ type BlockID struct {
...
-	L2           BlockID      `toml:"l2"`
-	L2Time       uint64       `toml:"l2_time" json:"l2_time"`
-	ExtraData    *HexBytes    `toml:"extra_data,omitempty"`
-	SystemConfig SystemConfig `toml:"system_config" json:"system_config" `
+	L1           BlockID            `toml:"l1"`
+	L2           BlockID            `toml:"l2"`
+	L2Time       uint64             `toml:"l2_time" json:"l2_time"`
+	ExtraData    *HexBytes          `toml:"extra_data,omitempty"`
+	SystemConfig SystemConfig       `toml:"system_config" json:"system_config" `
+	Config       GenesisChainConfig `toml:"config" json:"config"`
```
protolambda, line 50: Why not put it as sub-attribute directly of `superchain.ChainConfig`? This config is the op-stack-specific subset of geth `ChainConfig` equivalent, so it would make sense to put it in the chain-config. It's going to change later, with additional attributes for new parameters, so I don't particularly like putting it into the `ChainGenesis` struct.

protolambda (commented):
I wouldn't nest it into the `genesis` part, since it's not part of genesis, but just general chain-config. That said, this is definitely better than putting it into the compressed genesis allocs, so I'll dismiss my previous review.

refcell:
Pushed a fix for the rust-bindings @geoknee - sorry they get in the way. Will look at removing soon once I get bandwidth.

superchain/superchain.go
geoknee: Ok np I can un-nest it

superchain/configs/mainnet/race.toml
```diff
@@ -13,6 +13,11 @@ seq_window_size = 3600
 max_sequencer_drift = 600
 data_availability_type = "eth-da"
 
+[optimism]
+  eip1559_elasticity = 6
+  eip1559_denominator = 50
+  eip1559_denominator_canyon = 250
+
```
protolambda, line 20: The old `genesis.config` entry seems to be around still (duplicate config data)

geoknee: Fixed

refcell (approved):

PR #515, opened 2024-08-27
Add silence for OP Mainnet Optimism_Portal_2_Params

geoknee:
Our validation suite is currently failing because of the fallback to the permissioned dispute game type on OP Mainnet. We only see this on the hourly `validate-all` job and also on any PR which touches the `op.toml` file. 

Rather than demote or permanently exclude OP Mainnet, here I am silencing it until the Granite upgrade happens.

### DIFF
--- a/validation/exclusions_test.go
+++ b/validation/exclusions_test.go
@@ -18,6 +18,9 @@ func skipIfExcluded(t *testing.T, chainID uint64) {
 		if matches && exclusions[pattern][chainID] {
 			t.Skip("Excluded!")
 		}
+		if matches && silences[pattern][chainID].After(time.Now()) {
+			t.Skipf("Silenced until %s", silences[pattern][chainID].String())
+		}
 	}
 }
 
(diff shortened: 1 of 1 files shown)


validation/exclusions_test.go
```diff
@@ -59,6 +63,12 @@ var exclusions = map[string]map[uint64]bool{
 	},
 }
 
+var silences = map[string]map[uint64]time.Time{
+	"Optimism_Portal_2_Params": {
+		10: time.Unix(1726070401, 0), // mainnet/op silenced until Tue 11 Sep 2024 16:00:01 UTC
```
sebastianst, line 68: Why don't we access the Granite activation timestamp via the `superchain` Go module?

geoknee: Good idea -- [4226d6f](https://github.com/ethereum-optimism/superchain-registry/pull/515/commits/4226d6f346b13d27d9d5f8d9e657e786aecffe4a)

sebastianst (approved):

PR #517, opened 2024-08-28
feat: remove bindings

refcell:
# Adding a New Chain
This PR adds [Chain Name Here] to the registry.

## Checklist
- [ ] I have declared the chain at the appropriate [Superchain Level](../docs/glossary.md#superchain-level-and-rollup-stage).
- [ ] I have run `just validate <chain-id>` locally to ensure all local validation checks pass. 
- [ ] I have run `just codegen` to ensure that the chainlist and other generated files are up-to-date and include my chain.


clabby (approved):

superchain/configs/configs.toml
```diff
@@ -0,0 +1,1088 @@
+##############################################
```
geoknee, line 1: I wonder if we should just rip this out altogether?

geoknee (approved):

superchain/configs/configs.toml
refcell: It'll be a bit ugly for downstream bindings to try to clobber this together without either using custom go scripts or including all static config files and patching it together.

The issue is really traversing the directory in `no_std` doesn't work nicely - we'd end up having to explicitly list out the configs in rust and embed each config file

refcell: I can certainly play around with it in a follow-up pr, but I think this may be the best solution since all it requires is a quick `just codegen` and makes bindings 10x easier and less liable to miss config data

Issue #513, opened 2024-08-27
Get genesis validation checks running in CI. 


geoknee:
Prereq for merging #489 

The checks for genesis validation require a specific set of tool versions to be installed in the docker image in which the validation script runs - some of these tool versions are known at build time, whereas some other tool versions such as `go` is derived from the `go.mod` file  at the commit at which the validation runs. We currently use the image `ci-builder:v0.49.0` for running the all jobs in superchain registry. The image is built from this [Dockerfile](https://github.com/ethereum-optimism/optimism/blob/develop/ops/docker/ci-builder/Dockerfile). So far, we have explored two options to install the correct tool versions to get the genesis validation checks to run on CI runner:

* Install the required tool versions (nvm, go etc.) as pre-req steps in the Circle CI job that is being executed. With this approach, we are observing that the tool versions/settings aren't taking effect when running the actual validation.

* Modify the base image dockerfile to install the required tools in to the base image. I have been unable to build the image locally so far, even without changes (hitting timeout errors and such).

Continuing to explore both options.

geoknee:
The list of requirements are:

```
foundry
gvm
nvm
just
pnpm
yarn
```

PR #519, opened 2024-08-28
switch to cimg/go:1.22.6-node docker image for genesis validation

geoknee:
Seems to get us a bit closer to getting this check to run . 

Towards #513


geoknee:
In favour of #529

PR #514, opened 2024-08-27
Add Ethernity Testnet

denis-pingin:
# Adding a New Chain
This PR adds Ethernity Testnet to the registry.

## Checklist
- [x] I have declared the chain at the appropriate [Superchain Level](../docs/glossary.md#superchain-level-and-rollup-stage).
- [x] I have run `just validate <chain-id>` locally to ensure all local validation checks pass. 
- [x] I have run `just codegen` to ensure that the chainlist and other generated files are up-to-date and include my chain.

PR #522, opened 2024-08-28
fix gas-token_test

geoknee:
As is, we aren't dealing with 32 byte hex string encoded boolean return values.

### DIFF
--- a/validation/gas-token_test.go
+++ b/validation/gas-token_test.go
@@ -81,10 +81,10 @@ func getBool(method string, contractAddress Address, client *ethclient.Client) (
 		return false, err
 	}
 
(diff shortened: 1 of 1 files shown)


geoknee:
Open question whether we should be storing the default (ether) gas paying token in the chain configs. We could store nothing when we detect ether as the gas paying token. 

Example here https://github.com/ethereum-optimism/superchain-registry/pull/514/files#diff-d489766ee3f207b96f7be63ac46d4e87617cc6ff8f9fbf7cd92620148222a28a

vdamle (approved):
0️⃣ ✖️ 1️⃣   -> 1️⃣

Issue #520, opened 2024-08-28
Add contract implementation addresses to superchain config


vdamle:
Currently, superchain config does not have any information about the deployed contract implementation addresses for the contract set in a release. It would be nice to have superchain registry be the source of truth for this information so that other components such as `op-stack-manager` and even the chain `validation` code in superchain registry can refer to this.

See discord [thread](https://discord.com/channels/1244729134312198194/1276265029045456916)

In addition, we also found that this information is necessary to perform MCP L1 upgrade on certain legacy chains.

Oyase-shinobi:
Hello @vdamle 
Hope you are doing well
I have checked superchain github and very interested in this project
I would love to engage in your team
is there any opportunity to work with you?

alfonso-op:
@vdamle can you clarify which specific contract addresses you'd like added to the Superchain config?

PR #525, opened 2024-08-28
Config: Add contract impl addresses & versions to config toml

vdamle:
(updated 9/9/24)

* store and expose the contract implementation addresses & versions for each contracts release. this is intended to be used by internal/external components such as validation code within superchain-registry, op-stack-manager, monorepo upgrade tool etc. The intent is to persist a single source of truth for the implementation addresses.
* the addresses & versions are stored in `validation/standard/standard-versions.toml` per contracts release tag
* the latest contracts release information that is specified in the registry is specified in `superchain.toml` as `op_contracts_release`. Any external consumer should be able to read that tag and look for address/version information under that tag in `validation/standard/standard-versions.toml`
* Initial addresses/versions are recorded for `v1.3.0/MCP` & `v1.4.0` L1 contracts release.

This description is updated per discussion below on the originally created draft PR.

Addresses: https://github.com/ethereum-optimism/superchain-registry/issues/520


superchain/configs/mainnet/superchain.toml
```diff
@@ -12,3 +12,23 @@ granite_time = 1726070401 # Tue 11 Sep 2024 16:00:01 UTC
...
+
+[l1.ImplementationAddresses."op-contracts/v1.3.0"]
+  AddressManager = "0xdE1FCfB0851916CA5101820A69b13a4E276bd81F"
+  L1CrossDomainMessenger = "0xD3494713A5cfaD3F5359379DfA074E2Ac8C6Fd65"
+  L1ERC721Bridge = "0xAE2AF01232a6c4a4d3012C5eC5b1b35059caF10d"
+  L1StandardBridge = "0x64B5a5Ed26DCb17370Ff4d33a8D503f0fbD06CfF"
+  L2OutputOracle = "0xF243BEd163251380e78068d317ae10f26042B292"
+  OptimismMintableERC20Factory = "0xE01efbeb1089D1d1dB9c6c8b135C934C0734c846"
+  OptimismPortal = "0x2D778797049FE9259d947D1ED8e5442226dFB589"
+  SystemConfig = "0xba2492e52F45651B60B8B38d4Ea5E2390C64Ffb1"
```
geoknee, line 24: My one hesitation with this change is it breaks the current pattern of storing "standard data" in the `validation/standard` module+package. 

There are a few options:
1. Live with a mixture / scatter of standard data locations
2. Move this into `validation/standard`
3. Resolve in the other direction, and put standard data into the `superchain` module directly.

We have been so far working under the assumption that `superchain` is probably the right location for data which will be consumed by downstream applications, and `validation` is not really designed to be consumed. 

@mds1 @vdamle @bitwiseguy curious for your ideas on this.

vdamle: > What do you think about having this data live in validation/standard for consistency, and superchain.toml just specifies the latest standard release, such as:

@mds1 @geoknee @bitwiseguy Thank you for the comments and the proposal for what the spelling should look like and where it should live.

I'm in agreement that moving the "standard data" to the `validation/standard` makes sense since that is where most data is sourced/validated from. With that in mind, I have the following format in mind for storing the `{ "version", "implementation_address" }` in `validation/standard/standard-versions.toml`:

```
# Fault Proofs https://github.com/ethereum-optimism/optimism/releases/tag/op-contracts%2Fv1.4.0
["op-contracts/v1.4.0"]
address_manager = { version = "", implementation_address = ""}
l1_cross_domain_messenger = { version = "2.3.0", implementation_address = ""}
l1_erc721_bridge = { version = "2.1.0", implementation_address = ""}
l1_standard_bridge = { version = "2.1.0", implementation_address = ""}
optimism_mintable_erc20_factory = { version = "1.9.0", implementation_address = ""}
optimism_portal = { version = "3.10.0", implementation_address = ""}
system_config = { version = "2.2.0", implementation_address = ""}
protocol_versions = { version = "1.0.0", implementation_address = ""}
anchor_state_registry = { version = "1.0.0", implementation_address = ""}
delayed_weth = { version = "1.0.0", implementation_address = ""}
dispute_game_factory = { version = "1.0.0", implementation_address = ""}
fault_dispute_game = { version = "1.2.0", implementation_address = ""}
permissioned_dispute_game = { version = "1.2.0", implementation_address = ""}

# Fault Proofs https://github.com/ethereum-optimism/optimism/releases/tag/op-contracts%2Fv1.4.0
["op-contracts/v1.3.0"]
address_manager = { version = "", implementation_address = ""}
l1_cross_domain_messenger = { version = "2.3.0", implementation_address = ""}
l1_erc721_bridge = { version = "2.1.0", implementation_address = ""}
l1_standard_bridge = { version = "2.1.0", implementation_address = ""}
optimism_mintable_erc20_factory = { version = "1.9.0", implementation_address = ""}
optimism_portal = { version = "3.10.0", implementation_address = ""}
system_config = { version = "2.2.0", implementation_address = ""}
protocol_versions = { version = "1.0.0", implementation_address = ""}
anchor_state_registry = { version = "1.0.0", implementation_address = ""}
delayed_weth = { version = "1.0.0", implementation_address = ""}
dispute_game_factory = { version = "1.0.0", implementation_address = ""}
fault_dispute_game = { version = "1.2.0", implementation_address = ""}
permissioned_dispute_game = { version = "1.2.0", implementation_address = ""}
```

If we do decide on this, one aspect that is somewhat odd to me is that the table of addresses for each network is in the "config" section of the superchain: https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/configs/mainnet/base.toml#L38-L54 . Perhaps this also ought to live in the `validation` directory? Not suggesting I want to tackle that refactoring as well in this PR, but wanted to solicit feedback while we are on this topic!

geoknee: I'm in favour of moving all of the information into `validation/standard`. I do like the idea of separating out the declaration of which `op-contracts` tag is _currently_ standard (e.g. latest governance approved versions) from the list of historical/current/future tags. For the latter list I like having each tag be a table, with a row for each contract. The value can definitely be a `version`, `implementation_address` pair.

@vdamle your format looks like a mix of toml and json -- what would it look like as pure toml?

vdamle: @geoknee No, this is pure TOML - [inline tables](https://toml.io/en/v1.0.0#inline-table)

Also supported by `burntsushi/toml`: https://github.com/BurntSushi/toml/blob/master/_example/example.toml#L26-L31

bitwiseguy: I'm also in favor of moving all to this information into the `validation/standard` dir. 

I like the data format of `{ version, implementation_address }` for each contract, but I'm curious what `implementation_address` means here. Is it used because all of these contracts are proxies and we actually want the underlying logic contract address? Or could we just use `address` instead of `implementation_address`?

> the table of addresses for each network is in the "config" section of the superchain
Each chain's config file contains information that is specific to that chain. The `[addresses]` field contains addresses of contracts and roles for that chain, which vary from chain-to-chain. Some of the addresses are used by downstream software that consumes the configs (e.g. by op-node [here](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/superchain.go#L87-L88)) so I think that info should probably remain in the configs.

vdamle: > Is it used because all of these contracts are proxies and we actually want the underlying logic contract address? Or could we just use address instead of implementation_address?

They are all not proxies. For the ones that are, the `implementation_address` is for the implementation contract behind the proxy. The concern I have with naming the field as `address` could make it confusing to an end user. As your question would imply, for the contracts that are not proxies, we'll leave the implementation address as an empty string (or nil, if `omitempty` tag works for an inline table field). Let me know what you think @bitwiseguy

superchain/configs/configs.json
```diff
@@ -7,8 +7,10 @@
         "L1": {
           "ChainID": 1,
           "PublicRPC": "https://ethereum-rpc.publicnode.com",
-          "Explorer": "https://etherscan.io"
+          "Explorer": "https://etherscan.io",
+          "Versions": null
```
geoknee, line 11: This doesn't look quite right, do we need to add an `omitempty` struct tag somewhere?

superchain/configs/sepolia/superchain.toml
```diff
@@ -8,6 +8,8 @@ ecotone_time = 1708534800 # Wed 21 Feb 2024 17:00:00 UTC
 fjord_time =   1716998400 # Wed 29 May 2024 16:00:00 UTC
 granite_time = 1723478400 # Mon Aug 12 16:00:00 UTC
 
+op_contracts_release = "op-contracts/v1.4.0"
```
geoknee, line 11: Nit: IMO this information also lives in the `validation/standard` package.

validation/standard/standard-versions.toml
```diff
@@ -1,14 +1,34 @@
...
+anchor_state_registry = { version = "1.0.0" }
+delayed_weth = { version = "1.0.0" }
+dispute_game_factory = { version = "1.0.0" }
+fault_dispute_game = { version = "1.2.0" }
+permissioned_dispute_game = { version = "1.2.0" }
+mips = { version = "1.0.1" }
+preimage_oracle = { version = "1.0.0" }
+
+# Fault Proofs https://github.com/ethereum-optimism/optimism/releases/tag/op-contracts%2Fv1.4.0
+["op-contracts/v1.3.0"]
```
geoknee, line 20: Comment doesn't match table title.

validation/standard/standard-versions.toml
```diff
@@ -1,14 +1,34 @@
...
-protocol_versions = "1.0.0"
-anchor_state_registry = "1.0.0"
-delayed_weth = "1.0.0"
-dispute_game_factory = "1.0.0"
-fault_dispute_game = "1.2.0"
-permissioned_dispute_game = "1.2.0"
+# Fault Proofs https://github.com/ethereum-optimism/optimism/releases/tag/op-contracts%2Fv1.4.0
+["op-contracts/v1.4.0"]
+address_manager = { version = "" }
+l1_cross_domain_messenger = { version = "2.3.0" }
```
geoknee, line 4: Why no `address` for this tag (as you have for the below)?

vdamle: Reached out to @mds1 for the `1.4.0` implementation addresses. Will add here.

superchain/configs/sepolia/superchain.toml
vdamle: I was going by the suggestion in https://github.com/ethereum-optimism/superchain-registry/pull/525#discussion_r1736585302. I've now moved the `op_contracts_release` to `validation/standard`. There is a slight change in the toml spelling due to this.

Because `standard-versions.toml` now has multiple elements, i've had to:
* wrap the map and the tag it in a struct on the go side.
* prefix the contracts release string with an explicit name.

The automatic unmarshaling doesn't work with just the string as the key to the map. Wanted to check what you think about the new spelling:

`standard-versions.toml`

```
op_contracts_release = "op-contracts/v1.4.0"

[contract_tag]

# Fault Proofs https://github.com/ethereum-optimism/optimism/releases/tag/op-contracts%2Fv1.4.0
[contract_tag."op-contracts/v1.4.0"]
address_manager = { version = "" }
l1_cross_domain_messenger = { version = "2.3.0" }
l1_erc721_bridge = { version = "2.1.0" }
...
# Fault Proofs https://github.com/ethereum-optimism/optimism/releases/tag/op-contracts%2Fv1.3.0
[contract_tag."op-contracts/v1.3.0"]
address_manager = { version = "1.0.0", address = "0xdE1FCfB0851916CA5101820A69b13a4E276bd81F" }
...
```

superchain/configs/configs.json
vdamle: Scratch the above. Actually I should have remove the `Versions` field because it is now in `validation/standard`

validation/standard/standard-versions.toml
mds1: L1CrossDomainMessenger hasn't changed in a while, so it's the same implementation address as 1.3.0

validation/standard/standard-versions.toml
```diff
@@ -1,14 +1,38 @@
...
+delayed_weth = { version = "1.0.0" }
+dispute_game_factory = { version = "1.0.0" }
+fault_dispute_game = { version = "1.2.0" }
+permissioned_dispute_game = { version = "1.2.0" }
+mips = { version = "1.0.1" }
+preimage_oracle = { version = "1.0.0" }
+
+# Fault Proofs https://github.com/ethereum-optimism/optimism/releases/tag/op-contracts%2Fv1.3.0
+[contract_tag."op-contracts/v1.3.0"]
+address_manager = { version = "1.0.0", address = "0xdE1FCfB0851916CA5101820A69b13a4E276bd81F" }
```
mds1, line 25: Address Manager has one instance deployed per chain, i.e. chains cannot share an implementation, so we should remove the address from all releases

validation/standard/standard-versions.toml
```diff
@@ -1,14 +1,38 @@
...
+# Fault Proofs https://github.com/ethereum-optimism/optimism/releases/tag/op-contracts%2Fv1.4.0
+[contract_tag."op-contracts/v1.4.0"]
+address_manager = { version = "" }
+l1_cross_domain_messenger = { version = "2.3.0" }
+l1_erc721_bridge = { version = "2.1.0" }
+l1_standard_bridge = { version = "2.1.0" }
+optimism_mintable_erc20_factory = { version = "1.9.0" }
+optimism_portal = { version = "3.10.0" }
+system_config = { version = "2.2.0" }
+protocol_versions = { version = "1.0.0" }
```
mds1, line 14: `protocol_versions` is a superchain level contract, like SuperchainConfig, so we don't need that here

validation/standard/standard-versions.toml
```diff
@@ -1,14 +1,38 @@
...
+
+[contract_tag]
+
+# Fault Proofs https://github.com/ethereum-optimism/optimism/releases/tag/op-contracts%2Fv1.4.0
+[contract_tag."op-contracts/v1.4.0"]
+address_manager = { version = "" }
+l1_cross_domain_messenger = { version = "2.3.0" }
+l1_erc721_bridge = { version = "2.1.0" }
+l1_standard_bridge = { version = "2.1.0" }
+optimism_mintable_erc20_factory = { version = "1.9.0" }
```
mds1, line 11: Looks like a lot of addresses are missing here:
- [Here](https://www.notion.so/oplabs/reMCP-L1-Mainnet-7e216b38c06c48298fe7b38d7dbbf952?pvs=4) is where you can find 1.3.0 addresses
- [Here](https://www.notion.so/oplabs/Fault-Proofs-Mainnet-Deployment-041500821d614593b8565116ebc06f50?pvs=4) links to the 1.4.0 info, but I'd suggest confirming with proofs team
- [This](https://www.notion.so/oplabs/Mainnet-FP-Fixes-Deployment-8f641ce6d3094a5f8c6105184f43781b?pvs=4) should have the info for 1.6.0, again I'd suggest confirming with proofs team

validation/standard/standard-versions.toml
```diff
@@ -1,14 +1,38 @@
...
+optimism_portal = { version = "3.10.0" }
+system_config = { version = "2.2.0" }
+protocol_versions = { version = "1.0.0" }
+anchor_state_registry = { version = "1.0.0" }
+delayed_weth = { version = "1.0.0" }
+dispute_game_factory = { version = "1.0.0" }
+fault_dispute_game = { version = "1.2.0" }
+permissioned_dispute_game = { version = "1.2.0" }
+mips = { version = "1.0.1" }
+preimage_oracle = { version = "1.0.0" }
```
mds1, line 21: For FP contracts:
- OptimismPortal2, DisputeGameFactory, and DelayedWETH have shared implementations for all chains, so should have an address here
- PreimageOracle and MIPS are singletons (one, non-proxied instance for all chains). Therefore these probably should be at the superchain level with SuperchainConfig and ProtocolVersions
- Anchor state registry, fault dispute game permissioned dispute game: we know what semver to expect, but each one has a unique address per chain (i.e. these contracts cannot be reused)

validation/standard/standard-versions.toml
```diff
@@ -1,14 +1,38 @@
...
+l2_output_oracle = { version = "1.8.0", address = "0xF243BEd163251380e78068d317ae10f26042B292" }
+optimism_mintable_erc20_factory = { version = "1.9.0", address = "0xE01efbeb1089D1d1dB9c6c8b135C934C0734c846" }
+optimism_portal = { version = "2.5.0", address = "0x2D778797049FE9259d947D1ED8e5442226dFB589" }
+system_config = { version = "1.12.0", address = "0xba2492e52F45651B60B8B38d4Ea5E2390C64Ffb1" }
+protocol_versions = { version = "1.0.0" }
+anchor_state_registry = { version = "1.0.0" }
+delayed_weth = { version = "1.0.0" }
+dispute_game_factory = { version = "1.0.0" }
+fault_dispute_game = { version = "1.2.0" }
+permissioned_dispute_game = { version = "1.2.0" }
```
mds1, line 38: These do not exist in 1.3.0
```suggestion
```

validation/standard/versions.go
```diff
@@ -7,13 +7,19 @@ import (
...
 
 type (
-	VersionTags            = map[Tag]superchain.ContractVersions
 	BytecodeHashTags       = map[Tag]L1ContractBytecodeHashes
 	BytecodeImmutablesTags = map[Tag]ContractBytecodeImmutables
 )
 
+type VersionTags struct {
+	Tags    map[Tag]superchain.ContractVersions `toml:"contract_tag"`
+	Version string                              `toml:"op_contracts_release,omitempty"`
```
geoknee, line 16: Nit: 
```suggestion
	Version Tag                              `toml:"op_contracts_release,omitempty"`
```

geoknee: Could we call this `StandardVersion` or something?

In general it is a bit confusing to have a single enshrined standard version tag, but then to display detailed (per contract) version information for multiple tags. Are we planning on using the "other" tag information for anything?

validation/standard/standard-versions.toml
```diff
@@ -1,14 +1,38 @@
...
+mips = { version = "1.0.1" }
+preimage_oracle = { version = "1.0.0" }
+
+# Fault Proofs https://github.com/ethereum-optimism/optimism/releases/tag/op-contracts%2Fv1.3.0
```
geoknee, line 23: This release is named "MCP", not "Fault Proofs"

superchain/configs/sepolia/superchain.toml
geoknee: I think that TOML layout is OK, it is still fairly legible. See other comment about whether we need to have multiple contract tags in this file, though.

validation/standard/standard-versions.toml
vdamle: > Anchor state registry, fault dispute game permissioned dispute game: we know what semver to expect, but each one has a unique address per chain (i.e. these contracts cannot be reused)

Retained only semvers, dropped addresses.

> PreimageOracle and MIPS are singletons (one, non-proxied instance for all chains). Therefore these probably should be at the superchain level with SuperchainConfig and ProtocolVersions

@mds1 It appears that the contract addresses for MIPS and PreimageOracle are different for `v1.4.0` and `v1.6.0`. As is, there is no way to disambiguate contract addresses at the superchain level for different releases. Should we keep these addresses in `standard-versions.toml` so we have the ability to store an address per release?

validation/standard/standard-versions.toml
vdamle: Removed.

superchain/configs/sepolia/superchain.toml
vdamle: resolved per discussion in other comment

validation/standard/standard-versions.toml
vdamle: Added implementation addresses to the list, where relevant.

validation/standard/standard-versions.toml
vdamle: removed

validation/standard/standard-versions.toml
mds1: > Should we keep these addresses in `standard-versions.toml` so we have the ability to store an address per release?

Yea this seems like a good solution. I would just suggest that we have some way to differentiate the implementations from the singletons, e.g. maybe in the toml we use `[contract_tag."op-contracts/v1.4.0".implementations]` and `[contract_tag."op-contracts/v1.4.0".singletons]`, or you can add a third key like `{ version ="foo", address = "bar", kind = <"implementation"|"singleton">

validation/standard/versions.go
vdamle: Renamed per above

validation/standard/standard-versions.toml
vdamle: Thanks @mds1 . Received confirmation from @Inphi on Discord that these are indeed the addresses.

validation/standard/standard-versions.toml
```diff
@@ -1,14 +1,43 @@
...
-optimism_mintable_erc20_factory = "1.9.0"
-optimism_portal = "3.10.0"
-system_config = "2.2.0"
-protocol_versions = "1.0.0"
-anchor_state_registry = "1.0.0"
-delayed_weth = "1.0.0"
-dispute_game_factory = "1.0.0"
-fault_dispute_game = "1.2.0"
-permissioned_dispute_game = "1.2.0"
+standard_release = "op-contracts/v1.4.0"
```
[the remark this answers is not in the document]
geoknee: Following up on that here #589

validation/standard/standard-versions.toml
geoknee: Here's how we ended up codifying this https://github.com/ethereum-optimism/superchain-registry/pull/587

geoknee (approved):

PR #526, opened 2024-08-29
Update batcher proposer

cody-wang-cb:
# Update batcher and proposer addressess
We are moving batch sender and proposal sender to different keys on testnet, so updating sepolia-alpha and sepolia. 

## Checklist
- [ ] I have declared the chain at the appropriate [Superchain Level](../docs/glossary.md#superchain-level-and-rollup-stage).


geoknee:
@cody-wang-cb could you please run `just codegen`? Then we should be good to merge this 🙏

cody-wang-cb:
@geoknee Yup just ran codegen, looks like all tests are passed now!

geoknee (approved):

PR #529, opened 2024-08-29
Get genesis validation checks to run on CI

geoknee:
A work in progress. 

⚠️ goes into #489 

Closes https://github.com/ethereum-optimism/superchain-registry/issues/513


.circleci/config.yml
```diff
@@ -267,7 +278,10 @@ workflows:
       - golang-lint
       - golang-modules-tidy
       - golang-test
-      - golang-validate-genesis
+      - golang-validate-genesis:
+          matrix:
+            parameters:
+              chainid: ["7777777", "4801", "1750", "480","34443", "6806"]
```
vdamle, line 284: would it be helpful to add a comment above the chain IDs with the names of the chains as well.

validation/genesis/genesis-predeploy_test.go
```diff
@@ -26,7 +28,23 @@ func perChainTestName(chain *superchain.ChainConfig) string {
 	return chain.Name + fmt.Sprintf(" (%d)", chain.ChainID)
 }
 
+var temporaryOptimismDir string
+
 func TestGenesisPredeploys(t *testing.T) {
+	//
+	// Clone optimism into gitignored temporary directory (if that directory does not yet exist)
```
vdamle, line 35: Perhaps also note that the optimism repo is cloned into a directory that is at the same level as `superchain-registry` with some context (dep resolution errors). I imagine someone else will walk into the same trap as us in the future if they try to modify this 😬

validation/genesis/genesis-predeploy_test.go
```diff
@@ -35,6 +53,8 @@ func TestGenesisPredeploys(t *testing.T) {
 			})
 		}
 	}
+
+	// TODO cleanup temporaryOptimismDir
```
vdamle, line 57: Needs a github issue I think (if you are not planning to address in this PR)

geoknee: I will address that on the feature branch PR before merging that one.

PR #535, opened 2024-08-30
remove temporarily added chains

geoknee:
⚠️ goes into #489 

Note that genesis validation checks are still triggered because I am leaving the genesis validation input files in place. But those checks pass when there is no entry in the `OPChains` mapping (no tests to run). 

I believe this is a good state to leave things in, because we get to save the information necessary to validate chains which haven't been inserted into the registry yet.


vdamle (approved):
✂️

Issue #532, opened 2024-08-29
Add `optimism` config values to inflight PRs adding chains


geoknee:
Since #510 , we store more information in chain config toml files. But PRs created before that change will not have those fields. 

I propose we write a test which requires the fields exist (we can do this via the Go bindings, although they will use default values for the non pointer types, so 0 for `uints`). Then PRs will be forced to add the data before being merged to main:

Most chains will just add:

```toml
[optimism]
  eip1559_elasticity = 6
  eip1559_denominator = 50
  eip1559_denominator_canyon = 250
```

Perhaps we could use any config validation helpers from op-geth to check the values?

PR #536, opened 2024-08-30
validation: add Optimism Config test

bitwiseguy:
Fixes https://github.com/ethereum-optimism/superchain-registry/issues/532


validation/validation_test.go
```diff
@@ -37,10 +37,11 @@ func testValidation(t *testing.T, chain *ChainConfig) {
...
 func testUniversal(t *testing.T, chain *ChainConfig) {
-	t.Run("Genesis Hash Check", func(t *testing.T) { testGenesisHash(t, chain.ChainID) })
-	t.Run("Genesis RPC Check", func(t *testing.T) { testGenesisHashAgainstRPC(t, chain) })
-	t.Run("Uniqueness Check", func(t *testing.T) { testIsGloballyUnique(t, chain) })
-	t.Run("ChainID RPC Check", func(t *testing.T) { testChainIDFromRPC(t, chain) })
+	t.Run("Genesis Hash", func(t *testing.T) { testGenesisHash(t, chain.ChainID) })
+	t.Run("Genesis RPC", func(t *testing.T) { testGenesisHashAgainstRPC(t, chain) })
+	t.Run("Uniqueness", func(t *testing.T) { testIsGloballyUnique(t, chain) })
+	t.Run("ChainID RPC", func(t *testing.T) { testChainIDFromRPC(t, chain) })
+	t.Run("Optimism Config", func(t *testing.T) { testOptimismConfig(t, chain) })
```
geoknee, line 44: If these names are changed, we need to update the matchers in `exclusions_test.go`. We could also try to make that less brittle by defining some constants.

validation/rollup-config_test.go
```diff
@@ -14,3 +14,12 @@ func testRollupConfig(t *testing.T, chain *ChainConfig) {
 	assertIntInBounds(t, "Block Time", chain.BlockTime, standard.BlockTime)
 	assertIntInBounds(t, "Sequencer Window Size", chain.SequencerWindowSize, standard.SequencerWindowSize)
 }
+
+func testOptimismConfig(t *testing.T, chain *ChainConfig) {
```
geoknee, line 18: Maybe a comment explaining the history around this? These values used to be hardcoded and not read from the registry. Since we added them to the registry and remove the hardcodes from op-geth, this test protects us from chains which ran an old version of `add-chain`.

geoknee (changes requested):
Thanks @bitwiseguy . Once we address the exclusions issue, and the comment, we should be good to go.

validation/validation_test.go
bitwiseguy: Defined some constants since these string literals are used in multiple places

geoknee (approved):

PR #540, opened 2024-09-02
Adds validation metadata for chain with id 9897 (arena-z / nod)

geoknee:
I found a genesis creation commit that worked by following this strategy:
* `l1StartingBlockTag` in the deploy config, as well as the rollup.json l1 genesis block time and the genesis timestamp were all for July 31st. 
* I picked a commit just before that using (ensuring to do `git fetch` first)
```
git rev-list -1 --before=1722432480 --format="%cd %H" --reverse origin/develop

commit c2f290d479f7319776205c2640f140a8423cb987
```
* I realised that we were now after the time where a forge script  was being used to generate the allocs (aka the statedump), so made appropriate changes to handle that. 

---
TODO 
- [x] make sure we haven't broken existing chain validations here


superchain/configs/sepolia/arena-z.toml
```diff
@@ -0,0 +1,61 @@
+name = "Arena Z Testnet"
```
geoknee, line 1: We should remove this and related files before merging (keeping only the changes to the `validation` module).

geoknee: Note that having the file here triggers tests to run which are failing. So when we remove, CI should be green.

bitwiseguy (approved):
LGTM

superchain/configs/mainnet/superchain.toml
```diff
@@ -6,7 +6,7 @@ canyon_time =  1704992401 # Thu 11 Jan 2024 17:00:01 UTC
 delta_time =   1708560000 # Thu 22 Feb 2024 00:00:00 UTC
 ecotone_time = 1710374401 # Thu 14 Mar 2024 00:00:01 UTC
 fjord_time =   1720627201 # Wed 10 Jul 2024 16:00:01 UTC
-granite_time = 1726070401 # Wed 11 Sep 2024 16:00:01 UTC
```
geoknee, line 9: Don't want to commit this change.

superchain/configs/mainnet/superchain.toml
```diff
@@ -6,7 +6,7 @@ canyon_time =  1704992401 # Thu 11 Jan 2024 17:00:01 UTC
...
+granite_time = 1726070401 # Tue 11 Sep 2024 16:00:01 UTC
```
geoknee, line 9:
```suggestion
granite_time = 1726070401 # Wed 11 Sep 2024 16:00:01 UTC
```

PR #542, opened 2024-09-02
Update superchain.toml comment for the correct date

sbvegan:
# Description

The granite timestamp (`1726070401`) is for Wednesday, not Tuesday. https://www.epochconverter.com/


ajsutton (approved):
Confirmed the 11th is a Wednesday.

PR #552, opened 2024-09-06
Move `L1SecurityConfigs` and `SuperchainConfig` tests from standard candidate to standard only category

geoknee:
Also tidies up the test orchestration / exclusion logic.

We expect these checks to fail for many standard candidate chains, since they are on pre-MCP contract versions. Since we already do not require candidates to be on MCP yet, it makes sense to relax these other checks too. The checks will remain a blocker for promotion to standard chains, so the chains in question will need to go through an upgrade process to clear that hurdle.

This should unblock several in flight PRs to add new chains to the registry.

### DIFF
--- a/validation/exclusions_test.go
+++ b/validation/exclusions_test.go
@@ -39,21 +39,11 @@ var exclusions = map[string]map[uint64]bool{
 		11763072: true, // sepolia-dev-0/base-devnet-0     No Public RPC declared
 	},
 	UniquenessTest: {
-		11155421: true, // oplabs devnet 0, not in upstream repo
-		11763072: true, // base devnet 0, not in upstream repo}
+		11155421: true, // sepolia-dev-0/oplabs-devnet-0   Not in https://github.com/ethereum-lists/chains
+		11763072: true, // sepolia-dev-0/base-devnet-0     Not in https://github.com/ethereum-lists/chains
 	},
 
 	// Standard Checks
-	L1SecurityConfigTest: {
-		8453:      true, // base (incorrect challenger, incorrect guardian)
-		84532:     true, // base-sepolia (incorrect challenger)
-		7777777:   true, // zora (incorrect challenger)
-		1750:      true, // metal (incorrect challenger)
-		919:       true, // mode sepolia (incorrect challenger)
-		999999999: true, // zora sepolia (incorrect challenger)
-		34443:     true, // mode (incorrect challenger)
-		1740:      true, // metal-sepolia
-	},
 	StandardContractVersionsTest: {
 		11155421: true, // sepolia-dev0/oplabs-devnet-0
 		11763072: true, // sepolia-dev0/base-devnet-0
--- a/validation/exclusions_test.go
+++ b/validation/exclusions_test.go
@@ -5,7 +5,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/ethereum-optimism/superchain-registry/superchain"
+	. "github.com/ethereum-optimism/superchain-registry/superchain"
 	"github.com/stretchr/testify/require"
 )
 
@@ -44,10 +44,6 @@ var exclusions = map[string]map[uint64]bool{
 	},
 
 	// Standard Checks
-	StandardContractVersionsTest: {
-		11155421: true, // sepolia-dev0/oplabs-devnet-0
-		11763072: true, // sepolia-dev0/base-devnet-0
-	},
 	OptimismPortal2ParamsTest: {
 		11763072: true, // sepolia-dev0/base-devnet-0
 	},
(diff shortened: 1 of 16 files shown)


validation/exclusions_test.go
```diff
@@ -5,7 +5,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/ethereum-optimism/superchain-registry/superchain"
+	. "github.com/ethereum-optimism/superchain-registry/superchain"
```
sebastianst, line 8: Why this? It's best practice to use this very sparingly, mostly only when writing a test of a package `package` in a test package `package_test` for dependency reasons.

validation/superchain-genesis_test.go
```diff
@@ -14,9 +14,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func testGenesisHash(t *testing.T, chainID uint64) {
-	skipIfExcluded(t, chainID)
-
+var testGenesisHash SubTestForChain = func(t *testing.T, chain *ChainConfig) {
```
sebastianst, line 17: You do this for all the test functions. Is it necessary? I think if you define
```go
type (
	SubTest         = func(t *testing.T)
	SubTestForChain = func(t *testing.T, chain *ChainConfig)
)
```
as aliases, instead of new types (by adding `=`), you could still use the normal functions, which is maybe cleaner than defining global function `var`s.

validation/validation_test.go
```diff
@@ -30,6 +30,19 @@ const (
 	GenesisAllocsMetadataTest    = "Genesis_Allocs_Metadata"
 )
 
+type (
+	SubTest         func(t *testing.T)
+	SubTestForChain func(t *testing.T, chain *ChainConfig)
+)
```
sebastianst, line 36: See other comments, maybe
```suggestion
type (
	SubTest         = func(t *testing.T)
	SubTestForChain = func(t *testing.T, chain *ChainConfig)
)
```
instead?

validation/validation_test.go
```diff
@@ -61,40 +74,40 @@ func testValidation(t *testing.T, chain *ChainConfig) {
...
 // testStandardCandidate applies to Standard and Standard Candidate Chains.
 func testStandardCandidate(t *testing.T, chain *ChainConfig) {
 	// Standard Config Params
-	t.Run(RollupConfigTest, func(t *testing.T) { testRollupConfig(t, chain) })
-	t.Run(GasTokenTest, func(t *testing.T) { testGasToken(t, chain) })
-	t.Run(ResourceConfigTest, func(t *testing.T) { testResourceConfig(t, chain) })
-	t.Run(GasLimitTest, func(t *testing.T) { testGasLimit(t, chain) })
-	t.Run(GPOParamsTest, func(t *testing.T) { testGasPriceOracleParams(t, chain) })
-	t.Run(StartBlockRPCTest, func(t *testing.T) { testStartBlock(t, chain) })
-	t.Run(SuperchainConfigTest, func(t *testing.T) { testSuperchainConfig(t, chain) })
```
sebastianst, line 80: this was purposefully moved to `testStandard`?

validation/validation_test.go
```diff
@@ -61,40 +74,40 @@ func testValidation(t *testing.T, chain *ChainConfig) {
...
+	t.Run(RollupConfigTest, applyExclusions(chain, testRollupConfig))
+	t.Run(GasTokenTest, applyExclusions(chain, testGasToken))
+	t.Run(ResourceConfigTest, applyExclusions(chain, testResourceConfig))
+	t.Run(GasLimitTest, applyExclusions(chain, testGasLimit))
+	t.Run(GPOParamsTest, applyExclusions(chain, testGasPriceOracleParams))
+	t.Run(StartBlockRPCTest, applyExclusions(chain, testStartBlock))
 	// Standard Config Roles
-	t.Run(L1SecurityConfigTest, func(t *testing.T) { testL1SecurityConfig(t, chain.ChainID) })
```
sebastianst, line 82: this was purposefully moved to `testStandard`?

validation/exclusions_test.go
geoknee: The `validation` package is essentially a test package for `superchain`, it just happens to be a separate module. I find it a lot neater to do this.

validation/superchain-genesis_test.go
geoknee: It's not necessary, I don't think new types vs aliases makes much difference here (we can keep the normal functions if we want). My efforts here are to try and establish a convention for any future dev to follow when adding a new validation check. By declaring the function as a `SubTestForChain` they will get a hint about the function signature they need to adhere to.

validation/validation_test.go
geoknee: Yes, see PR title.

validation/validation_test.go
geoknee: Yeah this is maybe marginally better [90672aa](https://github.com/ethereum-optimism/superchain-registry/pull/552/commits/90672aa6faa305697baabe0b4c5099cfa3322ac5).

sebastianst: I mean, if you do this, you don't need to set global function `var`s, but can define normal functions.

validation/superchain-genesis_test.go
sebastianst: If you define the type alias, you still hint at the function signature, but can keep on defining normal functions instead of global function `var`s.

validation/validation_test.go
sebastianst: 😂

sebastianst (approved):

PR #554, opened 2024-09-06
Some tweaks to mode's config + tweak to `EIP1559DenominatorCanyon` validation behaviour.

geoknee:
I reinserted the chain manually and saw a small diff, so took the opportunity to clean this up.


validation/rollup-config_test.go
```diff
@@ -22,7 +22,7 @@ func testOptimismConfig(t *testing.T, chain *ChainConfig) {
 	require.NotEmpty(t, chain.Optimism, "Optimism config cannot be nil")
 	require.NotEmpty(t, chain.Optimism.EIP1559Elasticity, "EIP1559Elasticity cannot be 0")
 	require.NotEmpty(t, chain.Optimism.EIP1559Denominator, "EIP1559Denominator cannot be 0")
-	if chain.CanyonTime != nil {
-		require.NotEmpty(t, chain.Optimism.EIP1559DenominatorCanyon, "EIP1559DenominatorCanyon cannot be 0 if canyon time is set")
+	if chain.CanyonTime != nil && *chain.CanyonTime == 0 {
+		require.NotEmpty(t, chain.Optimism.EIP1559DenominatorCanyon, "EIP1559DenominatorCanyon cannot be 0 if canyon was activated at genesis")
```
geoknee, line 26: @bitwiseguy this is a bug, I believe. A chain can activate canyon _after_ genesis, and in that case they may not have anticipated the `EIP1559DenominatorCanyon` param at genesis in that case.

It's a bit unclear that these parameters are actually a part of the genesis of the chain.

bitwiseguy: Would the fix for this bug be to set the default `chain.Optimism.EIP1559DenominatorCanyon` value in the `superchain` init() function for chains that activate canyon but don't have the param set? That would be similar to [this code](https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/init.go#L59) in the init() function that populates some hardfork timestamps based on the default value set in the `superchain.toml` file.

Or would it make more sense to explicitly write that param into the individual chain's config .toml?

geoknee: I think I got confused here, I had repeated my old mistake of thinking of the `OptimismConfig` as a static configuration fixed from genesis (in actuality, even though it is stored in `genesis.json`, it is not part of the initial genesis block or anything so can change over time). 

Therefore I think you are right we want to have an automated way to add a hardfork timestamp together with any new configuration introduced by that hardfork to keep things in sync. 

I'll revert this change for now.

superchain/configs/configs.json
```diff
@@ -493,8 +493,7 @@
           "DataAvailabilityType": "eth-da",
           "optimism": {
             "eip1559Elasticity": 6,
-            "eip1559Denominator": 50,
-            "eip1559DenominatorCanyon": 250
```
geoknee, line 497: This parameter was not present when the chain was started up, but it needs to be there now since canyon was activated.

PR #555, opened 2024-09-06
fix(ser): Serialize chain genesis in JSON correctly

clabby:
Corrects the serialization of the `ChainGenesis` and `BlockID` types. In the
[reference](https://github.com/ethereum-optimism/optimism/blob/cl%2Flmdb-witness/op-node/rollup/types.go#L40-L40), fields are serialized in `lower_snake_case`, and right now, they're serialized as `UpperCamelCase`.


superchain/superchain.go
```diff
@@ -38,11 +38,11 @@ type OptimismConfig struct {
...
 type ChainGenesis struct {
-	L1           BlockID      `toml:"l1"`
-	L2           BlockID      `toml:"l2"`
-	L2Time       uint64       `toml:"l2_time" json:"l2_time"`
-	ExtraData    *HexBytes    `toml:"extra_data,omitempty"`
-	SystemConfig SystemConfig `toml:"system_config" json:"system_config" `
+	L1           BlockID      `json:"l1" toml:"l1"`
+	L2           BlockID      `json:"l2" toml:"l2"`
+	L2Time       uint64       `json:"l2_time" toml:"l2_time"`
+	ExtraData    *HexBytes    `json:"extra_data,omitempty" toml:"extra_data,omitempty"`
```
clabby, line 44: There's also no such thing as `ExtraData` in the reference `ChainGenesis` type. Why are we using a duplicated type here rather than re-using the one from the `op-node`?

refcell: Ahh this was the issue ty

refcell: Let's review this in a follow-on pr to unblock

Issue #557, opened 2024-09-09
`add-chain` should detect legacy rollup.json formats using "plasma"


geoknee:
And direct the user to instructions for migrating the `rollup.json`.

PR #559, opened 2024-09-09
add-chain: print plasma error message

bitwiseguy:
Fixes https://github.com/ethereum-optimism/superchain-registry/issues/557

Print an error message if any legacy plasma fields are present in the chain's rollup.json when the add-chain command is run. The error message points the user to the docs section that describes how to adjust the rollup.json to use the new `altda` name and config format.

### DIFF
--- a/add-chain/config/config.go
+++ b/add-chain/config/config.go
@@ -34,6 +34,21 @@ func ConstructChainConfig(
 	if err != nil {
 		return superchain.ChainConfig{}, fmt.Errorf("error reading file: %w", err)
 	}
+
+	// Detect any legacy "plasma" config fields and alert user to adjust their config to match
+	// the new "altda" format
+	var plasmaConfig LegacyPlasma
+	if err = json.Unmarshal(file, &plasmaConfig); err != nil {
+		return superchain.ChainConfig{}, fmt.Errorf("error unmarshaling legacy plasma json: %w", err)
+	}
+	if err = plasmaConfig.CheckNonNilFields(); err != nil {
+		fmt.Printf(
+			"❌ Detected legacy plasma configuration fields in rollup json file.\n" +
+				"❌   - Please follow the guide at the following url to migrate to the new altda format:\n" +
+				"❌   - https://docs.optimism.io/builders/chain-operators/features/alt-da-mode#breaking-changes-renaming-plasma-mode-to-alt-da-mode\n")
+		return superchain.ChainConfig{}, err
+	}
+
 	var chainConfig superchain.ChainConfig
 	if err = json.Unmarshal(file, &chainConfig); err != nil {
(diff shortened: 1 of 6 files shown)


add-chain/config/plasma.go
```diff
@@ -0,0 +1,34 @@
...
+	DACommitmentType           *string             `json:"da_commitment_type"`
+	DAChallengeWindow          *uint64             `json:"da_challenge_window"`
+	DAResolveWindow            *uint64             `json:"da_resolve_window"`
+	UsePlasma                  *bool               `json:"use_plasma"`
+	PlasmaConfig               *interface{}        `json:"plasma_config"`
+}
+
+func (lp *LegacyPlasma) CheckNonNilFields() error {
+	v := reflect.ValueOf(*lp)
+	typeOfT := v.Type()
```
geoknee, line 22: This variable name jars a bit, how about `typeOfV`?

superchain/superchain.go
```diff
@@ -156,7 +156,8 @@ type AltDAConfig struct {
 	DAChallengeWindow *uint64 `json:"da_challenge_window" toml:"da_challenge_window"`
 	// DA resolve window value set on the DAC contract. Used in altDA mode
 	// to compute when a challenge expires and trigger a reorg if needed.
-	DAResolveWindow *uint64 `json:"da_resolve_window" toml:"da_resolve_window"`
+	DAResolveWindow  *uint64 `json:"da_resolve_window" toml:"da_resolve_window"`
+	DACommitmentType *string `json:"da_commitment_type" toml:"da_commitment_type"`
```
geoknee, line 160: Do we need to update downstream packages to pick up this change? Or can it wait?

add-chain/config/config.go
```diff
@@ -34,14 +34,29 @@ func ConstructChainConfig(
...
+		return superchain.ChainConfig{}, fmt.Errorf("error unmarshaling legacy plasma json: %w", err)
+	}
+	if err = plasmaConfig.CheckNonNilFields(); err != nil {
+		fmt.Printf(
+			"❌ Detected legacy plasma configuration fields in rollup json file.\n" +
+				"❌   - Please follow the guide at the following url to migrate to the new altda format:\n" +
+				"❌   - https://docs.optimism.io/builders/chain-operators/features/alt-da-mode#breaking-changes-renaming-plasma-mode-to-alt-da-mode\n")
+		return superchain.ChainConfig{}, err
+	}
+
```
geoknee, line 51: Nit/Optional: we could extract this into a helper and call it from the test?

geoknee (approved):
🙏

add-chain/config/config.go
bitwiseguy: What do you mean by "call it from the test"?

bitwiseguy: Ahh ok I see. I'll keep that in mind for a potential refactor in a follow-up

PR #564, opened 2024-09-11
ci: move validate-modified-chains job to a github action

bitwiseguy:
Changed the github setting so that a maintainer must approve the workflow in github before any github action will run on a forked pr (and this requires a new approval if any new commit is added to the forked pr)


.github/workflows/validation.yml
```diff
@@ -0,0 +1,43 @@
+name: Validate modified chains
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
```
bitwiseguy, line 5: This trigger type allows github secrets (such as our rpc endpoints) to be shared with forked prs. Github documentation [here](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target).

.github/workflows/validation.yml
```diff
@@ -0,0 +1,43 @@
...
+
+jobs:
+  golang-validate-modified:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
```
bitwiseguy, line 14: Since we are using `pull_request_target` trigger (which runs in the context of the base of the pull request), we need to specify here to checkout the forked pr's code before running subsequent checks

geoknee: I found this article pretty useful https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/. I didn't yet convince myself that this pattern is going to be secure. 

 Can we prevent the CI jobs running until we have checked the PR over?

PR #568, opened 2024-09-12
Get OP sepolia validation passing

geoknee:
The only remaining discrepancy is the WETH9 bytecode. The part of the diff that is changed is only the metadata hash https://docs.soliditylang.org/en/latest/metadata.html, which is not semantically meaningful in terms of the contract behaviour. 

Next steps are to:
- [x] trim this metadata hash from both the regenerated and registry allocs before doing the comparison so that the validation job passes
- [x] remove the exclusion for op sepolia genesis metadata check

This should be done in a highly legible / documented way, we anticipate we may see this again for other chains we want to validate.

The monorepo commit/PR at which we were able to verify OP-Sepolia: https://github.com/ethereum-optimism/optimism/pull/6651

h/t to @vdamle who figured this out.


validation/genesis/genesis-allocs_test.go
```diff
@@ -184,6 +196,64 @@ func testGenesisAllocs(t *testing.T, chain *ChainConfig) {
...
+				fmt.Println("The regenerated bytecode length is less than 100 characters; no trimming performed.")
+			}
+		} else {
+			log.Fatalf("Field 'code' not found or is not a string in regenerated bytecode: %s", key)
+		}
+	} else {
+		log.Fatalf("Key %s not found in regenerated bytecode", key)
+	}
+
+	// repeat for gotData
```
geoknee, line 224: Would it be neater to have this function just operate on a single `input []byte`, to reduce duplication? And then just call it twice (one for each regenerated/registry inputs).

geoknee:
@vdamle perhaps let's take this opportunity to remove any unused .patch files, too?

validation/genesis/genesis-allocs_test.go
vdamle: updated

validation/genesis/genesis-allocs_test.go
```diff
@@ -221,37 +223,12 @@ func trimWeth9BytecodeMetadataHash(expectedData []byte, gotData []byte) ([]byte,
...
-			log.Fatalf("Field 'code' not found or is not a string in registry bytecode: %s", key)
-		}
-	} else {
-		log.Fatalf("Key %s not found in registry bytecode", key)
-	}
-
-	expectedData, err = json.Marshal(weth9ExpectedData)
+	data, err = json.Marshal(weth9Data)
 	if err != nil {
 		fmt.Printf("error marshalling regenerated bytecode: %s", err)
```
geoknee, line 228: Here we use "regenerated" even though the function is generic. 

Might be nice to return an error form this function, and then the caller has the context to print out information about which data was being manipulated.

validation/genesis/genesis-allocs_test.go
```diff
@@ -176,6 +177,18 @@ func testGenesisAllocs(t *testing.T, chain *ChainConfig) {
 	gotData, err := json.MarshalIndent(g.Alloc, "", " ")
 	require.NoError(t, err)
 
+	// special handling of weth9 for op-sepolia
+	// We've observed that the contract [metadata hash](https://docs.soliditylang.org/en/latest/metadata.html)
+	// does not match for the bytecode that is generated and what's in the superchain-registry.
+	// The issue is likely a difference in compiler settings when the contract artifacts were generated and
+	// stored in the superchain registry. To account for this, we trim the metadata hash portion of the
+	// weth9 contract bytecode before writing to file/comparing the outputs
+	if chainId == uint64(11155420) {
```
geoknee, line 186: I am wondering if to be extra safe here we require the commitID to be one that we are confident leads to bytecodes with metadata appended, since that is not guaranteed https://docs.soliditylang.org/en/latest/metadata.html. This would avoid issues if we ever change the commit we are using, we would need to whitelist the new commit for it to be treated in this way?

validation/genesis/genesis-allocs_test.go
vdamle: Done

validation/genesis/genesis-allocs_test.go
vdamle: Yeah, wouldn't hurt. I've added this too.

vdamle (approved):
Addressed all of @geoknee comments.

PR #578, opened 2024-09-12
validate: update default monorepo commit to op-contracts/v.1.6.0 tag

bitwiseguy:
This pr updates the default monorepo commit used when validating new chains. The commit used be default is the one for the monorepo tag `op-contracts/v1.6.0`, which is now the recommended commit [in the docs](https://docs.optimism.io/builders/chain-operators/deploy/smart-contracts#best-practices) for new chains to deploy contracts from if they intend to become standard at any point.


.env.example
```diff
@@ -32,7 +32,7 @@ SCR_DEPLOY_CONFIG=${SCR_MONOREPO_DIR}/packages/contracts-bedrock/deploy-config/g
 # This is the commit in the https://github.com/ethereum-optimism/optimism/ repo
 # at which the chain's genesis was created. This is necessary to have our validation checks
 # recreate the genesis file using identical source code
-SCR_GENESIS_CREATION_COMMIT=""
+SCR_GENESIS_CREATION_COMMIT="33f06d2d5e4034125df02264a5ffe84571bd0359" # commit for tag op-contracts/v1.6.0
```
geoknee, line 35: Should we hint that this is only the default value, and it is possible/ok to override it?

geoknee (approved):

PR #580, opened 2024-09-13
ci: separate validate-modified job to require approval on forks

bitwiseguy:
First workflow requires a maintainer approval.

Second workflow requires on the first workflow to complete. Then it can access secrets and run against the forked pr's code

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -1,21 +1,8 @@
-# Workflow 1: requires maintainer approval on fork prs
-name: Pre-check Validate Modified Chains
-
-on:
-  pull_request:
-    types: [opened, synchronize, reopened]
-
-jobs:
-  hold-for-approval:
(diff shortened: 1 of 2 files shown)


geoknee (dismissed):

.github/workflows/validation.yml
```diff
@@ -0,0 +1,38 @@
...
+      uses: actions/checkout@v4
+      with:
+        ref: ${{ github.event.workflow_run.head_commit.id }}
+        fetch-depth: 0  # Ensure the full history is fetched
+
+    - name: Fetch base branch
+      run: |
+        git fetch origin ${{ github.event.workflow_run.base_sha }}
+
+    - name: "Install just"
```
geoknee, line 23: We could do this in a followup PR, but this would be neater https://github.com/extractions/setup-just

geoknee (approved):

PR #581, opened 2024-09-13
Move validation action back to circleci

geoknee:


raffaele-oplabs (approved):
This looks correct to me.

sebastianst (approved):

PR #591, opened 2024-09-16
dedupe types with a generic type

### LINKED ISSUE #590: Deduplicate `ContractBytecodeHashes` and `ContractVersions`
Maybe outside the scope of this pr, but related to the cleanup in this pr: is there a way to combine the `ContractBytecodeHashes` and `ContractVersions` structs into one? It seems they are tracking the same list of contracts, so there is a lot of duplicate code with struct fields and toml tags.

Also the comment describing `ContractBytecodeHashes` looks like a copy/paste error: https://github.com/ethereum-optimism/superchain-registry/blob/gk/simpler-version-check/superchain/superchain.go#L343-L345

_Originally posted by @bitwiseguy in https://github.com/ethereum-optimism/superchain-registry/pull/589#discussion_r1761465298_

geoknee:
closes #590

### DIFF
--- a/superchain/superchain.go
+++ b/superchain/superchain.go
@@ -340,29 +340,31 @@ func (a AddressList) AddressFor(name string) (Address, error) {
 // contract. They are keyed by the semantic version.
 type AddressSet map[string]Address
 
-// ContractVersions represents the desired semantic version of the contracts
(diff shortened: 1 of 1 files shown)


bitwiseguy (approved):
Excellent!

PR #599, opened 2024-09-17
add-chain,validation: update dependency on optimism

geoknee:
Brings in changes from https://github.com/ethereum-optimism/optimism/pull/11830


bitwiseguy (approved):

geoknee:
Note that our diff bots are broken by recent changes to github actions permissions. But still we can see there is a diff generated by this PR https://github.com/ethereum-optimism/superchain-registry/actions/runs/10906969982/job/30269496423?pr=599 @bitwiseguy can you confirm this makes sense to you?

bitwiseguy:
@geoknee `ChannelTimeoutBedrock` is now a constant within `op-node` as seen [here](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/params/constants.go). It is no longer included in the `superchain` or `op-node` rollup config structs

PR #614, opened 2024-09-25
ethernity(mainnet): remove placeholder custom gas token address

bitwiseguy:
The gas token address stored in Ethernity's chain config was a placeholder and not being used. The chain is not actually configured to use a custom gas token. This pr removes that placeholder

### DIFF
--- a/chainList.json
+++ b/chainList.json
@@ -48,8 +48,7 @@
     "parent": {
       "type": "L2",
       "chain": "mainnet"
-    },
-    "gasPayingToken": "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE"
+    }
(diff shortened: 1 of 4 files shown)


sbvegan (approved):
looks good, I'll link this in their other PR

geoknee:
To clarify, this value is not a placeholder but an magic value meaning "ETH", which our tool read from L1 and wrote into the chain config. 

https://specs.optimism.io/experimental/custom-gas-token.html#selection-of-ether_token_address

geoknee:
I'd prefer to fix the `add-chain` tool first if we don't want to store this when there is no _custom_ gas token. Then we can clean up all existing instances.

geoknee (approved):

PR #617, opened 2024-09-26
add-chain: check isCustomGasToken onchain

bitwiseguy:
Adds a call within the add-chain flow for `SystemConfig.isCustomGasToken`. This ensures the `gas_paying_token` does not get populated in the chain config .toml if the feature is not enabled

### DIFF
--- a/add-chain/main.go
+++ b/add-chain/main.go
@@ -244,8 +244,21 @@ func getGasPayingToken(l1rpcURl string, SystemConfigAddress superchain.Address) 
 	if err != nil {
 		return nil, err
 	}
-
 	opts := bind.CallOpts{}
+
+	usingCustomGasToken, err := sc.IsCustomGasToken(&opts)
+	if err != nil {
+		if strings.Contains(err.Error(), "execution reverted") {
+			// This happens when the SystemConfig contract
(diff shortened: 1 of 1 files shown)


add-chain/main.go
```diff
@@ -255,7 +255,19 @@ func getGasPayingToken(l1rpcURl string, SystemConfigAddress superchain.Address)
...
 
+	result, err := sc.GasPayingToken(&opts)
+	if err != nil {
+		if strings.Contains(err.Error(), "execution reverted") {
+			// This happens when the SystemConfig contract
+			// does not yet have the CGT functionality.
+			return nil, nil
+		}
+		return nil, err
+	}
```
geoknee, line 270: Do we need to check for execution reverted here? If the SystemConfig contract does not yet have the CGT functionality, wouldn't we detect that on L251?

geoknee (approved):

add-chain/main.go
bitwiseguy: I'm not really sure. Would have to look through SystemConfig.sol source code. Probably a safe assumption but not sure if there is a case where one of those functions existed without the other. Safer to leave the additional error handling in now and can remove later if/when we confirm

PR #625, opened 2024-10-08
Add Cyber Mainnet

atenjin:
# Adding a New Chain
This PR adds Cyber Mainnet to the registry.

## Checklist
- [x] I have declared the chain at the appropriate [Superchain Level](../docs/glossary.md#superchain-level-and-rollup-stage).


atenjin:
Notice:
When cyber mainnet launching, we set daChallengeWindow and daResolveWindow to 1 at genesis, however op requires us to set to 3600(12hour) later. Then we updated those two value in rollup.json, and update the value in l1 contract (DataAvailabilityChallengeProxy 0x10E34EfE14E4D270C0f77Bf1aF01b6C832161B49)

While for this pr, in deploy-config.json I still keep the old value, but pass the ci.
If you require us to update deploy-config.json for this value, please ping me, I will update for it. 
Thanks.

sbvegan (approved):

parkgunou (approved):

Issue #622, opened 2024-10-01
validation: add checks for newly-added proofs params


bitwiseguy:
Additional params added [here](https://github.com/ethereum-optimism/specs/pull/379/files?w=1). [This section](https://www.notion.so/oplabs/OPCM-Run-superchain-registry-validations-for-new-chain-10ef153ee162809c8646ca250438ac9d?pvs=4#10ef153ee16280aab84bff05164a1970) shows the checks that smart-contracts team did manually that need to be added to the Superchain Registry

Details on some of the new params (work with proofs team to fill in the gap):
* `Fault Game Genesis Output Root`: will be the same for all chains deploying straight to permissioned, but not necessarily for existing chains that upgrade to permissionless fault proofs
* `Fault Game Genesis Block`: Zero for all chains that use fault proofs (permissioned or permissionless) since genesis, but unsure what it should be for chains upgrading to fault proofs
* `Fault Game Absolute Prestate`: for all new chains we expect this value, but it changes with new op-program releases (I am unsure how frequently those are)
* `Large Preimage Proposal Challenge Period`: Must be the same for all chains, whether new or existing
* `Minimum Large Preimage Proposal Size, Bond Withdrawal Delay, Game Clock Extension, Max Game Clock Duration, Fault Game Split Depth, Fault Game Max Depth`: Must be the same for all chains, whether new or existing

ajsutton:
`Fault Game Genesis Output Root` for permissioned game only chains can be anything - makes sense to require it to a specific and obviously wrong value like `0xdead...`.  For permissionless games this should be verified as being the correct output root for the canonical block at `Fault Game Genesis Block` and that the block is finalized. When configuring the chain it should be set to a fairly recent finalized block but is otherwise arbitrary.

[...]

PR #636, opened 2024-10-16
validation: add fault game param checks

bitwiseguy:
## Details

Adds new validation checks related to fault game parameters.

## Tests
* op-mainnet: `just validate 10` passes
* op-sepolia: `just validate 11155420` fails
  * op-program release hash maps to version `1.3.0-rc.3`. Currently the test fails if the hash is not mapped to a standard, non `-rc` tag. What do? Should we change the test to allow `-rc` tags or does op-sepolia need to be updated to use a different op-program prestate?
  * Update: this chain now passes since we are allowing `-rc` tagged releases for testnets

## Metadata
Closes https://github.com/ethereum-optimism/superchain-registry/issues/622

Related monorepo pr: https://github.com/ethereum-optimism/optimism/pull/12523

Depends on golang version update in this pr: https://github.com/ethereum-optimism/superchain-registry/pull/639


validation/fault_game_params_test.go
```diff
@@ -0,0 +1,98 @@
+package validation
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/ethereum-optimism/optimism/op-program/prestates"
```
mds1, line 7: > op-program release hash maps to version `1.3.0-rc.3`. Currently the test fails if the hash is not mapped to a standard, non `-rc` tag. What do? Should we change the test to allow `-rc` tags or does op-sepolia need to be updated to use a different op-program prestate?

I recall the prestate @ajsutton originally suggested we used was from a release candidate and not an official release, but I'm unsure why. Do op-program releases need governance approval?

ajsutton: op-program defines the derivation process that fault proofs uses and is committed to in the contract upgrade signed by the security council. So the op-program being released is generally called out in the governance proposal. We haven't actually been very consistent with proposing an RC and then finalizing on approval or just proposing the final release (Fjord finalized the 1.2.0 release prior to gov approval). 

op-program _also_ includes the chain configuration so testnets will almost always run RCs not final releases because they get hard forks first without gov approval so need a new op-program version that has the hard fork enabled.  Then mainnet needs a different op-program again since it needs to add in the hard fork activation on mainnet. I think the key thing here is that mainnets need gov approval, testnets do not and hard forks are routinely activated on testnets without gov approval, so testnets should be allowed to use RCs.

The other interesting corner case is for new mainnet chains - they will need a new op-program. It should have the same derivation rules applied, but will need a new op-program release to update the superchain-registry to include the new chain. That version won't be governance approved, I'm not sure if it should be an RC or final release.

bitwiseguy: > I think the key thing here is that mainnets need gov approval, testnets do not and hard forks are routinely activated on testnets without gov approval, so testnets should be allowed to use RCs.

This all makes sense to me. I will implement the following:
* testnets can use any release (full, -rc, etc.) and still pass the validation checks
* mainnets must use a full release with the `governanceApproved` flag set

bitwiseguy: Done here and tidied up in the commits after: https://github.com/ethereum-optimism/superchain-registry/pull/636/commits/26c954e57150857b47866860262cd78b9137acb9

parkgunou (approved):

Inphi (approved):
Looks good.

PR #639, opened 2024-10-18
repo: update to go1.22

bitwiseguy:
Bump go version to match monorepo

### DIFF
--- a/add-chain/go.mod
+++ b/add-chain/go.mod
@@ -1,18 +1,14 @@
 module github.com/ethereum-optimism/superchain-registry/add-chain
 
-go 1.21
-
-replace github.com/ethereum-optimism/superchain-registry/superchain => ../superchain
+go 1.22
 
 replace github.com/ethereum/go-ethereum => github.com/ethereum-optimism/op-geth v1.101408.0-rc.4.0.20240828150145-60038121c757
-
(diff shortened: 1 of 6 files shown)


add-chain/go.mod
```diff
@@ -1,18 +1,16 @@
 module github.com/ethereum-optimism/superchain-registry/add-chain
 
-go 1.21
+go 1.22
 
-replace github.com/ethereum-optimism/superchain-registry/superchain => ../superchain
```
geoknee, line 5: Did you mean to remove this and other replacement directives? Can you explain the reasoning?

bitwiseguy: Yes it was intentional. Just a cleanup. Here is the reasoning:

When you are using a go.work file, Go automatically resolves dependencies between the modules listed in the use block based on their local paths. This eliminates the need for the replace directives to point to local directories. The workspace mode will already know where the local modules are located, and it will use them instead of fetching those dependencies from a remote repository.

bitwiseguy: Will revert that part of the change if its concerning. Main purpose of this pr is to bump the go version

geoknee: Reasoning makes sense, thanks.

geoknee (approved):

PR #650, opened 2024-10-23
Regenerated genesis bug

blmalone:
Discoverd in this [pr](https://github.com/ethereum-optimism/superchain-registry/pull/648), we found that the regenerated genesis contains storage allocations that have zero as the value field. This causes the genesis allocs test to fail because op-deployer doesn't do this, nor should it.

### DIFF
--- a/validation/genesis/genesis-allocs_test.go
+++ b/validation/genesis/genesis-allocs_test.go
@@ -159,6 +159,21 @@ func testGenesisAllocs(t *testing.T, chain *ChainConfig) {
 		require.NoError(t, err)
 		allocs := types.GenesisAlloc{}
 		err = json.Unmarshal(expectedData, &allocs)
+		for _, account := range allocs {
+			toDelete := make([]common.Hash, 0)
+
+			for slot, value := range account.Storage {
+				if value == (common.Hash{}) {
+					toDelete = append(toDelete, slot)
+				}
+			}
+
+			for _, slot := range toDelete {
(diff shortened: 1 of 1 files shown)


mds1 (approved):

geoknee:
I'm a bit confused by this change. AFAICT this is not a bug in the validation code, but possibly a bug in the genesis creation code itself. Seems weird to "fix" it here -- wouldn't this risk a valid genesis allocs (valid by our working definition, that it was generated using standard tooling invocations at a specified commit in the monorepo) failing validation (because it would have the the zeros which are now removed from the expectation).

mds1:
The key point is that for slots that are not set in genesis, the default storage value is zero. So two genesis files, where one omits zero valued slots and one explicitly sets them, are equivalent. 

Therefore I think the ideal fix is for the superchain registry to normalize both the provided genesis and the generated genesis, by removing all slots where the value being set in storage is zero. This way it doesn't matter whether or not the provided and generated genesis files have these no-op keys being set, since both versions are equivalent 

In op-deployer, we do this same normalization when generating genesis. But you are right that this diff might cause genesis files generated by other scripts that don't normalize to fail. So we could extend this PR to normalize both genesis files to be more robust

geoknee (approved):

bitwiseguy:
Should we remove empty storage slots in the version of the genesis file that gets stored in the repo? Might save us some bytes

PR #659, opened 2024-10-25
Follow up PR for OPCM in SCR

blmalone:
Follow up pr to: https://github.com/ethereum-optimism/superchain-registry/pull/640


bitwiseguy (approved):
Looks good!

Since the new test only affects standard chains, I ran `just validate 10` (op-mainnet, standard chain) to confirm it passes the updated test suite. ✅

PR #666, opened 2024-10-31
validation: check all contracts for correct superchainConfig()

bitwiseguy:
Updates the `testSuperchainConfig` to ensure that the following contracts are all pointed at the standard superchain-wide `SuperchainConfig` contract.
```
* AnchorStateRegistry
* DelayedWETH
* L1CrossDomainMessenger
* L1ERC721Bridge
* L1StandardBridge
* OptimismPortal
```
Previously we were only checking the OptimismPortal contract

### Tests
I manually ran the following to ensure our existing standard chains pass this test:
```
just validate 10 // op-mainnet
just validate 11155420 // op-sepolia
```

### DIFF
--- a/add-chain/main.go
+++ b/add-chain/main.go
@@ -90,7 +90,6 @@ func entrypoint(ctx *cli.Context) error {
 	sequencerRPC := ctx.String(flags.SequencerRpcFlag.Name)
 	explorer := ctx.String(flags.ExplorerFlag.Name)
 	superchainTarget := ctx.String(flags.SuperchainTargetFlag.Name)
-	monorepoDir := ctx.String(flags.MonorepoDirFlag.Name)
 
 	chainName := ctx.String(flags.ChainNameFlag.Name)
 	rollupConfigPath := ctx.String(flags.RollupConfigFlag.Name)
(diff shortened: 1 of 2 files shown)


add-chain/main.go
```diff
@@ -90,7 +90,6 @@ func entrypoint(ctx *cli.Context) error {
 	sequencerRPC := ctx.String(flags.SequencerRpcFlag.Name)
 	explorer := ctx.String(flags.ExplorerFlag.Name)
 	superchainTarget := ctx.String(flags.SuperchainTargetFlag.Name)
-	monorepoDir := ctx.String(flags.MonorepoDirFlag.Name)
```
geoknee, line 93: Were the changes to this file intentional?

validation/superchain-config_test.go
```diff
@@ -20,12 +20,27 @@ func testSuperchainConfig(t *testing.T, chain *ChainConfig) {
...
 	require.NoErrorf(t, err, "could not dial rpc endpoint %s", rpcEndpoint)
-	opp := Addresses[chain.ChainID].OptimismPortalProxy
 
-	got, err := getAddress("superchainConfig()", opp, client)
-	require.NoError(t, err)
+	checkSuperchainConfig(t, client, Addresses[chain.ChainID].OptimismPortalProxy, *expected)
+	checkSuperchainConfig(t, client, Addresses[chain.ChainID].AnchorStateRegistryProxy, *expected)
+	checkSuperchainConfig(t, client, Addresses[chain.ChainID].L1CrossDomainMessengerProxy, *expected)
+	checkSuperchainConfig(t, client, Addresses[chain.ChainID].L1ERC721BridgeProxy, *expected)
+	checkSuperchainConfig(t, client, Addresses[chain.ChainID].L1StandardBridgeProxy, *expected)
```
geoknee, line 28: Nit: Do you think it would be worth enshrining more of this information into the standard config toml files? I think there is precedent for declaring that `contractFoo.MethodBar` should resolve to `AddressBaz` already.

geoknee (approved):
Approving but we should hold off merging until the governance question resolves.

add-chain/main.go
bitwiseguy: Good catch. They are intentional but I will separate those changes into a different pr to keep this one cleaner

validation/superchain-config_test.go
bitwiseguy: Since changing those toml files requires a governance approval, I think we should hold off on enshrining this change for now

Issue #668, opened 2024-11-04
Error running validation checks on all chains


vdamle:
**Bug Description**

Occasionally, the test to run validation checks on all chains - `just validate "*"` fails many checks. Upon looking at the error logs in CI, there appears to be a `concurrent map read and map write` occurring which causes this error. See [CI Job output](https://app.circleci.com/pipelines/github/ethereum-optimism/superchain-registry/9295/workflows/71929054-0c64-4502-9f89-48506e0e4eac/jobs/38925)

**Steps to Reproduce**

run `just validate "*"`

**Expected behavior**

Validation should pass without errors/crashes.

**Environment Information:**
N/A
**Configurations:**

`just validate "*"`

**Logs:**

```
--- PASS: TestValidation/Shape_(360)/Uniqueness (0.00s)
--- PASS: TestValidation/Shape_Sepolia_Testnet_(11011)/Uniqueness (0.00s)
--- PASS: TestValidation/World_Chain_(480)/ChainID_RPC (0.10s)
concurrent map read and map write
--- PASS: TestValidation/Unichain_Sepolia_Testnet_(1301)/Uniqueness (0.00s)
--- PASS: TestValidation/Base_(8453)/Uniqueness (0.00s)

goroutine 217103 [running]:
github.com/ethereum-optimism/superchain-registry/validation.chainIDSet.AddIfUnique(...)
        /home/circleci/project/validation/uniqueness_test.go:26
github.com/ethereum-optimism/superchain-registry/validation.testIsGloballyUnique(0xc04e8864e0, 0xc00027c808)
        /home/circleci/project/validation/uniqueness_test.go:126 +0x1a5
github.com/ethereum-optimism/superchain-registry/validation.testUniversal.applyExclusions.func3(0xc04e8864e0)
        /home/circleci/project/validation/validation_test.go:43 +0x42
```

**Additional context**
N/A

PR #669, opened 2024-11-04
Validation: wrap local chain info maps with mutexes

vdamle:
* Appears that there's a race when attempting to update local chain ID/Name maps from multiple tests which are kicked off concurrently.
* Also refactored the code a bit to wrap the global vars into a struct

## Checklist
- [ ] ~I have declared the chain at the appropriate [Superchain Level](../docs/glossary.md#superchain-level-and-rollup-stage).~

ref: https://github.com/ethereum-optimism/superchain-registry/issues/668

### DIFF
--- a/validation/uniqueness_test.go
+++ b/validation/uniqueness_test.go
@@ -21,38 +21,21 @@ type uniqueProperties struct {
 	RPC       []string
 }
 
-type (
-	chainIDSet   map[uint64]bool
-	chainNameSet map[string]bool
-)
-
 type chainInfo struct {
-	localChainIds   chainIDSet
-	localChainNames chainNameSet
-	chainIDMutex    sync.RWMutex
-	chainNameMutex  sync.RWMutex
+	localChainIds   sync.Map
+	localChainNames sync.Map
 }
 
 func (c *chainInfo) AddIfUnique(chainId uint64, chainName string) error {
 	// Check if chainId is unique
-	c.chainIDMutex.RLock()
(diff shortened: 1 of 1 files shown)


validation/uniqueness_test.go
```diff
@@ -20,23 +21,39 @@ type uniqueProperties struct {
...
+		return fmt.Errorf("Chain with ID %d is duplicated", chainId)
+	}
+	c.chainIDMutex.RUnlock()
 
-func (s chainNameSet) AddIfUnique(name string) error {
-	if s[name] {
-		return fmt.Errorf("Chain with name %s is duplicated", name)
+	// Check if chainName is unique
+	c.chainNameMutex.RLock()
+	if c.localChainIds[chainId] {
```
geoknee, line 47: Don't you want to check `c.localChainNames` here?

Also, where do you write to `c.localChainIds`?

geoknee (commented):
Did you consider using https://pkg.go.dev/sync#Map ?

validation/uniqueness_test.go
vdamle: This was a copy-paste error. moved to `sync.Map` per your other suggestion, this is no longer relevant.

vdamle:
> Did you consider using https://pkg.go.dev/sync#Map ?

Thanks for the suggestion, agree that's much simpler to use for this case and modified code accordingly.

geoknee (approved):

PR #685, opened 2024-11-15
validation: unmarshal standard release into exported struct member

geoknee:
and panic if we get an empty string.

This fixes a bug where the effective standard contract versions were not bound to the TOML files properly. This was causing contract version validation to **always pass** regardless of the contract versions used by any particular chain.


geoknee:
See #687 which adds a test for this scenario.

bitwiseguy (approved):

PR #701, opened 2024-11-21
remove SuperchainConfig from per-chain config files

geoknee:
This is a superchain-wide contract whose address is either discovered on chain during validation checks, or the standard address can be looked up in the superchain.toml file.


superchain/superchain.go
```diff
@@ -297,7 +297,6 @@ type AddressList struct {
 	OptimismPortalProxy               Address `json:"OptimismPortalProxy,omitempty" toml:"OptimismPortalProxy,omitempty"`
 	SystemConfigProxy                 Address `json:"SystemConfigProxy" toml:"SystemConfigProxy"`
 	ProxyAdmin                        Address `json:"ProxyAdmin" toml:"ProxyAdmin"`
-	SuperchainConfig                  Address `json:"SuperchainConfig,omitempty" toml:"SuperchainConfig,omitempty"`
```
sebastianst, line 300: Did we check that there aren't any downstream dependencies that rely on this field? If there are, how should the access this field from Go?

geoknee: I can't find any references, I will double check by updating the imports in op-geth and the monorepo and seeing if we get any compiler errors.

geoknee: Found one reference, we would need to make this change in the monorepo https://github.com/ethereum-optimism/optimism/compare/gk/remove-superchain-config-from-scr?expand=1

superchain/extra/addresses/addresses.json
```diff
@@ -42,7 +42,6 @@
     "Proposer": "0xf8Ed03b83c15aa3d0b52095F3c9971225948B777",
     "ProxyAdmin": "0xd60a706Bf6108F090d055787B9B353FA7EEE1355",
     "ProxyAdminOwner": "0x4c4526C8C55c4e1a6F2aaf91e37f07c97786e0f5",
-    "SuperchainConfig": "0xe5b3692266FF4Ab8A96A9C7Da6EeEe532CCc7916",
```
mds1, line 45: What is the intent of `addresses.json` and who should consume it? I would expect that it moves in the opposite direction to serve as a flat, comprehensive list of all contracts for that chain, to avoid needing view a few different files to collect all addresses

It seems like currently for chains not using the standard `SuperchainConfig`, there is no way to read their `SuperchainConfig` from the registry? In which case we could look up their `SuperchainConfig` during codegen and that way the flat codegen'd `addresses.json` contains all addresses a suprchain registry consumer would need

superchain/superchain.go
geoknee: I did check `op-geth`, no changes required there. But see the other thread about whether we want to make this change overall.

superchain/extra/addresses/addresses.json
geoknee: I believe it is there for the convenience of downstream applications / other users. Those apps may not wish to discover addresses on chain -- it's worse than just going over multiple files since the `SuperchainConfig` can actually vary per chain and not match the standard superchain address (we only validate standard chains are compliant). 

I think the codegen here is a bit of a distraction, it will contain the same information as the individual TOML files. It is just that some of the individual TOML files are missing the `SuperchainConfig` address.

It sounds like the better change might be to backfill missing addresses then, and add a validation check for all chains that the in chain references resolve to the ones in the registry. Would you be in favor of this @sebastianst @mds1 ? I'll push this back to a draft PR until we can align.

Issue #686, opened 2024-11-15
validation: add chain config negative testing


geoknee:
Much or our validation code is untested, which means we are very prone to bugs like this #685.

We should be using local dummy data to ensure validation checks catch chains with non-standard or missing data. We should refactor tests as necessary to enable them to be tested in this way.

alfonso-op:
@bitwiseguy to create sub issues for different validation tests we want to add tests for.

alfonso-op:
Spoke with @bitwiseguy, closing this as this is not relevant given that we have switched to a report-based approach from a validation one for some time now.

PR #711, opened 2024-11-26
validation: refactor uniqueness check

bitwiseguy:
Incremental step towards https://github.com/ethereum-optimism/superchain-registry/issues/686

Establishes the following pattern for chain `validations`:
* each validation has a `check` function (e.g. `checkIsGloballyUnique`)
  * takes all relevant accumulated data as inputs and then runs checks
  * does NOT take a `*testing.T` as an input
  * returns an error if any data is invalid
  * has unit tests with dummy data to check negative cases
* each validation has a `test` function (e.g. `testIsGloballyUnique`)
  * thin wrapper around the `check` function
  * invoked by the high level `validation` entrypoint, which runs against live chains

### DIFF
--- a/validation/uniqueness.go
+++ b/validation/uniqueness.go
@@ -1,0 +1,191 @@
+package validation
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"slices"
+	"strings"
+	"sync"
+	"testing"
+
+	. "github.com/ethereum-optimism/superchain-registry/superchain"
+	"github.com/stretchr/testify/require"
+)
+
+type uniqueProperties struct {
+	Name      string
+	ShortName string
+	RPC       []string
+}
+
+type chainInfo struct {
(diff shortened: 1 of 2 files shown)


validation/uniqueness.go
```diff
@@ -0,0 +1,191 @@
...
+	ErrChainNameDuplicated     = errors.New("chain name duplicated")
+	ErrChainPublicRpcNotListed = errors.New("chain public RPC not listed in chainid.network")
+)
+
+func testIsGloballyUnique(t *testing.T, chain *ChainConfig) {
+	err := checkIsGloballyUnique(globalChainIds, localChains, chain)
+	require.NoError(t, err)
+}
+
+func checkIsGloballyUnique(gcids map[uint]*uniqueProperties, lcs *chainInfo, chain *ChainConfig) error {
```
geoknee, line 43: Nit: the variable names `gcids` and `lcs` don't seem super obvious to me.

geoknee (approved):
👏

PR #716, opened 2024-11-27
validation: unit test for zero values after read from standard tomls

bitwiseguy:
Adds a unit test suite for `validation/standard` to ensure the `init()` function correctly populates fields when reading from toml. Should help prevent downstream bugs and silent failures in code which depends on these values


validation/standard/init_test.go
```diff
@@ -0,0 +1,54 @@
...
+			versions, ok := NetworkVersions[network]
+			require.True(t, ok, "NetworkVersions[%s] should exist", network)
+			require.NotNil(t, versions, "NetworkVersions[%s] should not be nil", network)
+			require.NotZero(t, len(versions.Releases), "NetworkVersions[%s].Releases should not be empty", network)
+		})
+	}
+
+	t.Run("Release", func(t *testing.T) {
+		require.NotEmpty(t, Release, "Release should not be empty")
+	})
```
geoknee, line 43: Could we add a check that there is an entry in each of:
* versions
* bytecodes
* immutables
which matches this release?

Essentially, this is just a pointer to one of the tables in those files, so we want to make sure it point somewhere.

validation/standard/init_test.go
```diff
@@ -0,0 +1,54 @@
+package standard
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestConfigInitialization(t *testing.T) {
```
geoknee, line 10: What are your thoughts on running this as a test versus doing the check in the `init()` function and panicking if there is an error?

geoknee (approved):

validation/standard/init_test.go
bitwiseguy: Converting these from tests to checks in `init()` would allow us to catch any errors sooner (i.e. when the package is first initialized vs only when unit tests are run). However, negative test cases (ensure failure when an invalid value is detected) would still have to go into test files. Will leave as-is for now but open to refactoring later

validation/standard/init_test.go
bitwiseguy: Good call. Added in https://github.com/ethereum-optimism/superchain-registry/pull/716/commits/99a93dc5f8ea06a32f2f7c073934d34fe3449d7c and https://github.com/ethereum-optimism/superchain-registry/pull/716/commits/392c1f0eaaa721cc125dba1b2621288b786cd08d

PR #717, opened 2024-11-27
refactor(add-chain): refactor default command

edobry:
# Description

This PR refactors the `add-chain` tool to improve DevX in two ways:

- make the `.env` file optional
- does not require flags not needed for a given subcommand

This is achieved by extracting the default action (what happens when the tool is run without a subcommand) into a command in the `cmd` directory, and adding logic to the app entrypoint to call it if no subcommand is provided. This then allows us to set flags at the subcommand level, which means flags set as `required` will no longer cause subcommands not needing them to fail if they're not provided. In addition, this simplifies some command's logic by making it unnecessary to manually check if required flags are provided.


justfile
```diff
@@ -4,7 +4,7 @@ alias l := lint-all
 
 # Adding a chain
 add-chain:
-	go run ./add-chain
+	go run ./add-chain add-new-chain
```
geoknee, line 7: Wondering if it would make sense to rename the top level cmd to something more general like `scr-ops`? Don't feel strongly, though.

geoknee (approved):
Thanks for this, will be a definite improvement in the developer experience.

Are there any documentation changes required to accompany this? I was thinking here https://github.com/ethereum-optimism/superchain-registry/blob/main/docs/add-chain.md#3-run-script but I guess we are preserving the `just add-chain` API so no changes needed?

justfile
edobry: I was also thinking this, but didn't feel confident making a larger change like that. do you think we should do that as part of this PR, or as follow-up work?

edobry: i've gone ahead and made this change in #725

geoknee: Thanks. I think in a follow-up makes sense. We should also get @bitwiseguy's opinion.

Issue #676, opened 2024-11-07
add "Governed By Optimism" field to the chain lists


tessr:
In the time since we originally spec'd out the SCR, we've aligned on "Governed by Optimism" as a key criteria and differentiator for chains. We should include this field in `chainList.toml` and `chainList.json` in addition to the `superchainLevel`.

(Note that `superchainLevel` requires that chains _both_ have their ProxyAdminOwner role set to the 2/2 _and_ that the chain uses permissionless FPs, e.g., that the chain is on the latest gov-approved version. "Governed by Optimism" should refer only to the key handover status.)

PR #718, opened 2024-11-27
feat(superchain): add GovernedByOptimism field to chain config

edobry:
### Description

This PR adds a boolean `GovernedByOptimism` field to the chain config, which indicates if a certain chain is subject to Optimism governance. It also adds a `GovernedByOptimismTest` test to the `validation` package, which ensures that any chain which has the `governed_by_optimism` field set to `true` in its config also has its `ProxyAdmin` address set to the Optimism multisig.

It also sets the new `governed_by_optimism` field on all chain configs as appropriate

### Metadata

- **In service of**: #676 
- **Depends on:** #717


geoknee:
Two main questions:

1. If we declare a chain as governed by optimism, but it actually isn't, how would we tell?

I am wondering if this change be accompanied by some edits to this validation check:

https://github.com/ethereum-optimism/superchain-registry/blob/c56b99231b9033d4db7d64f4fe4eeca72d733a8a/validation/key-handover_test.go#L12-L32

EDIT: just spotted the other PR #719. I think it would make sense to commit the changes in a single PR, if possible.

2. When does this boolean get written / updated? During the `add-chain` tool? Manually? During promotion to a standard chain? Do we need to update the .env schema / add-chain tool at all?

validation/governed-by-optimism_test.go
```diff
@@ -0,0 +1,23 @@
...
+func testGovernedByOptimism(t *testing.T, chain *ChainConfig) {
+	chainID := chain.ChainID
+	superchain := OPChains[chainID].Superchain
+
+	optimismMultisig := standard.Config.MultisigRoles[superchain].KeyHandover.L1.Universal["ProxyAdmin"]["owner()"]
+
+	if chain.GovernedByOptimism == true {
+		require.Equal(t, optimismMultisig, chain.Addresses.ProxyAdminOwner.String(), "Chains using Optimism governance must have their ProxyAdminOwner set to the Optimism multisig")
+	}
+}
```
bitwiseguy, line 23: Could this be combined with the existing `key-handover_test.go` code? Seems like that test is doing basically the same thing, except for checking the new boolean

superchain/configs/mainnet/op.toml
```diff
@@ -4,6 +4,7 @@ public_rpc = "https://mainnet.optimism.io"
 sequencer_rpc = "https://mainnet-sequencer.optimism.io"
 explorer = "https://explorer.optimism.io"
 superchain_level = 1
+governed_by_optimism = true
```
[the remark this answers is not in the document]
edobry: I was thinking that it may not be important to set it for the `false` cases, but I can go ahead and do this

validation/governed-by-optimism_test.go
```diff
@@ -0,0 +1,22 @@
+package validation
+
+import (
+	"testing"
+
+	. "github.com/ethereum-optimism/superchain-registry/superchain"
+	"github.com/ethereum-optimism/superchain-registry/validation/standard"
+
+	// "github.com/ethereum/go-ethereum/common"
```
bitwiseguy, line 9: This commented line can be removed

validation/governed-by-optimism_test.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func testGovernedByOptimism(t *testing.T, chain *ChainConfig) {
+	chainID := chain.ChainID
+	superchain := OPChains[chainID].Superchain
+
+	if chain.GovernedByOptimism == true {
+		optimismMultisig := standard.Config.MultisigRoles[superchain].KeyHandover.L1.Universal["ProxyAdmin"]["owner()"]
+		require.Equal(t, optimismMultisig, chain.Addresses.ProxyAdminOwner.String(), "Chains using Optimism governance must have their ProxyAdminOwner set to the Optimism multisig")
```
bitwiseguy, line 20: Should we also check for `chain.Addresses.ProxyAdminOwner.String() != optimismMultisig` in the case that `chain.GovernedByOptimism == false`? It would be nice if we did that so that this validation check would fail if someone updated a chain config's `ProxyAdminOwner` but forgets to update the `governed_by_optimism` bool

tessr:
Just chiming in to confirm that this can be the same logic as the key handover test. We are functionally checking the same thing (whether the proxy admin owner is set to the correct address).

opgitgovernance:
Detected nonzero diff in rollup config. This PR modifies the configuration of at least one chain in the registry, which may or may not have been intended. You can see the detail [here](https://circleci.com/gh/ethereum-optimism/superchain-registry/50281).

opgitgovernance:
Detected nonzero diff in genesis. This PR modifies the genesis data of at least one chain in the registry, which may or may not have been intended. You can see the detail [here](https://circleci.com/gh/ethereum-optimism/superchain-registry/50280).

validation/governed-by-optimism_test.go
edobry: done

edobry:
I've updated this PR with several changes:

1. restructured the `validation/standard/*.toml` files to improve DevX
2. added the negative test case [suggested here](https://github.com/ethereum-optimism/superchain-registry/pull/718#discussion_r1869691265)
3. added standard configs for the `sepolia-dev-0` chain to account for newly-failing tests caused by 2.

I explored whether we can reuse the `Key_Handover` test for this usecase but this doesn't seem possible, as it doesn't check the configured address directly but rather calls a contract method; I tested this by making changes in the config file which should've caused it to fail, but it didn't, not fully sure why. Please LMK if I've misunderstood something here.

validation/standard/config/networks/mainnet/standard-config-roles-mainnet.toml
bitwiseguy, line 1: I believe some of these `standard-*.toml` files are referenced in the governance blockspace charter. @tessr @geoknee do you know if moving these files around will break any links or cause any other issues? If so, I would suggest reverting these toml file refactors from within this pr. Some of the proposed op-deployer/superchain-registry integration work will refactor these files as well, so might be best to postpone this part of the work and just do it once

validation/security-configs_test.go
```diff
@@ -17,29 +17,42 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-var checkResolutions = func(t *testing.T, r standard.Resolutions, chainID uint64, client *ethclient.Client) {
-	for contract, methodToOutput := range r {
+func resolveAddress(t *testing.T, chainID uint64, contract string) (Address, error) {
```
bitwiseguy, line 20: Nit: could we change this function name to `getAddressFromConfig` and change the argument `contract` --> `contractName`

validation/standard/config/networks/sepolia-dev-0/standard-versions-sepolia-dev-0.toml
bitwiseguy, line 1: Have you verified that the addresses in this file are correct? We currently have a gap in our validations that doesn't actually check that the addresses used by a chain match the addresses in this file, which is likely the reason we aren't seeing an error for sepolia-dev-0 now that this file exists

opgitgovernance:
Detected nonzero diff in rollup config. This PR modifies the configuration of at least one chain in the registry, which may or may not have been intended. You can see the detail [here](https://circleci.com/gh/ethereum-optimism/superchain-registry/51847).

opgitgovernance:
Detected nonzero diff in genesis. This PR modifies the genesis data of at least one chain in the registry, which may or may not have been intended. You can see the detail [here](https://circleci.com/gh/ethereum-optimism/superchain-registry/51846).

validation/standard/config/networks/sepolia-dev-0/standard-versions-sepolia-dev-0.toml
edobry: changes moved to https://github.com/ethereum-optimism/superchain-registry/pull/761

validation/standard/init_test.go
```diff
@@ -28,18 +28,28 @@ func TestConfigInitialization(t *testing.T) {
...
 		})
 
 		t.Run(fmt.Sprintf("MultisigRoles[%s]", network), func(t *testing.T) {
+			roles := Config.MultisigRoles[network]
+
 			// Ensure network MultisigRoles are populated
-			require.NotNil(t, Config.MultisigRoles[network], "Config.MultisigRoles[%s] should not be nil", network)
-			require.NotZero(t, Config.MultisigRoles[network], "Config.MultisigRoles[%s] should not be zero value", network)
+			require.NotNil(t, roles, "Config.MultisigRoles[%s] should not be nil", network)
+			if roles != nil {
```
bitwiseguy, line 43: Nit: since you require that `roles != nil` on the line above, don't think you need this `if` statement. Same for the `if l1Roles != nil` statement below

bitwiseguy (approved):
LGTM

validation/standard/init_test.go
edobry: my thinking was that since `require.NotNil` only fails the test, the program would continue and could potentially panic on the next line, and that could confuse the test output

bitwiseguy: I think `require.NotNil` will exit the test execution immediately upon failure. This is different from `assert.NotNil`, which behaves the way you describe

edobry: nvm tested again and you're right, i'll make this change

PR #735, opened 2024-12-04
validation: move contract versions code from superchain to validation module

bitwiseguy:
The `superchain` module gets imported by `op-geth` and `op-node`, therefore it is advantageous to keep that module as small as possible. The contract versions code lived in the `superchain` module, but was only used by the `validation` module. Therefore it made sense to move this code around and get rid of the indirection

Additional change:
* separated `AddressList` code from `superchain/superchain.go` into its own `superchain/address_list.go` file to make it easier to navigate through the superchain module

### DIFF
--- a/validation/standard/versions.go
+++ b/validation/standard/versions.go
@@ -1,12 +1,136 @@
 package standard
 
 import (
+	"errors"
+	"fmt"
 	"reflect"
+	"strings"
 
 	"github.com/ethereum-optimism/superchain-registry/superchain"
+	"golang.org/x/mod/semver"
 )
 
 type Tag string
+
+type MappedContractProperties[T string | VersionedContract] struct {
+	L1CrossDomainMessenger       T `toml:"l1_cross_domain_messenger,omitempty"`
+	L1ERC721Bridge               T `toml:"l1_erc721_bridge,omitempty"`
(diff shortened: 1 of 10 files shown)


validation/standard/versions.go
```diff
@@ -1,13 +1,137 @@
...
+// ContractBytecodeHashes stores a bytecode hash against each contract
+type ContractBytecodeHashes MappedContractProperties[string]
+
+// VersionedContract represents a contract that has a semantic version.
+type VersionedContract struct {
+	Version string `toml:"version"`
+	// If the contract is a superchain singleton, it will have a static address
+	Address *superchain.Address `toml:"implementation_address,omitempty"`
+	// If the contract is proxied, the implementation will have a static address
+	ImplementationAddress *superchain.Address `toml:"address,omitempty"`
```
geoknee, line 47: The struct tags don't look quite right, here.

bitwiseguy:
Too many merge conflicts right now. Will reimplement this change in a different pr

validation/standard/versions.go
bitwiseguy: You are correct, although this pr did not introduce those tags. Rather they were just moved from their previous file so the mismatch already existed. That has since been fixed in https://github.com/ethereum-optimism/superchain-registry/pull/739

PR #749, opened 2024-12-06
Binary testnet holocene timestamp added

priyanshuthapliyal55:
# Adding a New Chain
This PR adds [Chain Name Here] to the registry.

## Checklist
- [x] I have declared the chain at the appropriate [Superchain Level](../docs/glossary.md#superchain-level-and-rollup-stage).


superchain/configs/sepolia/tbn.toml
```diff
@@ -8,8 +8,9 @@ batch_inbox_addr = "0xFf00000000000000000000000000000000042069"
 canyon_time = 0 # Thu 1 Jan 1970 00:00:00 UTC
 delta_time = 0 # Thu 1 Jan 1970 00:00:00 UTC
 ecotone_time = 0 # Thu 1 Jan 1970 00:00:00 UTC
-fjord_time = 1723204838  #Saturday June 8 2024 10:47:18 AM UTC
```
geoknee, line 11: Strange that these timestamp comments incorrect before -- @bitwiseguy do we know why?

superchain/configs/sepolia/tbn.toml
```diff
@@ -8,8 +8,9 @@ batch_inbox_addr = "0xFf00000000000000000000000000000000042069"
...
-granite_time = 1723545055 #Monday June 12 2024 1:57:35 PM UTC
+fjord_time = 1723204838  # Fri Aug 09 2024 12:00:38 UTC
+granite_time = 1723545055 # Tue Aug 13 2024 10:30:55 UTC
```
sebastianst, line 12: verified that the timestamps in the corrected comments are ✅

superchain/configs/sepolia/tbn.toml
```diff
@@ -8,8 +8,9 @@ batch_inbox_addr = "0xFf00000000000000000000000000000000042069"
...
+holocene_time = 1732633200 # Tue Nov 26 15:00:00 UTC 2024
```
sebastianst, line 13: Note that the Sepolia activation has already passed over a week ago. Did you locally set this with an override on all your nodes?

If that's intentional, you probably want to receive all future hardfork from Holocene onwards, so better just set the `superchain_time`.

```suggestion
superchain_time = 1732633200 # Holocene - Tue Nov 26 15:00:00 UTC 2024
```

superchain/configs/sepolia/tbn.toml
sebastianst: The old ones look like they were created manually, not with `date -u -r`.

superchain/configs/sepolia/tbn.toml
sebastianst: Setting `superchain_time` simply means that you inherit the activation times that are set for the superchain automatically. That also works for CGT chains.

sebastianst: But again, the Sepolia activation has already passed. Do you intentionally retroactively set the Sepolia Holocene activation time because you locally activated last week with overrides?

Issue #675, opened 2024-11-07
encode ProxyAdminOwner exceptions for Base and Unichain


tessr:
The Standard Rollup Charter, as ratified, has specific exemptions for the ProxyAdminOwner for Base and Unichain. These chains can have their PAO set to a 3/3 instead of the specified 2/2.

We should make sure that these chains have exceptions encoded in the SCR validation checks that match these requirements.

bitwiseguy:
@tessr a few questions for clarification:

1. Is there a single, standard address for the 3/3 that we should validate against for Base and Unichain ProxyAdminOwner, or is the address different for those two chains? Do you know what the address is?
2. Is the 3/3 only used for mainnet for Base and Unichain or on sepolia as well?
3. Should we enshrine the 3/3 address in the `standard-roles-*.toml` file(s) (e.g. [here](https://github.com/ethereum-optimism/superchain-registry/blob/main/validation/standard/standard-config-roles-mainnet.toml#L11))?

alfonso-op:
Currently blocked on Unichain launch, TBD on Base.

alfonso-op:
Currently blocked on Unichain launch, will become urgent once ready.

PR #764, opened 2024-12-10
feat(superchain,validation): base & unichain PAO exclusions

edobry:
### Description

This PR sets the `governed_by_optimism` field to `true` for Base and Unichain chains, and then excludes them from the `GovernedByOptimism` test.

### Metadata

- **In service of:** #675 
- **Depends on:** #718


tessr (changes requested):
This is probably on me for writing an ambiguous ticket, but we need to adjust the approach here before it gets merged. 

We should not auto-set these chains to be "governed by Optimism," rather we should check that their PAO is set to 3/3s that include the Optimism Security Council. We can hardcode those PAO values. 

Note that Base is actually not yet governed by Optimism either, so we should hold off on creating an exclusion for them. Again, apologies for ambiguity in that ticket, it's a bit more of a tracking item at the moment for Base.

PR #767, opened 2024-12-11
fix incorrect addresses for contracts which weren't released

geoknee:
this was a copypaste error from the addresses on sepolia. I fixed them to match the addresses from the previous release.


BlocksOnAChain (approved):
this looks good to me.

sebastianst (approved):
🙈

PR #768, opened 2024-12-11
Add consistency check for standard versions

### LINKED ISSUE #745: ensure version check consistency
In the [Standard Rollup Charter](https://gov.optimism.io/t/season-6-standard-rollup-charter-ratification/8135), we commit to relying on three key TOML files for the substance of our validation checks. They are:

1. Params: https://github.com/ethereum-optimism/superchain-registry/blob/main/validation/standard/standard-config-params-mainnet.toml 
2. Roles: https://github.com/ethereum-optimism/superchain-registry/blob/main/validation/standard/standard-config-roles-mainnet.toml
3. Versions: https://github.com/ethereum-optimism/superchain-registry/blob/main/validation/standard/standard-versions-mainnet.toml 

Our version checks _do_ use `standard-versions-mainnet.toml` to validate the contract versions. However, we also use [`standard-bytecodes.toml`](https://github.com/ethereum-optimism/superchain-registry/blob/main/validation/standard/standard-bytecodes.toml) to confirm that contracts are not only using the correct version number but are also consistent in their implementation. 

We should build some internal validation to ensure consistency between `standard-versions-mainnet.toml` and `standard-bytecodes.toml` (and likely `standard-immutables.toml` as well, since the bytecodes file relies on these immutables).

geoknee:
Closes #745 

This PR adds a consistency check on the standard versions TOML files. These files commit to:
* semver strings
* deployment addresses
* bytecode hashes
* immutable references (used to mask out the bytecode to compute a chain agnostic hash)

which were, are, or are about to become "standard". They are resolved by `op-contracts/va.b.c` tags.

Until now there is no consistency check on whether the contract deployed at the declared addressed actually have the semver and bytecode matching the declared values.

This PR adds that check. It is not a validation check on any particular chain, but a more traditional CI check which lives in the `standard` package. 

I did not yet spend a lot of time tidying up the test, and did some hacks to expose utilities in the rest of the repo that helped me get this done quickly. It was already enough to find a couple of bugs.


validation/standard/standard-immutables.toml
```diff
@@ -8,10 +8,10 @@
 #
 # "immutableReferences":{"85798":[{"start":178,"length":32},{"start":1771,"length":32}]}}
 #
-[releases."op-contracts/v1.8.0-rc.3"] # Holocene https://github.com/ethereum-optimism/optimism/releases/tag/op-contracts%2Fv1.8.0-rc.3
+["op-contracts/v1.8.0-rc.3"] # Holocene https://github.com/ethereum-optimism/optimism/releases/tag/op-contracts%2Fv1.8.0-rc.3
```
geoknee, line 11: This was a big footgun. In `standard-versions-*.toml`, the tables are nested under "releases", but in the other TOML files they are not.

validation/standard/standard-version-consistency_test.go
```diff
@@ -0,0 +1,74 @@
...
+	for _, superchain := range []string{"sepolia", "mainnet"} {
+		rpcEndpoint := Superchains[superchain].Config.L1.PublicRPC
+		require.NotEmpty(t, rpcEndpoint)
+
+		client, err := ethclient.Dial(rpcEndpoint)
+		require.NoErrorf(t, err, "could not dial rpc endpoint %s", rpcEndpoint)
+
+		for tag, release := range standard.NetworkVersions[superchain].Releases {
+			for _, contract := range release.GetNonEmpty() {
+				releaseVersion := release
```
bitwiseguy, line 31: What's the purpose of this line?

validation/standard/standard-version-consistency_test.go
```diff
@@ -0,0 +1,74 @@
...
+func TestStandardVersionConsistency(t *testing.T) {
+	for _, superchain := range []string{"sepolia", "mainnet"} {
+		rpcEndpoint := Superchains[superchain].Config.L1.PublicRPC
+		require.NotEmpty(t, rpcEndpoint)
+
+		client, err := ethclient.Dial(rpcEndpoint)
+		require.NoErrorf(t, err, "could not dial rpc endpoint %s", rpcEndpoint)
+
+		for tag, release := range standard.NetworkVersions[superchain].Releases {
+			for _, contract := range release.GetNonEmpty() {
```
bitwiseguy, line 30: In general it seems like this `GetNonEmpty` method is dangerous because we don't actually know what we're checking for any given chain/release. Could we define some clear definitions of valid contract sets (e.g. known expected contracts for [1] permissioned fault-proofs, [2] permissionless fault-proofs [3] non-fault proofs) ? Or are there too many combinations of contract sets to make that feasible?

bitwiseguy: Similar function used for `L1ContractBytecodeHashes.GetNonEmpty()`

validation/standard/standard-version-consistency_test.go
```diff
@@ -0,0 +1,74 @@
...
+					addressToCheck = vc.Address
+				}
+				if vc.ImplementationAddress != nil {
+					addressToCheck = vc.ImplementationAddress
+				}
+				if addressToCheck != nil {
+					version, err := validation.GetVersion(context.Background(), common.Address(*addressToCheck), client)
+					if err != nil {
+						t.Fatal(err, "could not get version for %s", addressToCheck, "contract", contract, "superchain", superchain, "release", tag)
+					}
```
bitwiseguy, line 50: This is valid code, but curious why you wouldn't use `require.NoError` here instead of `t.Fatal` for consistency with rest of the test

validation/standard/standard-version-consistency_test.go
geoknee: No purpose I think, we should be able to remove it.

validation/standard/standard-immutables.toml
bitwiseguy: Any reason we couldn't/shouldn't do the same for `standard-versions-mainnet.toml` and `standard-versions-sepolia.toml`?

validation/standard/standard-version-consistency_test.go
bitwiseguy: Can address this in a separate pr since its a large refactor outside the scope of this work

validation/standard/standard-immutables.toml
bitwiseguy: Done in https://github.com/ethereum-optimism/superchain-registry/pull/768/commits/656b480c77d31e90c5663f42a64cb1d8be5a855c

justfile
```diff
@@ -39,7 +39,7 @@ test-superchain: clean-ops
 
 # Unit test all Go code in the validation module, and do not run validation checks themselves
 test-validation: clean-ops
-	TEST_DIRECTORY=./validation/... go run gotest.tools/gotestsum@latest --format testname -- -run '[^TestValidation|^TestPromotion|^TestGenesisAllocs]'
+	TEST_DIRECTORY=./validation/... go run gotest.tools/gotestsum@latest --format testname -- -count=1 -run '[^TestValidation|^TestPromotion|^TestGenesisAllocs]'
```
geoknee, line 42: Do we want to commit this?

validation/standard/standard-version-consistency_test.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+// This test will check:
+// For each each superchain
+// For each contract release
+// For each contract version declaration which specifies an address
+// There is a contract at that address with the matching bytecode and semver.
+// This is a consistency check on the standard package itself, not any particular chain
+func TestStandardVersionConsistency(t *testing.T) {
+	for _, superchain := range []string{"sepolia", "mainnet"} {
```
geoknee, line 22: Can we make this less brittle so that if/when a new superchain is added it automatically works?

validation/standard/versions.go
```diff
@@ -52,14 +53,38 @@ type VersionTags struct {
 
 var (
 	Release            Tag
-	NetworkVersions                           = make(map[string]VersionTags)
+	ContractVersions                          = make(map[string]map[Tag]superchain.ContractVersions)
```
geoknee, line 56: Can we delete the `VersionTags` type just above?

validation/standard/versions.go
```diff
@@ -52,14 +53,38 @@ type VersionTags struct {
...
 	BytecodeHashes     BytecodeHashTags       = make(BytecodeHashTags, 0)
 	BytecodeImmutables BytecodeImmutablesTags = make(BytecodeImmutablesTags, 0)
 )
 
 // L1ContractBytecodeHashes represents the hash of the contract bytecode (as a hex string) for each L1 contract
 type L1ContractBytecodeHashes superchain.ContractBytecodeHashes
 
+func (bch L1ContractBytecodeHashes) GetBytecodeHashFor(name string) (string, error) {
```
geoknee, line 64: Perhaps a unit test for this?

validation/standard/versions.go
bitwiseguy: Good call. Done here: https://github.com/ethereum-optimism/superchain-registry/pull/768/commits/f7a0a3a55765aba4dd795dc700938bfc44a46e03

opgitgovernance:
Detected nonzero diff in rollup config. This PR modifies the configuration of at least one chain in the registry, which may or may not have been intended. You can see the detail [here](https://circleci.com/gh/ethereum-optimism/superchain-registry/52926).

opgitgovernance:
Detected nonzero diff in genesis. This PR modifies the genesis data of at least one chain in the registry, which may or may not have been intended. You can see the detail [here](https://circleci.com/gh/ethereum-optimism/superchain-registry/52925).

justfile
bitwiseguy: We don't have consistency with that flag: we use it for test-ops but not test-superchain or test-validation. Will remove it for now

geoknee:
We are seeing diffbot comments on this PR, but this is because the branch is out of date with `main` (this is a known issue with the CI job). If we update the PR the comments should disappear.

validation/standard/versions.go
bitwiseguy: Done in https://github.com/ethereum-optimism/superchain-registry/pull/768/commits/0647af459a5f5c57ae76575458cae111c644d4fd

validation/standard/standard-version-consistency_test.go
bitwiseguy: Can do this in a follow-up. We have the same brittleness in a few other spots

validation/standard/standard-versions-sepolia.toml
```diff
@@ -1,12 +1,10 @@
-[releases]
-
 # Contracts which are
 #   * unproxied singletons: specify a standard "address"
 #   * proxied             : specify a standard "implementation_address"
 #   * neither             : specify neither a standard "address" nor "implementation_address"
 
 # Holocene https://github.com/ethereum-optimism/optimism/releases/tag/op-contracts%2Fv1.8.0-rc.4
-[releases."op-contracts/v1.8.0-rc.4"]
+["op-contracts/v1.8.0-rc.4"]
```
geoknee, line 7: This is a breaking change, with respect to at least the Holocene upgrade scripts. I think we should still do it, but coordinate with this PR https://github.com/ethereum-optimism/optimism/pull/13412 to cope with the change.

bitwiseguy (approved):

Issue #800, opened 2024-12-16
update gas limit checks for Base


tessr:
Base has experimented with pushing its gas limit beyond what is specified in the standard config. 

Currently, this is fine: Since they’re not yet a standard chain, they aren’t bound by the recommended gas limit.

Long term, this should remain acceptable: The gas limit guideline exists to ensure chains maintain performance. Base has shown a commitment to responsible scaling, and their research will help us refine the standard config limits when they aim to become a standard chain.

The [most recent update](https://x.com/jessepollak/status/1867047210938143172) to the Base gas limit *has* officially pushed them over what’s permitted in the standard config. We’re currently checking this aggressively (but unnecessarily) for Base, so this is causing CI failures for the whole SCR. 

To fix this, we should add an exception for Base to the gas limit checks, with a clear TODO to remove the exception when Base decides to adhere to the standard charter. We would like to maintain Base's status as a standard candidate in order to reflect their commitment to eventually becoming a standard chain.

bitwiseguy:
One nuance here is that base is likely using the `GETH_MINER_EFFECTIVEGASLIMIT` override (or `reth` equivalent) such that the `SystemConfig.gasLimit()` is not what is actually capping their gas per block. After smart contract changes introduced by holocene, that env var override should be deprecated and we can read everything from the `SystemConfig` contract.

Related: I think we should also be validating standard chains' per block gas _target_ in addition to the gas _limit_. And combining that with the block frequency for performance standards/limits. One thing that restricts gas limit is what is fault-provable within a given block/batch. I believe gas target is more closely related to on-going performance.

PR #801, opened 2024-12-17
Gas validation: Add an exception for Base

vdamle:
As noted in issue#800, we are adding an exception for the gas validation check for Base with an explicit directive for it to be removed when Base adheres to the standard charter.

# Adding a New Chain
N/A

## Checklist
- [ ] ~I have declared the chain at the appropriate [Superchain Level](../docs/glossary.md#superchain-level-and-rollup-stage).~

Ref: https://github.com/ethereum-optimism/superchain-registry/issues/800


bitwiseguy:
Could we move the exception [here](https://github.com/ethereum-optimism/superchain-registry/blob/main/validation/exclusions_test.go#L53-L56) instead? Its nice to have a central place for all exceptions/exclusions

vdamle:
> Could we move the exception [here](https://github.com/ethereum-optimism/superchain-registry/blob/main/validation/exclusions_test.go#L53-L56) instead? Its nice to have a central place for all exceptions/exclusions

Thanks @bitwiseguy . Updated now

bitwiseguy (approved):

PR #802, opened 2024-12-17
validation: decouple GasLimit tests

bitwiseguy:
Closes https://github.com/ethereum-optimism/superchain-registry/issues/772


validation/gas_limit_test.go
```diff
@@ -0,0 +1,55 @@
...
+func TestCheckGasLimit(t *testing.T) {
+	mockClient := &ethclient.Client{}
+	mockChain := &superchain.ChainConfig{
+		ChainID:    10,
+		Superchain: "mainnet",
+	}
+
+	// Restore the original function after the test
+	originalFunc := getGasLimitWithRetriesFunc
+	defer func() { getGasLimitWithRetriesFunc = originalFunc }()
```
geoknee, line 24: I wonder if there's an alternative to this which avoids mutating global variables. 

What if we pass some optional / variadic `opts` to `CheckGasLimit`, which can override the default `getGasLimitWithRetriesFunc` with an injected mock but can also be left blank and the default then applies. wdyt?

bitwiseguy: I am trying to keep the `Check*` function signature the same across all validations in case we want to create an interface for it later. Would the variadic `opts` require every `Check*` function to have awareness of the `getGasLimitWithRetriesFunc` even though `CheckGasLimit` is the only one that will use it?

I think the global default/override is functionally equivalent to the using `opts` and overriding by checking if the opt field is set. Or do you feel like the global override is more dangerous?

geoknee: Here's roughly what I had in mind https://github.com/ethereum-optimism/superchain-registry/pull/804. 

I think it is functionally equivalent, but IMO it's a better pattern to inject a mock by passing in to a function rather than mutating a global var. There could be some nasty race conditions if we ran more tests in parallel for example.

bitwiseguy: Yea good point about test race conditions. 

What about this part:
> Would the variadic opts require every Check* function to have awareness of the getGasLimitWithRetriesFunc even though CheckGasLimit is the only one that will use it?

Do you think we should aim to have same function signature across all `Check*` so that all validations can implement the same interface, or not much value in that?

bitwiseguy: Actually I wonder if we might want to use this instead:
```
func NewValidationSuite(overrides validateOverrides) (validationSuite, error)
```
That will return a `validationSuite` with all global defaults unless we set any of the overrides in the input arg. Then all of the helper functions and attributes can be fields on the `validationSuite` type. And each validation can be its own method. Then we can have methods for each validation classification:
```
func (s *validationSuite) ValidateStandard(chain *ChainConfig) []error {
  var errs []error
  if err := s.checkSuperchainConfig(chain); err != nil {
    errs := append(errs, err)
  }
  if err := s.checkForStandardVersions(chain); err != nil {
    errs := append(errs, err)
  }

  return errs
}
```
That way we can share dependencies, both functions and data objects like `l1Client` and `standard.Config`, across all validations for a each chain. And pass in mocks for all unit testing. And it would allow us to run these validations as part of the op-node startup.

Biggest downside of that design might be that it makes it more difficult to run all validations for all chains in parallel, but it sounds like one of the goals is to get rid of hourly validation ci jobs anyways.

geoknee: From our chat today:

* we are aligned on the idea of a `validationSuite` struct which stores things which can easily mocked by injecting mocks when constructing the `validationSuite`. 
* one idea for nice interfaces to mock would be either a interface for each contract (e.g. `SystemConfig`) with the getters we need only (and we can delete the go bindings that we have); or even a single interface `L1Contracts` with a getter for each property we need e.g. `getGasLimitFromSystemConfig()`/`getGasLimit()`/`getSystemConfigOwner()`
* ideally these methods are _async_ for example returning a channel where the result will be sent instead the result itself. This will allow us to write async (or sync) code around the interfaces easily. But, we can add this later as an optimization if that speeds up development
* open minded about how to get to this ideal state. Current plan is:
  - first decouple validation code from testing framework
  - add unit tests for the validation code, adding mocks which are passed in to the validation functions
  - finally bring it all together and make the functions methods on a struct so they can access the getters without them having to be passed in (gives a cleaner API). 
* Instead of the `ValidateStandard` method returning `[]error` it could return a struct with named fields. Gives us some type safety / warnings when we change the interface
* `ValidationSuite` could also embed the chain rather than the chain being passed in. Not sure the pros and cons yet.

justfile
```diff
@@ -39,7 +39,7 @@ test-superchain: clean-ops
 
 # Unit test all Go code in the validation module, and do not run validation checks themselves
 test-validation: clean-ops
-	TEST_DIRECTORY=./validation/... go run gotest.tools/gotestsum@latest --format testname -- -run '[^TestValidation|^TestPromotion|^TestGenesisAllocs]'
+	TEST_DIRECTORY=./validation/... SKIP_TESTMAIN=true go run gotest.tools/gotestsum@latest --format testname -- -tags "" -run '[^TestValidation|^TestPromotion|^TestGenesisAllocs]'
```
geoknee, line 42: Should we rename `SKIP_TESTMAIN` to something like `SKIP_GENESIS_ALLOCS`? I'm unsure about what the variable is for with this generic name. And does this replace the `^TestGenesisAllocs` filter?

validation/gas_limit_test.go
```diff
@@ -0,0 +1,62 @@
...
+func testGasLimit(t *testing.T, chain *superchain.ChainConfig) {
+	rpcEndpoint := superchain.Superchains[chain.Superchain].Config.L1.PublicRPC
+	require.NotEmpty(t, rpcEndpoint)
+
+	client, err := ethclient.Dial(rpcEndpoint)
+	require.NoErrorf(t, err, "could not dial rpc endpoint %s", rpcEndpoint)
+
+	err = CheckGasLimit(chain, client, nil)
+	require.NoError(t, err)
+}
```
geoknee, line 24: Wondering if this should live in `gas_limit.go` or not? Don't feel strongly but it might be nice to clearly separate the new unit tests from the hook for the existing validation suite.

geoknee (approved):
👏

validation/gas_limit_test.go
bitwiseguy: I don't feel strongly either. I changed it in https://github.com/ethereum-optimism/superchain-registry/pull/802/commits/1e73866041c407e6399bdb63ea5613b901a3513d. Will revert that change

PR #803, opened 2024-12-17
validation: decouple GasToken tests

bitwiseguy:
Closes https://github.com/ethereum-optimism/superchain-registry/issues/773


validation/testutils/mock_ethclient.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+// MockEthClient is a mock that implements EthCaller for testing.
+type MockEthClient struct {
+	// Setup response maps keyed by method 4-byte signature + address, or
+	// just by method signature if that's simpler for your tests.
+	Responses map[string][]byte
+	Err       error
+}
```
bitwiseguy, line 17: This can be reused by many of the validation unit tests so I extracted it into this `testutils` package where we can put other shared test mocks/helpers

validation/gas_token.go
```diff
@@ -0,0 +1,104 @@
...
+func testGasToken(t *testing.T, chain *superchain.ChainConfig) {
+	l1Client, err := ethclient.Dial(superchain.Superchains[chain.Superchain].Config.L1.PublicRPC)
+	require.NoError(t, err, "Failed to connect to the Ethereum client at RPC url %s", chain.PublicRPC)
+	defer l1Client.Close()
+
+	err = CheckGasToken(chain, l1Client)
+	require.NoError(t, err)
+}
+
+func CheckGasToken(chain *superchain.ChainConfig, l1Client *ethclient.Client) error {
```
bitwiseguy, line 27: Consider using `EthClient` interface here instead of `*ethclient.Client`

validation/gas_token.go
```diff
@@ -0,0 +1,104 @@
...
+		// Pre: reverting is acceptable
+		return err
+	} else {
+		// Post: must be set to false
+		if isCustomGasToken {
+			return fmt.Errorf("L1Block.isCustomGasToken() must return false")
+		}
+	}
+
+	isCustomGasToken, err = getBool("isCustomGasToken()", superchain.Addresses[chain.ChainID].SystemConfigProxy, l2Client)
```
geoknee, line 58: Shouldn't this use an l1Client?

I'd like us to revisit the "reverting is acceptable" case b/c it seems like it could be hiding many skeletons like this one. Can we double check it?

validation/gas_token.go
```diff
@@ -0,0 +1,104 @@
...
+	l1Client, err := ethclient.Dial(superchain.Superchains[chain.Superchain].Config.L1.PublicRPC)
+	require.NoError(t, err, "Failed to connect to the Ethereum client at RPC url %s", chain.PublicRPC)
+	defer l1Client.Close()
+
+	err = CheckGasToken(chain, l1Client)
+	require.NoError(t, err)
+}
+
+func CheckGasToken(chain *superchain.ChainConfig, l1Client EthClient) error {
+	l2Client, err := ethclient.Dial(chain.PublicRPC)
```
geoknee, line 28: Shouldn't we pass in a l2Client so that we can mock it?

validation/gas_token.go
```diff
@@ -0,0 +1,104 @@
...
+	} else {
+		// Post: must be set to false
+		if isCustomGasToken {
+			return fmt.Errorf("SystemConfigProxy.isCustomGasToken() must return false")
+		}
+	}
+	return nil
+}
+
+func getBytes(method string, contractAddress superchain.Address, client EthClient) ([]byte, error) {
```
geoknee, line 71: Could these functions would become methods on a new client type that we could define? Then they can be reused by the mock and by the real eth client?

For example, we could call it a `HandyClient` and it embeds an `EthClient`, exposing more ergonomic methods to the user. But it's constructor takes the interface so we can inject a mock to make a `FakeHandyClient`.

validation/gas_token_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/ethereum-optimism/superchain-registry/validation/testutils"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetBytes_Success(t *testing.T) {
+	t.Parallel()
+	mockClient := &testutils.MockEthClient{
+		Responses: map[string][]byte{
+			string(testutils.MethodID("myMethod()")): {0xde, 0xad, 0xbe, 0xef},
+		},
```
geoknee, line 17: I wonder if we could have the mock expose a method for adding a mocked response, so that you could just do `mockClient.Prime("myMethod", {0xde, 0xad, 0xbe, 0xef})`.

validation/gas_token_test.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	contractAddress := superchain.MustHexToAddress("0x1234567890abcdef1234567890abcdef12345678")
+	result, err := getBytes("myMethod()", contractAddress, mockClient)
+	require.NoError(t, err)
+	require.Equal(t, []byte{0xde, 0xad, 0xbe, 0xef}, result)
```
geoknee, line 22: Nit: could use a constant for this byte slice literal.

validation/gas_token.go
```diff
@@ -0,0 +1,104 @@
...
+func testGasToken(t *testing.T, chain *superchain.ChainConfig) {
+	l1Client, err := ethclient.Dial(superchain.Superchains[chain.Superchain].Config.L1.PublicRPC)
+	require.NoError(t, err, "Failed to connect to the Ethereum client at RPC url %s", chain.PublicRPC)
+	defer l1Client.Close()
+
+	err = CheckGasToken(chain, l1Client)
+	require.NoError(t, err)
+}
+
+func CheckGasToken(chain *superchain.ChainConfig, l1Client EthClient) error {
```
geoknee, line 27: Should we pass clients by pointer?

validation/gas_token_test.go
geoknee: (And similarly for the other tests).

validation/gas_token_test.go
```diff
@@ -0,0 +1,97 @@
...
+	require.NoError(t, err)
+	require.Equal(t, "00112233", hexVal)
+}
+
+func TestGetHexString_Error(t *testing.T) {
+	t.Parallel()
+	mockClient := &testutils.MockEthClient{
+		Err: errors.New("getHexString error"),
+	}
+	contractAddress := superchain.MustHexToAddress("0x1234567890abcdef1234567890abcdef12345678")
```
geoknee, line 54: As discussed, this address is not really used by the mock at all. In a future improvement, we could allow the mock to behave differently for different addresses (using say a nested map).

validation/gas_token.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	callContract := func(msg ethereum.CallMsg) ([]byte, error) {
+		return client.CallContract(context.Background(), msg, nil)
+	}
+
+	return Retry(callContract)(callMsg)
+}
+
+func getHexString(method string, contractAddress superchain.Address, client EthClient) (string, error) {
```
geoknee, line 85: These could be moved to `utils`, along with the unit tests?

validation/gas_token.go
geoknee, line 27: Did we add a unit test for this yet? That is what I would expect in `gas_token_test.go`.

validation/gas_token_test.go
bitwiseguy: I'm going to refactor this whole `MockEthClient` following the example in the monorepo [here](https://github.com/ethereum-optimism/optimism/blob/develop/op-service/eth/transactions_test.go#L14-L29). I wrote the monorepo one 8 months ago but had forgotten where it lived 😆 .

That mock pattern eliminates the need for the custom `Responses` mapping referenced here

validation/gas_token.go
bitwiseguy: Yea this should be an l1Client. Good catch. What are the pre/post comments referring to? Pre-what? Those comments were just copied from existing code

validation/gas_token.go
bitwiseguy: Yea will move the l2Client connection into the test wrapper then pass it in here

validation/gas_token.go
geoknee: I think what this is trying to do is to cope with deployments of the `SystemConfig` which don't expose the `isCustomGasToken()` method. This would indicate they don't support custom gas tokens and therefore should automatically pass the test. A better approach would be to read the semver here, and unless it is greater than or equal to some version (which we would need to look up or ask someone for) we don't do the rest of the check.

validation/gas_token.go
bitwiseguy: l1Client is already a pointer here. The `EthClient` type is an interface with pointer methods. Interfaces are always passed by value

validation/gas_token.go
bitwiseguy: Maybe there is a semver for the `SystemConfigProxy` we could check? 

If its < a certain value, we don't make the following calls at all (because the method doesn't exist?):
* SystemConfigProxy.isCustomGasToken()
* L1BlockPredeploy.isCustomGasToken()

validation/gas_token.go
bitwiseguy: I'm open to that idea, but that seems like follow-up work that could be addressed after the test decoupling when we create the `ValidationSuite` struct

validation/gas_token.go
bitwiseguy: They could be, although they are only used by this one validation so there's not much value in moving them to a common file. I'm inclined to leave them here so that helpers specific to one validation are kept close to the validation logic

validation/gas_token_test.go
bitwiseguy: This is now enabled by the introduction of a new mock client type in https://github.com/ethereum-optimism/superchain-registry/pull/803/commits/a5cb887436406788a4296e13a926da5cc528e4f6

validation/gas_token.go
bitwiseguy: Discussed some findings on semver in [this discord thread](https://discord.com/channels/1244729134312198194/1319743226319474731/1319780018804756501). Not sure how to proceed at the moment. The `isCustomGasToken()` method seems to be missing on the latest standard version of the `SystemConfig` contract

bitwiseguy:
Closing in favor of https://github.com/ethereum-optimism/superchain-registry/pull/816

PR #804, opened 2024-12-17
use variadic overrides trick

geoknee:


bitwiseguy:
These ideas included in https://github.com/ethereum-optimism/superchain-registry/pull/802

Issue #765, opened 2024-12-11
Push SuperchainConfig addresses into each chain's config toml file


geoknee:
This address can be read from the `OptimismPortal`. We should
* do that manually for all existing chains, and 
* update the `sc-ops`/`add-chain` tool to scrape it when a chain is added (either from chain or from the deployments artifact)
* write a validation check for standard chains, that their superchain config address must match the superchain wide address
             
             
              I believe it is there for the convenience of downstream applications / other users. Those apps may not wish to discover addresses on chain -- it's worse than just going over multiple files since the `SuperchainConfig` can actually vary per chain and not match the standard superchain address (we only validate standard chains are compliant). 

I think the codegen here is a bit of a distraction, it will contain the same information as the individual TOML files. It is just that some of the individual TOML files are missing the `SuperchainConfig` address.

It sounds like the better change might be to backfill missing addresses then, and add a validation check for all chains that the in chain references resolve to the ones in the registry. Would you be in favor of this @sebastianst @mds1 ? I'll push this back to a draft PR until we can align.

_Originally posted by @geoknee in https://github.com/ethereum-optimism/superchain-registry/pull/701#discussion_r1852615370_

geoknee:
This aligns with the approach of simplifying things like we plan to with #715.

We should also see the `SuperchainConfig` appear in the codegen'ed `addreses.json` file.

geoknee:
Related problem that we should clear up: 

https://github.com/ethereum-optimism/superchain-registry/blob/ad6f08a52802ec7a55bb4a91d57937551504885c/superchain/configs/mainnet/superchain.toml#L3

is actually the adderess of the **proxy** so the name should reflect that.

Issue #751, opened 2024-12-06
Consider storing all hardfork times explicitly in each chain's config file


geoknee:
I've been wondering if we can leverage/extend the existing codegen tooling to help keep this stuff tidy. It could:
1. throw an error if folks have set superchain_time in a way that doesn't make sense.
2. actually write all active hardforks into each chain's TOML file using the existing algorithm for resolving them from superchain-wide values + superchain_time + any overrides. We can then do the usual git diff --exit-code to ensure the modified TOMLs are checked in and reproducible.

Doing 2 has the advantage that hardfork times are no longer implicit / the result of an opaque resolution process, but there for all to see in the raw TOML files. We could also kill the diffbots (which are there to compute essentially the same thing) which are proving difficult to maintain in a secure way.
When we set e.g. the Isthmus time, we run codegen and it changes all existing chains' TOML files if appropriate. We see the diff in the actual Github diff, and don't need a bot to compute it and comment on the PR with the result.

sebastianst:
If this helps overall, I'm supportive. My main opposition is that we duplicate information from the `superchain.toml` into individual chains' toml files. We'd only do this for fork activation times, but not for other fields in the `superchain.toml` file, which seems inconsistent.

What I also like about 1. is that this would mean we only see explicit diffs to superchain activation times in individual chains' configs, so only overrides that don't equal the superchain time. When we do 2. it's not obvious from looking at an individual chain's config which is an override changing the superchain time, and which is just a copy of a superchain activation.

geoknee:
Maintaining our CI config "diff bots" has also proven to be more work than anticipated, and this change would likely remove the need for those diff bots.

PR #810, opened 2024-12-19
superchain: write all hardforks, SuperchainConfig address to config .tomls

bitwiseguy:
## Overview

To automate the `SuperchainConfig` address addition to each chain config file, I temporarily added the following code to `/ops/cmd/update-configs.go`:
```
rpcEndpoint := superchain.Superchains[chain.Superchain].Config.L1.PublicRPC
if rpcEndpoint == "" {
	panic("l1 PublicRPC missing")
}

address, err := validation.CastCall(superchain.Addresses[chain.ChainID].OptimismPortalProxy, "superchainConfig()(address)", nil, rpcEndpoint)
if err != nil {
	fmt.Printf("cast error, chainID: %d, chainName: %s\n", chain.ChainID, chain.Name)
	continue
}
fmt.Printf("cast result: %s, chainID: %d\n", address, chain.ChainID)
chain.Addresses.SuperchainConfig = superchain.MustHexToAddress(address[0])
```
Then I ran the following command, which was already in place to populate the hardfork timestamps in each chain config .toml:
```
go run ./ops update-configs
```
This same command can be run anytime we introduce a new hardfork activation timestamp in one of the `superchain.toml` files and want to write that value to config files for chains that have opted-in to inherit that timestamp (i.e. chains that have set `superchain_time` in their chain config .toml).

## Metadata
* Closes https://github.com/ethereum-optimism/superchain-registry/issues/751
* Closes https://github.com/ethereum-optimism/superchain-registry/issues/765


bitwiseguy:
Orderly and Lyra have old versions of `OptimismPortalProxy` and I was not able to read the `SuperchainConfig` address from that contract. Should I assume that the `superchain_config_addr` in the `superchain.toml` file is the correct one to use for those chains? Or is there another way to find the address that is set? Or is it possible they're not using a `SuperchainConfig` at all and the address should remain absent in their chain configs?

PR #816, opened 2024-12-28
Superchain Registry 2.0

mslipper:
See [here](https://docs.google.com/document/d/14XgLo7WIhclmz3Zsfp3IICJ3rn-KsWb7-T_f7LqhIXs/edit?tab=t.0#heading=h.ovm15kfzrdr8) for context.


superchain/configs/mainnet/arena-z.toml
```diff
@@ -1,21 +1,22 @@
 name = "arena-z"
-chain_id = 7897
 public_rpc = "https://rpc.arena-z.gg"
 sequencer_rpc = "https://rpc.arena-z.gg"
 explorer = "https://explorer.arena-z.gg"
 superchain_level = 0
 governed_by_optimism = true
-standard_chain_candidate = true # This is a temporary field which causes most of the standard validation checks to run on this chain
+data_availability_type = "eth-da"
+chain_id = 7897
```
geoknee, line 8: From memory we may have wanted the chain_id nearer the top of the file (for easier human readability). I don't feel strongly but wanted to flag for @tessr's input.

superchain/configs/mainnet/arena-z.toml
```diff
@@ -36,14 +37,16 @@ data_availability_type = "eth-da"
...
+[roles]
   SystemConfigOwner = "0xBeA2Bc852a160B8547273660E22F4F08C2fa9Bbb"
   ProxyAdminOwner = "0x5a0Aae59D09fccBdDb6C6CcEB07B7279367C3d2A"
   Guardian = "0x09f7150D8c019BeF34450d6920f6B3608ceFdAf2"
   Challenger = "0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A"
   Proposer = "0x5f16E66D8736B689a430564a31c8d887ca357CD8"
   UnsafeBlockSigner = "0xb774Ca8438319d2a97B9925F4CD248e4C470Ac5B"
   BatchSubmitter = "0x2b8733E8c60A928b19BB7db1D79b918e8E09AC8c"
+
+[addresses]
```
bitwiseguy, line 49: Could we use the term `contracts` instead of `addresses` here to disambiguate between this and `roles` (since roles are also addresses)?

superchain/configs/mainnet/automata.toml
```diff
@@ -1,27 +1,30 @@
 name = "Automata Mainnet"
-chain_id = 65536
 public_rpc = "https://rpc.ata.network"
 sequencer_rpc = "https://automata-mainnet.alt.technology/"
 explorer = "https://explorer.ata.network"
 superchain_level = 0
+governed_by_optimism = false
+data_availability_type = "alt-da"
+chain_id = 65536
 batch_inbox_addr = "0xff00000000000000000000000000000001111111"
```
geoknee, line 9: Could be saved for future work but while we're in here -- we could moved batch inbox address and gas paying token to `addresses` table? Or the `addresses` table gets renamed to something more specific such as `l1_contract_addresses`.

superchain/configs/mainnet/automata.toml
```diff
@@ -1,27 +1,30 @@
...
+  fjord_time = 0 # Thu 1 Jan 1970 00:00:00 UTC
+
 [optimism]
   eip1559_elasticity = 10
   eip1559_denominator = 50
   eip1559_denominator_canyon = 250
 
 [alt_da]
-  da_challenge_contract_address = "0x08c5DCDD5e46d31CC1591ee15b084663507597f3"
+  da_challenge_contract_address = "0x08c5dcdd5e46d31cc1591ee15b084663507597f3"
```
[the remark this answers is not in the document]
geoknee: We should probably also reinstate a CI check that all addresses are checksummed? Previously, we checked this via the go bindings at init-time.

chainList.json
```diff
@@ -1,40 +1,4 @@
 [
-  {
```
geoknee, line 2: Was it intentional to make changes / reorder this file?

ops/cmd/apply_hardforks/main.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	cfgs, err := manage.CollectChainConfigs(paths.SuperchainDir(wd, superchain))
+	if err != nil {
+		return fmt.Errorf("error collecting chain configs: %w", err)
+	}
+
+	for _, cfg := range cfgs {
+		chainCfg := cfg.Config
+		if chainCfg.SuperchainTime != nil && int64(*chainCfg.SuperchainTime) < time.Now().Unix() {
```
geoknee, line 70: Is it necessary to restrict this to when `SuperchainTime` is in the past? I think chains ought to be able to set it to the the future and things work as expected? I think in general we shouldn't depend on `Now` at all. 

Reference for `superchain_time` semantics:  https://github.com/ethereum-optimism/superchain-registry/blob/main/docs/hardfork-activation-inheritance.md

ops/go.work
```diff
@@ -0,0 +1,7 @@
+go 1.23.1
+
+use (
+	.
+	../validation
+)
```
geoknee, line 6: Not sure what the purpose of this file is, haven't seen it used in this way before. We did have one previously but it was in the root of the repo. Can you clarify?

ops/internal/config/hardforks_test.go
```diff
@@ -0,0 +1,60 @@
...
+		})
+	}
+}
+
+func TestCopyHardforks(t *testing.T) {
+	a := &Hardforks{
+		CanyonTime: NewHardforkTime(1),
+		DeltaTime:  NewHardforkTime(2),
+	}
+	b := &Hardforks{
```
geoknee, line 50: Nit: if we used `src` and `dest` this would be slightly easier to read since there is an implicit direction here.

.gitignore
```diff
@@ -31,3 +31,7 @@ go.work.sum
 
 # editor specific config
 .vscode
+
+ops/bin
+
+.staging
```
bitwiseguy, line 37: I'm confused by this. If this directory is gitignored, how does the user add their intent.toml and state.json file to it and push it as part of a pr? Or is that not the intended user flow for generating the report?

ops/internal/manage/staging_test.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/ethereum-optimism/superchain-registry/ops/internal/config"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCopyDeployConfigHFTimes(t *testing.T) {
+	a := &genesis.UpgradeScheduleDeployConfig{
+		// include regolith to demonstrate how unknown destination fields
+		// are not copied
+		L2GenesisRegolithTimeOffset: new(hexutil.Uint64),
```
geoknee, line 16: Do we want to silently ignore fields like this? I want to make sure we would catch a situation where the next hardfork is missing from the SCR structs.

ops/internal/manage/chains.md.tmpl
```diff
@@ -0,0 +1,11 @@
+# Superchain Ecosystem
+
+{{ range $index, $superchain := .Superchains -}}
+### {{ $superchain }}
+
+| Chain Name | OP Governed | Upgradeable | Explorer | Public RPC | Sequencer RPC
+|---|---|---|---|---|---|
+{{- range index $.ChainData $index }}
+| {{ .Name }} | {{ checkmark .GovernedByOptimism }} | {{ optedInSuperchain .SuperchainTime }} | {{ .Explorer }} | `{{ .PublicRPC }}` | `{{ .SequencerRPC }}` |
```
bitwiseguy, line 9: How does the `chainConfig.GovernedByOptimism` value get set now? Will that be part of the chain's intent.toml or state.json file? I'm wondering if we should remove that boolean from the ChainConfig struct and dynamically determine if the chain is currently governed by optimism when generating the CHAINS.md file instead of relying on a static field

justfile
```diff
@@ -1,93 +1,51 @@
-set positional-arguments
+set positional-arguments := true
+set dotenv-load := true
+
 alias t := test-all
 alias l := lint-all
 
-# Adding a chain
-add-chain:
```
bitwiseguy, line 6: What is the complete user flow that replaces this command? I think the answer to that will help me review this pr overall. Is there any op-deployer in-flight work that needs to be merged to make it compatible with the SR 2.0? I noticed [this op-deployer pr](https://github.com/ethereum-optimism/optimism/pull/13519) was closed but not sure if there is a replacement

chainList.json
mslipper: The order was not consistently applied, so I updated the file to list all chains by name.

ops/cmd/apply_hardforks/main.go
mslipper: According to the linked doc, this field functions like a hardfork activation timestamp:

> All superchain-wide hardfork activations will be inherited **starting from this timestamp.**

If we don't check `time.Now()` then we will be applying superchain-wide hardfork activations prior to the timestamp in `superchain_time`.

ops/go.work
mslipper: `ops` needs it so that local changes in the validation package carry through without requiring manual version bumps.

ops/internal/manage/staging_test.go
mslipper: I can update this so that it panics.

ops/internal/manage/chains.md.tmpl
mslipper: It's set in `meta.toml`.

superchain/configs/mainnet/automata.toml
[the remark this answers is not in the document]
mslipper: If the checksums are important I can bring them back.

ops/internal/report/l1.go
```diff
@@ -0,0 +1,327 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get chain ID: %w", err)
+	}
+
+	opcmAddr, err := standard.ManagerImplementationAddrFor(chainID.Uint64(), release.Tag)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get OPCM address: %w", err)
+	}
+
+	receipt, err := client.TransactionReceipt(ctx, deploymentTx)
```
bitwiseguy, line 93: Instead of parsing the `DeployedEvent`, which just contains information about the addresses at a previous fixed point in time, would it be better for the `report` command to query the `SystemConfig` contract and gather all the contract addresses and config info needed through that? That way the report would give info for the current state of the chain instead of potentially outdated information. Could grab the `SystemConfigProxy` address from the chain config stored in the superchain-registry.

A few advantages to removing dependence on deployment tx hash:
1. Chains that did not deploy via op-deployer can still generate a report
2. Consumers of the report (e.g. a superchain health dashboard) can continually generate real-time reports based on current chain config params (read from on-chain) and show a _**current**_ standard compliance status across all chains

superchain/configs/mainnet/arena-z.toml
mslipper: I'm OK with that.

.circleci/artifact_upload_config.yml
```diff
@@ -1,111 +0,0 @@
...
-            FOLDER="chains"
-
-            gsutil cp ./generate-genesis/output-${COMMIT_HASH}/${CHAIN}.json \
-              gs://${GCS_BUCKET}/${FOLDER}/${CHAIN}/genesis.json
-            gsutil cp ./generate-rollup-config/output-${COMMIT_HASH}/${CHAIN}.json \
-              gs://${GCS_BUCKET}/${FOLDER}/${CHAIN}/rollup.json
-
-            # Upload with commit hash for historical records
-            gsutil cp ./generate-genesis/output-${COMMIT_HASH}/${CHAIN}.json \
-              gs://${GCS_BUCKET}/${FOLDER}/${CHAIN}/${COMMIT_HASH}/genesis.json
```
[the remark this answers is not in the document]
mslipper: I have no idea if this was used anywhere.

ops/internal/report/l1.go
mslipper: The deployment event proves that the system was deployed via op-deployer, whereas querying the systemconfig does not. This allows us to skip certain checks altogether as long as we assume that OPCM is correct.

bitwiseguy: But almost all chains (or maybe all?) that are currently in the registry did not use op-deployer. So does that mean we just won't run the reports for those chains? Or can those chains be onboarded onto OPCM and future contract upgrades for those chains will emit a `Deployed` event, at which time the registry could start to generate the report? Or are you only concerned with running the report once when the chain enters the registry, and then never again so it won't matter for chains that are already in the registry?

mslipper: My strawman is that we grandfather those chains in, and focus on the chains that use our governance system.

.circleci/artifact_upload_config.yml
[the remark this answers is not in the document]
geoknee: @ImTei @zhwrd do you know if this is used?

ops/internal/manage/staging_test.go
mslipper: it returns an error now

ops/internal/manage/compress_test.go
```diff
@@ -0,0 +1,75 @@
...
+	require.NoError(t, CompressGenesis(wd, superchain, shortName, testGen))
+	require.FileExists(t, paths.GenesisFile(wd, superchain, shortName))
+	t.Cleanup(func() {
+		require.NoError(t, os.Remove(paths.GenesisFile(wd, superchain, shortName)))
+	})
+
+	readGen, err := DecompressGenesis(wd, superchain, shortName)
+	require.NoError(t, err)
+
+	require.Equal(t, testGen, readGen)
```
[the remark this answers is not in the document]
mslipper: what problems do you anticipate this would help prevent?

superchain/configs/mainnet/arena-z.toml
mslipper: Actually - I think that superchain ops reads the `addresses` struct, so I'd prefer not to change it in this case.

superchain/configs/mainnet/automata.toml
mslipper: see above comment re: the addresses struct.

superchain/configs/mainnet/arena-z.toml
mslipper: Can update this later.

ops/internal/manage/compress_test.go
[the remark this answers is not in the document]
mslipper: going to merge as-is - lmk your thoughts on the above on discord.

mslipper:
Merging this. It has been reviewed live with @tessr, @geoknee, and @bitwiseguy. We're going to work forward from here.

emlautarom1:
The linked document seems to be set to private. Could you change the permissions to make it publicly available?

PR #819, opened 2025-01-02
More verbose errors when reading chain addresses

yashvardhan-kukreja:
This PR add more verbosity to certain errors which are triggered intermittently as a consequence of flaky CI failures.

The added verbosity would help in more efficient troubleshooting for that flakiness.


sigma (approved):
should we consider a tiny helper function to generate these errors?
It feels like enforced consistency might just be desirable for these

yashvardhan-kukreja:
Done @sigma

PR #822, opened 2025-01-06
fix(CI): install-just to use cargo as its installer instead of apt backed by Prebuilt-MPR

yashvardhan-kukreja:
makedeb.org is showing 502 gateway errors on its subdomains including `proget.makedeb.org` which backs the installation of "just" currently (a pre-requisite in every "Required" CI check).

This PR shifts that installation to be backed by Cargo instead.

**Tests**

The correctness of this PR can be seen in all the following CI checks being able to install just in their runs now unlike before.

### DIFF
--- a/.circleci/main_config.yml
+++ b/.circleci/main_config.yml
@@ -32,7 +32,8 @@ commands:
           name: "Install just"
           command: |
             curl https://sh.rustup.rs -sSf | sh -s -- -y
-            export PATH="$PATH:$HOME/.cargo/bin"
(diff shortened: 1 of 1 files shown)


geoknee:
On board with this in principle, but I'm not sure it is working yet (CI checks failing).

yashvardhan-kukreja:
> On board with this in principle, but I'm not sure it is working yet (CI checks failing).


Whoops, ack!

geoknee (approved):

PR #823, opened 2025-01-07
feat: rename chain `minato` to `soneium-minato`

ashutoshvarma:
This PR changes the Soneium Minato Testnet identifier from `sepolia/minato` to `sepolia/soneium-minato`. The previous name when refer from node binaries using network flag was `minato-sepolia` which was confusing to many node operators, it will be changed to now more self describing name `soneium-minato-sepolia`.

**Question**
Is there a way to provide backward compatibility for operators running with old name via network flag in binaries that reference scr?


bitwiseguy:
@ashutoshvarma a few things to address:

1. The chain name now stored at chainid.network is "Soneium Testnet Minato" and your chain name stored [here](https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/configs/sepolia/minato.toml#L1) in superchain-registry needs to match that string exactly for the ci test suite to pass.

2. If you want to change the expected value for `OP_NODE_NETWORK` and `GETH_OP_NETWORK` env vars from `minato-sepolia` to `soneium-minato-sepolia`, then you will have to change the filename from `superchain-registry/superchain/configs/sepolia/minato.toml` --> `superchain-registry/superchain/configs/sepolia/soneium-minato.toml`. That change would not be picked up by op-node or op-geth until they update their imported version of `superchain-registry/superchain` module to include the change.

### Additional context

* relevant ChainConfig.Chain struct field and comment: [link](https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/superchain.go#L104-L106)
* how op-node interprests OP_NODE_NETWORK env var value: [link](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/chaincfg/chains.go#L61)
* how op-geth interprets GETH_OP_NODE env var value: [link](https://github.com/ethereum-optimism/op-geth/blob/optimism/params/superchain.go#L22-L37)

ashutoshvarma:
Thanks for the explaining @bitwiseguy! I've updated the chain name to reflect chainid.network

Our motivation is to mark the minato as testnet explicitly to avoid any confusion with chain operators

bitwiseguy:
@ashutoshvarma yes I can open some prs to update the optimism monorepo and op-geth import of superchain-registry after this is merged. Please keep in mind that anyone running op-node and op-geth for minato will have to manually update their configs/env-vars to reference this new name since this is a breaking change so you should communicate that to all relevant node operators

ashutoshvarma:
Thanks @bitwiseguy !
We will communicate with relevant node operators

bitwiseguy (approved):

PR #825, opened 2025-01-08
minato-sepolia: update chain name to match chainid.network

bitwiseguy:
Validation checks were failing for this chain. Testing this change locally via the following command and got it to pass. Soneium must have changed their chain name at chainid.network.
```
just validate 1946
```

### DIFF
--- a/chainList.json
+++ b/chainList.json
@@ -560,7 +560,7 @@
     }
   },
   {
-    "name": "Minato",
+    "name": "Soneium Testnet Minato",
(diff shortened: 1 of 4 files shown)


Wazabie (commented):
I believe we might be missing the short name here? "shortName": "soneium-minato"

bitwiseguy:
Closing in favor of https://github.com/ethereum-optimism/superchain-registry/pull/823

PR #842, opened 2025-01-18
SR2: fast follows

mslipper:
Fast follows from the big merge:

- Refactored inputs to be decoupled from op deployer
- Added a tool to take an op-deployer state file and generate a chain config
- Added docs
- Fixed test flakes in the diff function
- Fixed a version incompatibility with development OPCM
- Various smaller refactors
- Updated OP Deployer version


docs/ops.md
```diff
@@ -31,63 +31,47 @@ You will be raising a Pull Request from your fork to the upstream repo.
...
 
-The team will then review and merge your chain when it's ready.
+### 4. Update generated config
 
-### 6. Understand output
+Update `<shortname>.toml` with your chain's information. Most of the data is automatically populated from your deployer
+state file, however you will need to populate the following yourself:
 
-The tool will write the following data:
+- `name`: Human-readable name for your chain.
```
bitwiseguy, line 49: Do we still validate that this name matches the one stored at "https://chainid.network/chains_mini.json"? If so, it would be good to mention that here

justfile
```diff
@@ -49,3 +48,6 @@ print-staging-report: (_run_ops_bin 'print_staging_report')
 check-genesis-integrity: (_run_ops_bin 'check_genesis_integrity')
 
 codegen: (_run_ops_bin 'codegen')
+
+inflate-config SHORTNAME FILENAME:
```
bitwiseguy, line 52: `inflate-config` is a confusing term to me. Could we use `create-config` instead?

ops/cmd/print_staging_report/main.go
```diff
@@ -20,10 +21,15 @@ import (
...
+	SepoliaRPCURLFlag = &cli.StringFlag{
+		Name:    "sepolia-rpc-url",
+		Usage:   "The URL of the Sepolia RPC endpoint.",
+		EnvVars: []string{"SEPOLIA_RPC_URL"},
+	}
+	MainnetRPCURLFlag = &cli.StringFlag{
+		Name:     "mainnet-rpc-url",
+		Usage:    "The URL of the mainnet RPC endpoint.",
+		EnvVars:  []string{"MAINNET_RPC_URL"},
 		Required: true,
```
bitwiseguy, line 33: Why make these separate instead of using the `L1RPCURLFlag` as before? Also curious why `MainnetRPCURLFlag` is required but `SepoliaRPCURLFlag` is not. Are you expecting that a testnet will never be added to the registry by itself? Instead, it must always be accompanied by its mainnet when entering the registry?

.circleci/config.yml
```diff
@@ -85,6 +82,18 @@ jobs:
...
+    circleci_ip_ranges: true
+    docker:
+      - image: cimg/go:1.23.1
+    steps:
+      - checkout
+      - run:
+          name: Run staging report
+          command: |
+            cd ops
+            go run ./cmd/print_staging_report/main.go
```
bitwiseguy, line 95: This command has required args for `GithubTokenFlag`, `GitSHAFlag` and `GithubRepoFlag` but I don't see them passed here. Won't that cause this line to fail?

ops/go.work
```diff
@@ -4,4 +4,3 @@ use (
 	.
 	../validation
 )
-
```
bitwiseguy, line 7: What is the purpose of this go.work file? Is it necessary to keep it at this location instead of at the repo root?

ops/internal/config/chain.go
```diff
@@ -6,45 +6,47 @@ import (
...
+	BlockTime            uint64              `toml:"block_time"`
+	SeqWindowSize        uint64              `toml:"seq_window_size"`
+	MaxSequencerDrift    uint64              `toml:"max_sequencer_drift"`
+	GasPayingToken       *ChecksummedAddress `toml:"gas_paying_token,omitempty"`
+	Hardforks            Hardforks           `toml:"hardforks"`
+	Optimism             Optimism            `toml:"optimism"`
+	AltDA                *AltDA              `toml:"alt_da"`
+	Genesis              Genesis             `toml:"genesis"`
+	Roles                Roles               `toml:"roles"`
+	Addresses            Addresses           `toml:"addresses"`
```
bitwiseguy, line 46: Would it be better to rename from `addresses` --> `contracts`? Also curious if `BatchInboxAddr` could be moved inside of this struct, or within `roles` instead of staying as a top level field

ops/internal/config/chain.go
```diff
@@ -6,45 +6,47 @@ import (
...
+	SuperchainTime       *uint64             `toml:"superchain_time"`
+	DataAvailabilityType string              `toml:"data_availability_type"`
+	ChainID              uint64              `toml:"chain_id"`
+	BatchInboxAddr       *ChecksummedAddress `toml:"batch_inbox_addr"`
+	BlockTime            uint64              `toml:"block_time"`
+	SeqWindowSize        uint64              `toml:"seq_window_size"`
+	MaxSequencerDrift    uint64              `toml:"max_sequencer_drift"`
+	GasPayingToken       *ChecksummedAddress `toml:"gas_paying_token,omitempty"`
+	Hardforks            Hardforks           `toml:"hardforks"`
+	Optimism             Optimism            `toml:"optimism"`
```
bitwiseguy, line 42: Is there a better name than `optimism` for this field? The current name doesn't provide much information about what it is

ops/internal/manage/staging.go
```diff
@@ -144,6 +114,7 @@ func InflateChainConfig(sc *StagedChain) (*config.Chain, error) {
 
 	if dc.UseAltDA {
 		cfg.Addresses.DAChallengeAddress = config.NewChecksummedAddress(dc.DAChallengeProxy)
+		cfg.DataAvailabilityType = "alt-da"
```
bitwiseguy, line 117: There are two `if dc.UseAltDA` blocks (here and line 61-68). Could they be combined?

docs/ops.md
mslipper: It does not do this validation - I wasn't aware that was a requirement. I can bring it back if needed.

ops/internal/report/l2.go
```diff
@@ -1,85 +1,41 @@
...
+	"math/big"
 	"os"
 
 	"github.com/ethereum-optimism/optimism/op-chain-ops/foundry"
 	"github.com/ethereum-optimism/optimism/op-chain-ops/genesis"
-	"github.com/ethereum-optimism/optimism/op-deployer/pkg/deployer/artifacts"
 	"github.com/ethereum-optimism/optimism/op-deployer/pkg/deployer/broadcaster"
-	"github.com/ethereum-optimism/optimism/op-deployer/pkg/deployer/inspect"
-	"github.com/ethereum-optimism/optimism/op-deployer/pkg/deployer/opcm"
+	opcmv170 "github.com/ethereum-optimism/optimism/op-deployer/pkg/deployer/opcm/v170"
```
bitwiseguy, line 12: I don't think I've seen versioning like this before. What makes this scheme preferable over more standard go module versioning where only the major version is included in the module name (e.g., `opcm/v1`)? Maybe this is necessary due to upstream contracts versioning, and the contracts are the one's that should be bumping major version more often?

ops/cmd/print_staging_report/main.go
mslipper: Both flags should be required. The flags are separated due to CircleCI's env var design. It's easier to load both into CircleCI as env vars and have the tool choose which one to use than it is to have the CircleCI bash script in the job pick the right URL.

.circleci/config.yml
mslipper: They are injected via CCI env vars since they are secrets.

ops/go.work
mslipper: It's needed to allow local development of the validation package at the same time as the ops scripts. Otherwise we'd have to tag the validation package each time it changes.

ops/internal/config/chain.go
mslipper: Changing these would be a much bigger refactor. The monorepo contracts and superchain-ops both assume an addresses stanza.

BatchInboxAddr is a special case - it's more of a protocol constant than it is a live address with a contract deployed on it, so it doesn't really fit Addresses or Roles. Unless you're strongly opinionated about it I'd rather keep it where it is.

ops/internal/config/chain.go
mslipper: That's what the field is in the existing TOMLs, and the genesis files. It corresponds with the Optimism stanza in the chain params.

ops/internal/report/l2.go
mslipper: This maps to the contract version, not the Go version. OPD supports deploying multiple, backwards-incompatible versions of the contracts in one package so we need versioned sub-packages that contain the logic that describes how do deploy each version.

Changing the Go version to a new major version is a huge can of worms that I'd rather avoid while we're in fast development mode on OP Deployer.

ops/cmd/print_staging_report/main.go
mslipper: fixed the required flag

ops/internal/manage/staging.go
mslipper: fixed

docs/ops.md
bitwiseguy: Not sure that its a requirement from the standard charter. But its a nice sanity check to ensure the superchain-registry data matches chainid's data. Previously we required the chain to be added to chainid.network first, then we verified the data matched what was being added to the superchain-registry.

I see that you are still grabbing the data from chainid.network as part of the `FetchGlobalChainIDs` function, but it seems the test will fail if the chain already exists at chainid.network? Is that intentional?

bitwiseguy (approved):

docs/ops.md
mslipper: > Is that intentional?

Yes, but I didn't realize that there was a requirement that the chain already exist in the global chain IDs. I can update the code to fix that.

mslipper: fixed, there's a new CI check for it

Issue #857, opened 2025-01-20
display fault proof status for chains


tessr:
When we planned the Superchain Registry initially, we didn't expect chains to spend an extended period of time on permissioned FPs: We thought everyone would get quickly to permissionless FPs (and Stage 1). But as more chains spend a longer time with permissioned FPs, we need a way to call out the distinction (and the Superchain Index wants this info as well). 

We should detect and display whether chains are on:

* Permissionless FPs
* Permissioned FPs 
* No FPs (still using the L2OO) 

And display this in the chainList.json/toml files.

bitwiseguy:
How to Tell If a Chain Runs Fault proofs: [link](https://www.notion.so/oplabs/Fault-Proof-Answers-12cf153ee162802b84b7c9cb1054e818?pvs=4#19ff153ee16280e89762eddc4218a039)

alfonso-op:
Reassigned to @yashvardhan-kukreja based on our discussion earlier this week.

alfonso-op:
@yashvardhan-kukreja just synced with Sam and he clarified he'll take this. Sorry for the noise!

yashvardhan-kukreja:
No problemo, cheers! Thanks for the clarification :)

PR #860, opened 2025-01-21
Support multiple superchain levels

mslipper:
We previously had a `standard_chain_candidate` option that denoted chains who were standard except for key handover. This PR brings that back by updating the definition of the `superchain_level` field. The new definitions are:

- 0: Non-Standard
- 1: Standard Candidate
- 2: Standard

All config TOMLs have been updated to use this new definition.


docs/glossary.md
```diff
@@ -15,6 +15,8 @@ To help with clarity and a common understanding, here are some helpful terms and
 
 Chains in the superchain-registry are assigned a `superchain_level` (shown in individual config files as well as the `chainList.toml/json` summaries), depending on the set of validation checks that they pass.
 
-**Frontier** chains have `superchain_level = 0`. These are members of the Superchain ecosystem.
+**Non-Standard** chains have `superchain_level = 0`. These are members of the Superchain ecosystem.
```
bitwiseguy, line 18: What does "member of the Superchain ecosystem" mean? Would it be accurate and less ambiguous to say: "These are chains that run on the OP stack, but do not conform to the standard blockchain charter"

bitwiseguy:
Previously we made the change to align `superchain_level` with the l2beat decentralization stage. I'm supportive of the change in this pr, just flagging to @tessr for visibility.

Also wondering if we can rename from `superchain_level` --> `standard_level` for a more accurate description of that field

sebastianst (approved):
Changes LGTM. I don't have strong opinions on the semantics of the superchain level (or "standard level" as Sam suggests). Sounds more like a product question to me.

tessr:
This is fine with me. 

For the record, this is what I wrote in Discord: 

> We previously tried to have this adhere closer to "stages" ranking; in other words you had to be superchain_level = 1 in order to be stage 1. This has gotten trickier now that we have "green" chains in the superchain index that are not using permissionless FPs, so these concepts need to be revisited. 

Slipper also pointed out that it's weird to have "what's essentially an enum whose behavior changes if a secondary boolean changes", which is extremely fair. 

So I'm onboard with this. 

However, this means that the SR no longer effectively displays a chain's decentralization stage. We should prioritize adding information on a chain's FP status (#857) so that readers can see what decentralization stage each chain is at. (Stage 1 == OP governed + permissionless FPs)

PR #869, opened 2025-01-24
CHAINS.md: add column explanations

tessr:


mslipper:
@tessr just needs the test to be modified to match the new chains output, then this is good to go

scharissis (approved):

PR #870, opened 2025-01-25
Update standard prestates format

mslipper:


validation/standard/standard-prestates.toml
```diff
@@ -1,3 +1,114 @@
-["op-contracts/v1.6.0"]
-program_version = "1.3.1"
-absolute_prestate = "0x038512e02c4c3f7bdaec27d00edf55b7155e0905301e1a88083e4e0a6764d54c"
+latest_stable = "1.4.0"
+latest_rc = "1.4.1-rc.2"
```
bitwiseguy, line 2: Is it possible in the future we could have multiple latest_stable versions to call out here, one for each type? Or will there only ever be one type (cannon64 in this case) that can be considered latest_stable at a time. Wondering if we want a data structure here that would support latest stable for cannon64 + latest stable for kona + any other valid type?

validation/standard/standard-prestates.toml
bitwiseguy, line 1: All of this data is duplicated in the monorepo [here](https://github.com/ethereum-optimism/optimism/blob/develop/op-program/prestates/releases.json). Do we plan to keep the data in both locations or only here and get rid of the monorepo ones?

bitwiseguy (approved):
Approved with a few open questions

validation/standard/standard-prestates.toml
bitwiseguy: Also curious if the proofs team will be expected to update this file as new prestates versions are created. If so, want to make sure they are aware of that expectation

validation/standard/standard-prestates.toml
mslipper: i think it would be multiple types per latest RC. we have the ability to modify this in the future if we need to.

validation/standard/standard-prestates.toml
mslipper: i'll talk to adrian about it. i'll build tooling to copy it over if we need to.

PR #890, opened 2025-02-07
feat: Add mise support

blmalone:
This adds mise to the superchain-registry.

~EDIT: As part of this PR I added `mise` so that I could run `just` commands from the `config.yml` file.~

~As discussed in [discord](https://discord.com/channels/1244729134312198194/1335009077117456394/1336084466833293443), we need this to make sure that `addresses.json` is always up to date. The build should fail if the codegen job creates changes that aren't yet commited to the main branch.~

~I've put this on an hourly trigger that already existed in the superchain-registry configuration settings.~

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,70 @@
 version: 2.1
 
-orbs:
-  codecov: codecov/codecov@5.0.3
+parameters:
+  default_docker_image:
+    type: string
+    default: cimg/base:2024.01
+
+commands:
+  checkout-with-mise:
+    steps:
+      - checkout
+      - run:
+          name: Initialize mise environment
+          command: |
+            # This is used to create a per-user cache key to preserve permissions across different
+            # executor types.
+            user=$(whoami)
+            echo "$user" > .executor-user
+            echo "Set executor user to $user."
+
+            if [[ "$user" == "root" ]]; then
+              # Self-hosted runners will persist this cache between runs. Cleaning it up means that we
+              # preserve the semantics of the cache regardless of executor type. It's also much faster
+              # to delete the cache and recreate it than it is to overwrite it in place.
+              rm -rf /data/mise-data
+              echo "Cleaned up cache data."
+
+              mkdir -p /data/mise-data
+              echo "Created Mise data dir."
+              mkdir -p ~/.cache
+              echo "Created Mise cache dir."
+            else
+              sudo rm -rf /data/mise-data
+              echo "Cleaned up cache data."
+              sudo mkdir -p /data/mise-data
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -101,7 +167,24 @@ jobs:
...
+      - image: <<pipeline.parameters.default_docker_image>>
+    steps:
+      - checkout-with-mise
+      - run:
+          name: Checks if running codegen causes changes
+          command: |
+            just check-for-codegen-changes
+
 workflows:
+  scheduled-check-codegen-changes:
```
mds1, line 181: Is this setup to notify somewhere on slack when it fails?

justfile
```diff
@@ -53,3 +53,23 @@ create-config SHORTNAME FILENAME:
...
 check-chainlist: (_run_ops_bin 'check_chainlist')
+
+check-for-codegen-changes:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    if ! git diff-index --quiet HEAD --; then
+        echo -e "\033[31mError: Working directory is not clean. Please commit or stash your changes before checking for codegen changes.\033[0m"
+        exit 1
+    fi
+    just codegen
```
mds1, line 64: This just generates the addresses JSON based on what’s in the TOML, and doesn’t confirm that either TOML or JSON files are accurate and match the chain state? The codegen check runs on PRs already so I don’t think running this hourly is what we want, unless codegen does check chain state to verify

blmalone: I see, I understood it as [make sure `addresses.json` was up to date](https://github.com/ethereum-optimism/superchain-ops/issues/549). You're right I see this is already being ran in CI.

.circleci/config.yml
blmalone: No

justfile
mds1: Ah, I think the definition of “up to date” was unclear. CI currently enforces that the JSON is up to date based on the TOML, but the TOML is only updated manually. We should be proactive and check the chain for mismatches between chain data and registry data

.circleci/config.yml
```diff
@@ -3,15 +3,81 @@ version: 2.1
...
+  default_docker_image:
+    type: string
+    default: cimg/base:2024.01
+
+commands:
+  checkout-with-mise:
+    steps:
+      - checkout
+      - run:
+          name: Initialize mise environment
```
mslipper, line 16: [nit] we should make this into an orb later, but this is fine for now.

mslipper (approved):

.circleci/config.yml
blmalone: https://github.com/ethereum-optimism/superchain-registry/issues/892

PR #897, opened 2025-02-13
remove _proxy_ from opcm key name

maurelian:
The OPCM is no longer proxied. This just changes the name of the the key from `op_contracts_manager_proxy_addr` to `op_contracts_manager_addr`.

No idea what this might break, hoping to find out here.

### DIFF
--- a/ops/internal/manage/testdata/superchain/configs/sepolia/superchain.toml
+++ b/ops/internal/manage/testdata/superchain/configs/sepolia/superchain.toml
@@ -1,7 +1,7 @@
 name = "Sepolia"
 protocol_versions_addr = "0x79ADD5713B383DAa0a138d3C4780C7A1804a8090"
 superchain_config_addr = "0xC2Be75506d5724086DEB7245bd260Cc9753911Be"
-op_contracts_manager_proxy_addr = "0xF564eEA7960EA244bfEbCBbB17858748606147bf"
+op_contracts_manager_addr = "0xF564eEA7960EA244bfEbCBbB17858748606147bf"
 
 [hardforks]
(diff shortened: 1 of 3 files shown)


ops/internal/manage/testdata/superchain/configs/sepolia/superchain.toml
```diff
@@ -1,7 +1,7 @@
 name = "Sepolia"
 protocol_versions_addr = "0x79ADD5713B383DAa0a138d3C4780C7A1804a8090"
 superchain_config_addr = "0xC2Be75506d5724086DEB7245bd260Cc9753911Be"
-op_contracts_manager_proxy_addr = "0xF564eEA7960EA244bfEbCBbB17858748606147bf"
+op_contracts_manager_addr = "0xF564eEA7960EA244bfEbCBbB17858748606147bf"
```
mds1, line 4: Hm, this format is outdated too, and it should be something like this now:
```suggestion
[opcm]
"v1.8.0" = "0x123..."
"v2.0.0" = "0x456..."
```

mds1: actually this will need to be keyed by network name also
```toml
[opcm]
# Key refers to the L1 contracts release
"v1.8.0" = {mainnet = "0x123...", sepolia = "0xabc..." }
"v2.0.0" = {mainnet = "0x456...", sepolia = "0xdef..." }
```

mslipper:
These values should go into the standard-versions TOMLs. PR is here: https://github.com/ethereum-optimism/superchain-registry/pull/919

PR #835, opened 2025-01-17
[chore]: add camp network v2 testnet + run tests

anirudhmakhana:
# Adding a New Chain
This PR adds Camp Network Testnet V2 to the registry.

## Checklist
- [x] I have declared the chain at the appropriate [Superchain Level](../docs/glossary.md#superchain-level-and-rollup-stage).
- [x] I have run `just validate <chain-id>` locally to ensure all local validation checks pass. 
- [x] I have run `just codegen` to ensure that the chainlist and other generated files are up-to-date and include my chain.

PR #899, opened 2025-02-15
new chain: camp network testnet

anirudhmakhana:
# Adding a New Chain

This PR adds Camp network testnet to the registry.

## Checklist

- [x ] I have declared the chain at the appropriate [Superchain Level](../docs/glossary.md#superchain-level-and-rollup-stage).
- [ x] I have run `just codegen` to ensure that all generated files are created.
- [x ] I have run `just codegen` to ensure that the chainlist and other generated files are up-to-date and include my chain.
- [ x] I have checked `Allow edits from maintainers` on this PR.


mslipper:
After fixing some minor issues with your chain config and reviewing the contract deployment transactions, it appears like you used the legacy deployment script rather than OP Deployer. If this is the case, you will have to use an alternative means to verify your chain before adding it to the registry. If you did use OP Deployer, can you please correct the `deployment_tx_hash` value in your config to point to the correct transaction?

Wazabie:
> After fixing some minor issues with your chain config and reviewing the contract deployment transactions, it appears like you used the legacy deployment script rather than OP Deployer. If this is the case, you will have to use an alternative means to verify your chain before adding it to the registry. If you did use OP Deployer, can you please correct the `deployment_tx_hash` value in your config to point to the correct transaction?

I don't think this was deployed with op-deployer actually, there is this other [PR for Camp](https://github.com/ethereum-optimism/superchain-registry/pull/835) . Can we use this one to validate or we need @anirudhmakhana to open another?

Wazabie (approved):

PR #900, opened 2025-02-16
Fix op-sepolia dispute game contract addresses

pauldowman:
These were missed in the rollback from MIPS64.


ajsutton (approved):
LGTM. Just need to run the codeine.

PR #877, opened 2025-01-29
Add fraxtal

0xalex88:
# Adding a New Chain

This PR adds Fraxtal to the registry.

I've asked in our shared telegram chat about adding Fraxtal to the superchain repo and was suggested to add the manually created files directly (since our chain was already launched).

Please let me know if there's anything missing.

## Checklist

- [X] I have declared the chain at the appropriate [Superchain Level](../docs/glossary.md#superchain-level-and-rollup-stage).
- [X] I have run `just codegen` to ensure that all generated files are created.
- [X] I have run `just codegen` to ensure that the chainlist and other generated files are up-to-date and include my chain.
- [X] I have checked `Allow edits from maintainers` on this PR.

PR #940, opened 2025-03-14
Add fraxtal

0xalex88:
Follow up on https://github.com/ethereum-optimism/superchain-registry/pull/877


0xalex88:
@bitwiseguy so, trying to generate a geth datadir gives the expected hash but the test code somehow generates a different one:
```
❯ docker run --rm -ti -v `pwd`/superchain-registry/superchain/extra/genesis/mainnet/fraxtal.json:/genesis.json:ro  us-docker.pkg.dev/oplabs-tools-artifacts/images/op-geth:v1.101503.0 init /genesis.json
INFO [03-14|16:01:49.948] Maximum peer count                       ETH=50 total=50
INFO [03-14|16:01:49.949] Smartcard socket not found, disabling    err="stat /run/pcscd/pcscd.comm: no such file or directory"
INFO [03-14|16:01:49.952] Set global gas cap                       cap=50,000,000
INFO [03-14|16:01:49.953] Initializing the KZG library             backend=gokzg
INFO [03-14|16:01:49.972] Defaulting to pebble as the backing database
INFO [03-14|16:01:49.972] Allocated cache and file handles         database=/root/.ethereum/geth/chaindata cache=16.00MiB handles=16
INFO [03-14|16:01:49.989] Opened ancient database                  database=/root/.ethereum/geth/chaindata/ancient/chain readonly=false
INFO [03-14|16:01:49.990] State schema set to default              scheme=path
ERROR[03-14|16:01:49.991] Head block is not reachable
INFO [03-14|16:01:50.000] Opened ancient database                  database=/root/.ethereum/geth/chaindata/ancient/state readonly=false
INFO [03-14|16:01:50.000] Initialized path database                cache=16.00MiB buffer=64.00MiB history=90000
INFO [03-14|16:01:50.000] Writing custom genesis block
INFO [03-14|16:01:51.216] Successfully wrote genesis state         database=chaindata hash=521982..d6bfaa
```

do you have any idea why that would be?

bitwiseguy:
@0xalex88 - when you ran `just add-chain` locally did all checks pass locally ([ref](https://github.com/ethereum-optimism/superchain-registry/blob/pre-sr2/docs/ops.md#adding-a-chain))? If not, can you share the failure logs?

Also, can you share your full genesis.json file so that we can see if the error is happening at the compression/decompression step or if there is an actual diff? I do see that `cast block -r "https://rpc.frax.com/" 0` returns a block hash of `0x521982bd54239dc71269eefb58601762cc15cfb2978e0becb46af7962ed6bfaa`, which matches the chain config .toml file you are trying to commit

0xalex88:
> @0xalex88 - when you ran `just add-chain` locally did all checks pass locally ([ref](https://github.com/ethereum-optimism/superchain-registry/blob/pre-sr2/docs/ops.md#adding-a-chain))? If not, can you share the failure logs?
> 
> Also, can you share your full genesis.json file so that we can see if the error is happening at the compression/decompression step or if there is an actual diff? I do see that `cast block -r "https://rpc.frax.com/" 0` returns a block hash of `0x521982bd54239dc71269eefb58601762cc15cfb2978e0becb46af7962ed6bfaa`, which matches the chain config .toml file you are trying to commit

Sorry for the delay, add-chain worked, should I force push on thie branch or create a new PR? I only had to change the `addresses.go` file because it would try to extract the address from the `.chainId` file but the script finished properly and the genesis hash .

0xalex88:
@bitwiseguy force pushed the verrsion created with add-chain

Wazabie:
closing this as the chain is being added with #877

0xalex88:
Thank you so much @Wazabie ! 🚀

PR #950, opened 2025-03-26
Fix weths

maurelian:
The weth addresses are outdated due to previous upgrades. 

The new addresses can be verified by:

1. getting this [readSystemConfig](https://gist.github.com/maurelian/a9b7b2ccb0cca1806bdf05442038a9b1) script
2. for each chain, call `readSystemConfig $SYSTEM_CONFIG_ADDRESS`. It will crawl the chain and print whatever weth addresses it finds.

I've also renamed the toml key for the weth entries. We will see if something breaks in CI, but I doubt it.

Edit: the applyHardforks task overwrites everything. More work is needed here. Moving to draft.


bitwiseguy:
@maurelian - I am wrapping up [this pr](https://github.com/ethereum-optimism/superchain-registry/pull/951) which uses op-fetcher to crawl the onchain SystemConfig and grab all chain addresses (very similar to the script you mention in this pr). I think we should hold off on this pr in favor of that one.

I am planning to run a daily ci job which will fetch all contract+roles addresses for every chain in the superchain, and will automatically open a pr to update the codegen files (addresses.json, chainList.json, chainList.toml, CHAINS.md) if any of the data in the registry is stale. So we shouldn't have to make any more manual updates like in this pr

PR #951, opened 2025-03-27
ops: integrate op-fetcher

bitwiseguy:
## Overview
Integrates [op-fetcher](https://github.com/ethereum-optimism/optimism/tree/develop/op-fetcher) from the monorepo into this repo in order to pull chain config info on-demand from onchain instead of relying on static, hard-coded addresses in the chain config .toml files. This will help prevent data in the chain config .toml and codegen files from going stale and requiring manual updates. Instead, a daily ci job will run and open a pr to update the following codegen files if it notices any diffs between what is onchain and what is stored offchain in the registry:
* `addresses.json`
* `chainList.json`
* `chainList.toml`
* `CHAINS.md`

## Details
The existing `ops/cmd/codegen` command has been modified to use `op-fetcher` instead of pulling all of its information from hardcoded information in the chain config .toml files.

In addition to contract and roles addresses, `op-fetcher` also outputs a `faultProofsStatus` struct. This information is now stored in the `ChainListEntry` struct that comprises `chainList.json` and `chainList.toml`.

## Usage
1. Update all chains: this is how the ci job will invoke the command. Any superchain (mainnet, sepolia, sepolia-dev-0) will be updated if an l1-rpc-url with the appropriate l1 chainId is provided.
```
go run ./cmd/codegen \
  --l1-rpc-urls="$SEPOLIA_RPC_URL,$MAINNET_RPC_URL"
```
2. Update all chains for a single superchain:
```
go run ./cmd/codegen \
  --l1-rpc-urls="$SEPOLIA_RPC_URL,$MAINNET_RPC_URL" \
  --superchains="mainnet"
```
3. Update specific chains by passing a `--chain-ids` flag: this will be used when an individual chain is added or modified. This command will fail if none of the given `l1-rpc-urls` have the appropriate l1 chainId to query the l2(s) contracts.
```
go run ./cmd/codegen \
  --l1-rpc-urls="$SEPOLIA_RPC_URL,$MAINNET_RPC_URL" \
  --chain-ids=<l2-chain-id>
```

## Related Issues
* Closes https://github.com/ethereum-optimism/optimism/issues/14757
* Closes https://github.com/ethereum-optimism/superchain-registry/issues/857

## Follow-up Work
Remove all addresses from the individual chain config .toml files except the following ones. The full collection of addresses should be read from the more up-to-date `addresses.json` file, or read on-demand using op-fetcher.
* SystemConfigProxy
* L1StandardBridgeProxy
* OptimismPortalProxy ([op-node reads this from rollup configs](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/superchain.go#L95))
* ProxyAdminOwner (role)


.circleci/config.yml
```diff
@@ -124,6 +124,131 @@ jobs:
...
+      - utils/checkout-with-mise
+      - run:
+          name: Execute codegen
+          working_directory: ops
+          command: |
+            go run ./cmd/codegen/main.go --l1-rpc-urls="<< pipeline.parameters.sepolia_rpc_url >>,<< pipeline.parameters.mainnet_rpc_url >>"
+      - run:
+          name: Create PR if changes detected
+          working_directory: .
+          command: |
```
bitwiseguy, line 209: I have not tested this yet. Might be best to extract this daily-ci job from this pr and put into a follow-up pr

bitwiseguy: Removed in https://github.com/ethereum-optimism/superchain-registry/pull/951/commits/dbf4ebda3c1c7bd4bb963baa5f05670285c80a04 to make this pr cleaner/easier to review

ops/internal/config/address.go
```diff
@@ -43,6 +45,30 @@ func (a ChecksummedAddress) String() string {
...
+	}
+
+	// Convert to checksummed address
+	addr := common.HexToAddress(dataStr)
+	*a = ChecksummedAddress(addr)
+	return nil
+}
+
+func (a *ChecksummedAddress) MarshalJSON() ([]byte, error) {
+	if bytes.Equal(common.Address(*a).Bytes(), common.Address{}.Bytes()) {
```
mslipper, line 69: Why do you need a bytes.Equal here rather than just doing `common.Address(*a) == common.Address{}`?

ops/internal/config/address.go
```diff
@@ -43,6 +45,30 @@ func (a ChecksummedAddress) String() string {
 	return common.Address(a).Hex()
 }
 
-func (a ChecksummedAddress) MarshalJSON() ([]byte, error) {
-	return []byte(fmt.Sprintf(`"%s"`, common.Address(a).Hex())), nil
+func (a *ChecksummedAddress) UnmarshalJSON(data []byte) error {
```
mslipper, line 48: Why do we need to change the serialization?

ops/internal/config/chain.go
```diff
@@ -182,34 +183,93 @@ type AddressesWithRoles struct {
...
+}
+
 func (a AddressesWithRoles) MarshalJSON() ([]byte, error) {
-	// Create a map to hold all fields
-	allFields := make(map[string]*ChecksummedAddress)
+	// Create a map to hold non-zero address fields
+	allFields := make(map[string]string)
+
+	// Declare processStruct variable first to allow recursion
+	var processStruct func(interface{})
```
mslipper, line 231: What is this doing?

ops/internal/config/superchain.go
```diff
@@ -10,6 +18,12 @@ const (
 	SepoliaDev0Superchain Superchain = "sepolia-dev-0"
 )
 
+var SuperchainChainIds = map[Superchain]uint64{
```
mslipper, line 21: Aren't these chain IDs already included in the superchain config? Why are we making a new hardcoded map here?

ops/internal/config/superchain.go
```diff
@@ -27,6 +41,64 @@ func MustParseSuperchain(in string) Superchain {
...
+
+// getL1ChainId connects to an Ethereum RPC endpoint and retrieves its chain ID
+func getL1ChainId(rpcURL string) (uint64, error) {
+	client, err := ethclient.Dial(rpcURL)
+	if err != nil {
+		return 0, fmt.Errorf("failed to connect to L1 RPC: %w", err)
+	}
+	defer client.Close()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
mslipper, line 91: Don't declare context like this in private methods, instead have the caller pass them in

ops/internal/manage/codegen.go
```diff
@@ -5,140 +5,249 @@ import (
...
+	if err != nil {
+		return nil, fmt.Errorf("error reading addresses file: %w", err)
+	}
+	if err := json.Unmarshal(addressesData, &addresses); err != nil {
+		return nil, fmt.Errorf("error unmarshaling addresses.json: %w", err)
+	}
+
+	// Load chainList.json data
+	var chainList []config.ChainListEntry
+	chainListData, err := os.ReadFile(filepath.Join(wd, "chainList.json"))
```
mslipper, line 56: This should be one of the FS helpers.

ops/internal/manage/codegen.go
```diff
@@ -5,140 +5,249 @@ import (
...
+	}
+	if err := json.Unmarshal(chainListData, &chainList); err != nil {
+		return nil, fmt.Errorf("error unmarshaling chainList file: %w", err)
+	}
 
-func GenAllCode(rootP string) error {
-	if err := GenAddressesFile(rootP); err != nil {
-		return fmt.Errorf("error generating addresses file: %w", err)
+	// Load disk chain configs
+	configsDir := path.Join(wd, "superchain", "configs")
```
mslipper, line 65: Ditto - use/make an FS helper.

ops/internal/manage/fetch_onchain.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/ethereum-optimism/optimism/op-fetcher/pkg/fetcher/fetch/script"
+	"github.com/ethereum-optimism/superchain-registry/ops/internal/config"
+	"github.com/ethereum-optimism/superchain-registry/ops/internal/paths"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/log"
+	"golang.org/x/sync/errgroup"
+)
+
+// FetchChains fetches chain configurations for specified chain IDs or all chains if none specified
+func FetchChains(lgr log.Logger, wd string, l1RpcUrls []string, chainIdStrs []string) (map[uint64]script.ChainConfig, error) {
```
mslipper, line 19: Why are the chain IDs passed in as strings? They should be a `uint64` (or, for a future refactor, `eth.ChainID`) in public methods. The CLI should be the only thing that deals with string chain IDs.

ops/internal/manage/codegen.go
```diff
@@ -5,140 +5,249 @@ import (
...
-	case ".toml":
-		marshaler = func(entries []config.ChainListEntry) ([]byte, error) {
-			return toml.Marshal(config.ChainListTOML{
-				Chains: entries,
-			})
-		}
-	default:
-		return fmt.Errorf("unsupported file extension: %s", ext)
+	// Ensure the addresses directory exists
+	addressesPath := paths.AddressesFile(s.outputWd)
```
mslipper, line 182: FS helper

ops/internal/manage/collect.go
```diff
@@ -92,3 +93,79 @@ func CollectChainConfigs(p string) ([]DiskChainConfig, error) {
 
 	return out, nil
 }
+
+// FindChainConfig searches all superchains for given chain ID
+func FindChainConfig(wd string, chainId uint64) (*DiskChainConfig, config.Superchain, error) {
```
mslipper, line 98: This method doesn't seem to be called anywhere

ops/internal/manage/collect.go
```diff
@@ -92,3 +93,79 @@ func CollectChainConfigs(p string) ([]DiskChainConfig, error) {
...
+}
+
+type ChainConfigTuple struct {
+	Config     *DiskChainConfig
+	Superchain config.Superchain
+}
+
+// FindChainConfigs searches all superchains for the given chain IDs
+// - returns an error if any chain ID is not found
+func FindChainConfigs(wd string, chainIds []uint64) ([]ChainConfigTuple, error) {
```
mslipper, line 127: I'd rather modify the collect method to return the superchain data then have the caller do the filtering as opposed to passing in chain IDs to filter on here. We should only be walking the disk once.

ops/internal/manage/collect_test.go
```diff
@@ -33,3 +33,27 @@ func TestCollectChainConfigs(t *testing.T) {
 		},
 	}, chains)
 }
+
+func TestFindChainConfig(t *testing.T) {
```
mslipper, line 37: Need tests for `FindChainConfigs`

ops/internal/manage/fetch_onchain.go
```diff
@@ -0,0 +1,128 @@
...
+		return nil, err
+	}
+
+	lgr.Info("completed fetching", "totalChains", len(allChainConfigs))
+	return allChainConfigs, nil
+}
+
+// collectChainsBySuperchain assembles a map of chains grouped by their superchain
+// based on provided chain IDs or all available chains if no IDs are specified
+func collectChainsBySuperchain(wd string, chainIdStrs []string) (map[config.Superchain][]DiskChainConfig, error) {
```
mslipper, line 67: You can probably roll this into `CollectChainConfigs`, see my feedback above

mslipper (changes requested):
A few gaps before this is ready to merge. Specifically:

- Stuff that touches the disk should use an FS helper in `paths`
- We should be scanning the disk once for all chain configs/superchains, then using in-memory data for further filtering operations. This is the architecture for other ops tools, and makes things much simpler

ops/internal/config/address.go
bitwiseguy: Good catch. We don't

ops/internal/manage/collect.go
bitwiseguy: I refactored to always use FindChainConfigs instead of this function. Forgot to remove

ops/internal/config/superchain.go
```diff
@@ -27,6 +41,64 @@ func MustParseSuperchain(in string) Superchain {
 	return sup
 }
 
+// FindValidL1URL finds a valid l1-rpc-url for a given superchain by finding matching l1 chainId
+func FindValidL1URL(lgr log.Logger, urls []string, superchain Superchain) (string, error) {
```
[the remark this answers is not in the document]
bitwiseguy: Was trying to make it as simple as possible for the end user. Right now you just need this:
```
--l1-rpc-urls
```
Instead of:
```
--mainnet-rpc-url OR --sepolia-rpc-url
--superchain
```
Its 2 less decisions the end user needs to make with the current implementation. Also allows us to remove logic in ci and a few other places by always expecting the same input format

ops/internal/config/chain.go
bitwiseguy: Basically what the comment says. The recursion below doesn't work without first declaring the var

ops/internal/config/superchain.go
bitwiseguy: Done in https://github.com/ethereum-optimism/superchain-registry/pull/951/commits/e7b55db0fd5d180a19bf9bba848d3d9e8d8d0f86

ops/internal/manage/collect.go
bitwiseguy: Good call. I removed `FindChainConfigs` in https://github.com/ethereum-optimism/superchain-registry/pull/951/commits/de46b1a8fb7c8e7d1693c08351753fba87b73b6b. The `[]DiskChainConfigs` returned by `CollectChainConfigs` now includes the `Superchain` for each config

ops/internal/config/chain.go
mslipper: Sorry, I meant the entire method

ops/internal/config/address.go
bitwiseguy: We get errors when using Go's standard JSON unmarshaler because it doesn't know how to convert a string into the ChecksummedAddress type. Using this custom unmarshal means if incoming addresses (from state.json, etc) are not checksummed, we can still read them, then when we write them they will be checksummed

bitwiseguy: Its the same logic as the custom UnmarshalTOML that already exists in this file. Since we are now reading from `addresses.json`, instead of only writing to it, it makes sense to have the custom UnmarshalJSON to me.

bitwiseguy: Want to fail if read address is not checksummed

bitwiseguy: Cleaned this up and used a helper method to encapsulate the code used by both UnmarshalToml and UnmarshalJSON

ops/internal/config/superchain.go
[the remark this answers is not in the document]
bitwiseguy: Added optional superchains flag in https://github.com/ethereum-optimism/superchain-registry/pull/951/commits/6d3195bb390979cee842298720f4014c83bf80d7

justfile
```diff
@@ -47,8 +47,6 @@ print-staging-report: (_run_ops_bin 'print_staging_report')
 
 check-genesis-integrity: (_run_ops_bin 'check_genesis_integrity')
 
-codegen: (_run_ops_bin 'codegen')
```
mslipper, line 50: why was this deleted?

mslipper (approved):

justfile
bitwiseguy: Because you now are required to provide `--l1-rpc-urls` flag for `codegen` therefore I think its cleaner to invoke the `codegen` go program directly instead of via `just` recipe

PR #957, opened 2025-03-31
Update base devnet pectra blob schedule fix time

cody-wang-cb:
Update base devnet pectra blob schedule fix time


sebastianst (approved):
confirmed with Base in Slack that this is correct

PR #826, opened 2025-01-10
update standard versions and release to latest governance approved

geoknee:

PR #960, opened 2025-04-02
Add v2.0.0 contracts to standard-versions

maurelian:
This PR adds the [`op-contracts/v2.0.0` release](https://github.com/ethereum-optimism/optimism/releases/tag/op-contracts%2fv2.0.0) to the `superchain-registry`.

The `rc` tag was removed from the OPCM here: https://etherscan.io/tx/0x5f3530e593bbac37c61dc5b7755b6a40c06c20c1a3a1b13fca5b7d00cde65c29


bitwiseguy:
Should we update this [versionsToCheck](https://github.com/ethereum-optimism/superchain-registry/blob/main/ops/internal/validation/versions_test.go#L52-L56) section of the validation tests? I don't think there are any checks against the release you're adding unless we update the linked section.

Seems like we should also update the [standard-bytecodes.toml](https://github.com/ethereum-optimism/superchain-registry/blob/main/validation/standard/standard-bytecodes.toml) and [standard-immutables.toml](https://github.com/ethereum-optimism/superchain-registry/blob/main/validation/standard/standard-immutables.toml) files for this new release. Although I'm not sure where those get validated. I'm also curious why the latest release in those two files is `"op-contracts/v1.8.0"` but in the `standard-release-mainnet.toml` its `"op-contracts/v3.0.0-rc.2"`. Maybe @mslipper can help clarify the intended purpose of those files

maurelian:
Moving to draft as am awaiting feedback on this proposal:

I looked [at another PR](https://github.com/ethereum-optimism/superchain-registry/pull/826/files) that did the same thing as this one, and noticed that it also includes changes to `standard-bytecode.toml` and `standard-immutables.toml`. However my PR is passing CI despite not including changes to those files. Since `v2.0.0` was the first OPCM release, IMO we should no longer need to track this data, as the release is determined by the onchain addresses.

mslipper (approved):

PR #968, opened 2025-04-09
Revert "Add new value for EIP-1559 elasticity for WorldChain Sepolia and Mainnet"

sebastianst:
Reverts ethereum-optimism/superchain-registry#922

The elasticity parameters in the chain config must not be changed when they are dynamically changed later. These values must reflect the static config that was present pre-Holocene.

In a follow up, we'll move the `[optimism]` section into the `[genesis]` section to avoid this footgun (https://github.com/ethereum-optimism/superchain-registry/issues/969).

### DIFF
--- a/superchain/configs/mainnet/worldchain.toml
+++ b/superchain/configs/mainnet/worldchain.toml
@@ -20,7 +20,7 @@ max_sequencer_drift = 1800
   holocene_time = 1738238400 # Thu 30 Jan 2025 12:00:00 UTC
 
 [optimism]
-  eip1559_elasticity = 2
+  eip1559_elasticity = 10
   eip1559_denominator = 50
   eip1559_denominator_canyon = 250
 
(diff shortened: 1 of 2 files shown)


superchain/configs/mainnet/worldchain.toml
```diff
@@ -20,7 +20,7 @@ max_sequencer_drift = 1800
   holocene_time = 1738238400 # Thu 30 Jan 2025 12:00:00 UTC
 
 [optimism]
-  eip1559_elasticity = 2
```
tynes, line 23: These fields should be in the genesis part of the config. There are also some roles that should be in the genesis part of the config. We should have something dynamic that fetches the values from onchain if we care about what the values are at the moment.

sbvegan (approved):

protolambda (approved):

PR #975, opened 2025-04-14
chain config: move 'optimism' to 'genesis.fee_params'

### LINKED ISSUE #969: Move `[optimism]` section of chain configs into `[genesis]` section
The EIP-1559 config that sits inside the `[optimism]` section of chain configs belongs to the immutable genesis config, so we should move it there and avoid motivating chains to update it when they change it on-chain via Holocene SystemConfig parameter upgrades. 

This has happened in https://github.com/ethereum-optimism/superchain-registry/pull/922 and we revert this now, current releases cannot be used for worldchain sync from genesis (both sepolia and mainnet).

bitwiseguy:
# Overview
Makes the following changes to the chain config .toml struct:

* `batch_inbox_addr` field moved from top level to `[genesis.system_config]`
* `[optimism]` struct renamed/moved to [genesis.fee_params]

# Related Issues
Closes https://github.com/ethereum-optimism/superchain-registry/issues/969


ops/internal/config/chain.go
```diff
@@ -124,12 +123,13 @@ type GenesisRef struct {
...
 
 type SystemConfig struct {
-	BatcherAddr       ChecksummedAddress `json:"batcherAddr" toml:"batcherAddress"`
-	Overhead          common.Hash        `json:"overhead" toml:"overhead"`
-	Scalar            common.Hash        `json:"scalar" toml:"scalar"`
-	GasLimit          uint64             `json:"gasLimit" toml:"gasLimit"`
-	BaseFeeScalar     *uint64            `json:"baseFeeScalar,omitempty" toml:"baseFeeScalar,omitempty"`
-	BlobBaseFeeScalar *uint64            `json:"blobBaseFeeScalar,omitempty" toml:"blobBaseFeeScalar,omitempty"`
+	BatcherAddr       ChecksummedAddress  `json:"batcherAddr" toml:"batcherAddress"`
+	BatchInboxAddr    *ChecksummedAddress `json:"batchInboxAddr" toml:"batch_inbox_addr"`
```
clabby, line 127: This is an interesting one. The batch inbox address isn't actually in the [system config that gets derived](https://github.com/ethereum-optimism/optimism/blob/d032aea4ab029918b4ac0cec6f0a07de08590ed3/op-service/eth/types.go#L502-L502), just [the contract](https://github.com/ethereum-optimism/optimism/blob/fab8931ebed3c64ad3414b4badbdba8dce670146/packages/contracts-bedrock/src/L1/SystemConfig.sol#L282-L285).

The genesis system config vars, in my head, are things that can change (that the protocol supports changing in normal operation) - imo, it's fine to lift into `genesis`, but not `genesis.system_config`.

bitwiseguy: Ahh ok got it. Will lift it one level to `genesis` then. And will update the op-geth and monorepo prs accordingly

sebastianst: Maybe the inconsistency is that we would put this address that can be read from the system config (and that cannot be changed) into `genesis.SystemConfig`, but would still leave others in the `addresses` section. So @clabby @bitwiseguy wdyt about putting the batch inbox address into the `addresses` section instead?

PR #981, opened 2025-04-15
perf: optimize Hash.UnmarshalText with direct byte comparison

GarmashAlex:
Replaced bytes.HasPrefix() with direct byte comparison in Hash.UnmarshalText, removing unnecessary allocation and function call. This change simplifies the code while maintaining identical behavior and removing an unused import.

### DIFF
--- a/validation/types.go
+++ b/validation/types.go
@@ -1,7 +1,6 @@
 package validation
 
 import (
-	"bytes"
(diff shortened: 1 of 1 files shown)


ajsutton:
Thanks but the original is clearer and performance isn't an issue here.

PR #994, opened 2025-04-25
fix(superseed-mainnet): granite and other fork timestamps + timestamp default inheritance

emiliano-conduitxyz:


sbvegan (commented):
This generally looks good to me, there is one note about the automatic hardfork inheritance that I'll relay over to you on slack. I'll ping the platforms team and ask about the `codegen` error and see how we can get that fixed

superchain/configs/mainnet/sseed.toml
```diff
@@ -10,12 +10,16 @@ batch_inbox_addr = "0x8612014a343089F1ddBACfD42baf4Afbf9133593"
 block_time = 2
 seq_window_size = 3600
 max_sequencer_drift = 600
+superchain_time = 0
```
sbvegan, line 13: I recommend removing this for the reason I'll chat through with you on slack

```suggestion
```

sbvegan:
Closing in favor of: https://github.com/ethereum-optimism/superchain-registry/pull/995

PR #1009, opened 2025-05-14
ops: update monorepo import to 877b74f3ada002f78d7cc5149d9417a4437fe3e8

bitwiseguy:
This is a pre-requisite for the interop <> superchain-registry integration because we need to import the latest op-supervisor depset struct.

This pr updates the monorepo import and addresses breaking changes that were caused by the update.


ops/internal/manage/staging.go
```diff
@@ -93,27 +93,27 @@ func InflateChainConfig(st *state.State) (*config.StagedChain, error) {
 	cfg.Roles = config.Roles{
 		SystemConfigOwner: config.NewChecksummedAddress(chainIntent.Roles.SystemConfigOwner),
 		ProxyAdminOwner:   config.NewChecksummedAddress(chainIntent.Roles.L1ProxyAdminOwner),
-		Guardian:          config.NewChecksummedAddress(st.AppliedIntent.SuperchainRoles.Guardian),
+		Guardian:          config.NewChecksummedAddress(st.AppliedIntent.SuperchainRoles.SuperchainGuardian),
```
bitwiseguy, line 96: Is this correct or should it be the OpChainGuardian?

bitwiseguy:
Will update the monorepo import after we can address the op-deployer incompatibilities that get pulled in

PR #1011, opened 2025-05-15
ops: Add an `import_devnet` command

geoknee:
This will pull in multiple chains from a op-deployer state.json file and add them to the staging directory.

Closes https://github.com/ethereum-optimism/optimism/issues/15935

If you do 
```
just import-devnet ../devnets/alphanets/interop-rc-alpha/op-deployer/state.json ../devnets/alphanets/interop-rc-alpha/manifest.yaml
```
then 
```
just sync-staging --l1-rpc-urls=https://ethereum-sepolia-rpc.publicnode.com
```
You get the new superchain created and both chains added. It should work for any number of chains deployed together with op-deployer for a devnet. 

This is _not_ a general purpose command to create a new superchain. It's more tailored to importing devnets, as the name suggests. 

Testing: 

Manually via #1020 (test legacy behavior) for `create_chain` and  #1018 (the new behaviour for `import_devnet`)


justfile
```diff
@@ -50,4 +50,7 @@ check-genesis-integrity: (_run_ops_bin 'check_genesis_integrity')
 create-config SHORTNAME FILENAME:
 	@just _run_ops_bin "create_config" "--shortname {{SHORTNAME}} --state-filename $(realpath {{FILENAME}})"
 
+import-devnet STATEFILE MANIFESTFILE:
+	@just _run_ops_bin "import_devnet" "--state-filename $(realpath {{STATEFILE}}) --manifest-path $(realpath {{MANIFESTFILE}})"
```
geoknee, line 54: For now I am just adding another command, I did not change the behaviour of the existing `create-config` command (although much of the backing code now supports multiple chains being added together).

ops/cmd/check_chainlist/main.go
```diff
@@ -21,38 +21,40 @@ func mainErr() error {
...
+		output.WriteOK("fetched global chain IDs")
 
-	entry, ok := globalChainData[chainCfg.ChainID]
-	if !ok {
-		return fmt.Errorf("chain ID %d not found in global chain data", chainCfg.ChainID)
-	}
-	output.WriteOK("chain ID %d found in global chain data", chainCfg.ChainID)
+		entry, ok := globalChainData[chainCfg.ChainID]
+		if !ok {
+			return fmt.Errorf("chain ID %d not found in global chain data", chainCfg.ChainID)
```
geoknee, line 44: We could be a bit fancier here and collect all of the errors so we don't just fail as soon as there is one. Not sure it is worth it, though.

ops/cmd/import_devnet/main.go
```diff
@@ -0,0 +1,137 @@
...
+		Name:      "state-filename",
+		Usage:     "Filename of an op-deployer state file.",
+		Required:  true,
+		TakesFile: true,
+	}
+)
+var (
+	ManifestPath = &cli.StringFlag{
+		Name:      "manifest-path",
+		Usage:     "Path to a manifest.yaml file specifying chain names.",
```
geoknee, line 28: Might be nice to be able to provide either this file _or_ the superchain name and a comma separated list of individual chain names.

ops/cmd/import_devnet/main.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	type manifest struct {
+		Name string `yaml:"name"`
+		L2   struct {
+			Chains []struct {
+				Name string `yaml:"name"`
+			} `yaml:"chains"`
+		} `yaml:"l2"`
+	}
```
geoknee, line 70: If we are leaning on this data structure, it could be pulled out into a different package so it is more reusable.

ops/cmd/import_devnet/main.go
```diff
@@ -0,0 +1,137 @@
...
+		return fmt.Errorf(
+			"number of chains in manifest file (%d) does not match number of chains in state file (%d)",
+			len(m.L2.Chains), len(st.AppliedIntent.Chains))
+	}
+
+	output.WriteOK("inflating chain configs")
+	for i := 0; i < len(st.AppliedIntent.Chains); i++ {
+		cfg, err := manage.InflateChainConfig(&st, i)
+		if err != nil {
+			return fmt.Errorf("failed to inflate chain config: %w", err)
```
geoknee, line 86: maybe log out the index as well

ops/cmd/import_devnet/main.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+
+		// just use devnet name with numerical suffix
+		// OR parse the manifest.yaml file and get the name from there
+		cfg.ShortName = m.L2.Chains[i].Name
+		cfg.Name = m.L2.Chains[i].Name
+
+		cfg.Superchain = config.Superchain(m.Name) // Each devnet forms its own unique superchain
```
geoknee, line 94: Is this a safe assumption going forward?

ops/cmd/import_devnet/main.go
```diff
@@ -0,0 +1,137 @@
...
+		return fmt.Errorf("failed to read manifest file: %w", err)
+	}
+
+	if len(m.L2.Chains) != len(st.AppliedIntent.Chains) {
+		return fmt.Errorf(
+			"number of chains in manifest file (%d) does not match number of chains in state file (%d)",
+			len(m.L2.Chains), len(st.AppliedIntent.Chains))
+	}
+
+	output.WriteOK("inflating chain configs")
```
geoknee, line 82: Quite a bit of code duplication here with the `create_config` command, could be tidied up.

ops/cmd/import_devnet/main.go
```diff
@@ -0,0 +1,137 @@
...
+
+	sD := config.SuperchainDefinition{
+		Name:                   m.Name,
+		ProtocolVersionsAddr:   config.NewChecksummedAddress(st.SuperchainDeployment.ProtocolVersionsProxyAddress),
+		SuperchainConfigAddr:   config.NewChecksummedAddress(st.SuperchainDeployment.SuperchainConfigProxyAddress),
+		OPContractsManagerAddr: config.NewChecksummedAddress(st.ImplementationsDeployment.OpcmAddress),
+		Hardforks:              config.Hardforks{}, // superchain wide hardforks are added after chains are in the registry.
+		L1: config.SuperchainL1{
+			ChainID: st.AppliedIntent.L1ChainID,
+			// TODO grab this from a lookup or prompt user to fill it out afterwards
```
geoknee, line 125: I'm referring to the PublicRPC here. Chances are it is sepolia, but either way we know some RPCs for sepolia so we could just look one up with the chain id.

ops/cmd/print_staging_report/main.go
```diff
@@ -93,82 +93,85 @@ func PrintStagingReport(cliCtx *cli.Context) error {
...
 
-	chainCfg, err := manage.StagedChainConfig(wd)
-	if errors.Is(err, manage.ErrNoStagedConfig) {
-		output.WriteOK("no staged chain config found, exiting")
-		return nil
-	}
-	if err != nil {
-		return fmt.Errorf("failed to get staged chain config: %w", err)
-	}
+	stagedChainCfgs, err := manage.StagedChainConfigs(wd)
```
geoknee, line 96: We could test this command out by running 

```
just import-devnet ../devnets/alphanets/interop-rc-alpha/op-deployer/state.json ../devnets/alphanets/interop-rc-alpha/manifest.yaml
```
and raising a PR.

ops/cmd/import_devnet/main.go
```diff
@@ -0,0 +1,138 @@
...
+			return fmt.Errorf("failed to write chain config: %w", err)
+		}
+
+		output.WriteOK("writing genesis")
+		if err := manage.WriteGenesis(wd, path.Join(stagingDir, cfg.ShortName+".json.zst"), genesis); err != nil {
+			return fmt.Errorf("failed to write genesis: %w", err)
+		}
+	}
+
+	output.WriteOK("writing superchain manifest")
```
geoknee, line 113:
```suggestion
	output.WriteOK("writing superchain definition file")
```

ops/cmd/import_devnet/main.go
geoknee: We should be able to import it from the monorepo devnet-sdk package actually.

ops/cmd/import_devnet/main.go
```diff
@@ -0,0 +1,138 @@
...
+		L1: config.SuperchainL1{
+			ChainID: st.AppliedIntent.L1ChainID,
+			// TODO grab this from a lookup or prompt user to fill it out afterwards
+		},
+	}
+
+	// We write this to {m.Name}.superchain-toml and when it gets sync'ed later
+	// it will be moved to the appropriate superchain directory and renamed
+	// to superchain.toml. Validation and conflict resolution will be handled
+	// by the sync staging command.
```
geoknee, line 130: The file does have a `name` field, does that mean we don't need this magic.

ops/cmd/import_devnet/main.go
geoknee: Hmm seems like the struct we need is actually in one of our private repos https://github.com/ethereum-optimism/infrastructure-services/blob/e19a7a598b74d6cd715db5fe97e6840a19419db6/netchef/pkg/manifest/manifest.go#L10-L17

ops/cmd/import_devnet/main.go
geoknee: Or we can follow the pattern we already have where a staged chain config has extra fields which store this information (they are not copied over when we run `sync_staged`.

ops/cmd/import_devnet/main.go
geoknee: Actually we have to provide at least one RPC to the sync_staging command, so we can just inject when running that.

ops/cmd/import_devnet/main.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/urfave/cli/v2"
+)
+
+var (
+	StateFilename = &cli.StringFlag{
+		Name:      "state-filename",
+		Usage:     "Filename of an op-deployer state file.",
+		Required:  true,
+		TakesFile: true,
+	}
```
bitwiseguy, line 23: Might want to change this to `StatePath` for consistency with the `ManifestPath` flag

ops/cmd/print_staging_report/main.go
```diff
@@ -93,82 +93,85 @@ func PrintStagingReport(cliCtx *cli.Context) error {
...
+		case config.MainnetSuperchain:
+			stdConfigs = validation.StandardConfigParamsMainnet
+			stdRoles = validation.StandardConfigRolesMainnet
+			l1RPCURL = cliCtx.String(MainnetRPCURLFlag.Name)
+		case config.SepoliaSuperchain:
+			stdConfigs = validation.StandardConfigParamsSepolia
+			stdRoles = validation.StandardConfigRolesSepolia
+			l1RPCURL = cliCtx.String(SepoliaRPCURLFlag.Name)
+		default:
+			return fmt.Errorf("unsupported superchain: %s", chainCfg.Superchain)
```
bitwiseguy, line 134: Won't this always fail for a new devnet superchain because it will hit the default case here. I wonder if we should change the handling of superchains across this ops code. Instead of using a string alias [here](https://github.com/ethereum-optimism/superchain-registry/blob/main/ops/internal/config/superchain.go#L12), we could just use a string so that its easier to dynamically add valid superchains. There could be an `init` function in the `ops/internal/config` package which just scans for all the dirs nested under `superchain/configs`, and that determines which superchains are valid. Then any check for superchain validity could be run agains that slice. Something like this:
```
var ValidSuperchains []string

func init() {
	repoRoot, err := paths.FindRepoRoot()
	if err != nil {
		panic(err)
	}

	configsDir := paths.SuperchainConfigsDir(repoRoot)
	// Read the directory entries
	entries, err := os.ReadDir(configsDir)
	if err != nil {
		panic(fmt.Sprintf("failed to read superchain configs directory: %v", err))
	}

	// Add each directory name to the ValidSuperchains slice
	for _, entry := range entries {
		if entry.IsDir() {
			ValidSuperchains = append(ValidSuperchains, entry.Name())
		}
	}
}
```

geoknee: Yes we need to permit essentially any string as a superchain name going forward. 

Also agree that populating a list of existing superchains could be useful, as long as we don't block the addition of entirely fresh superchains.

ops/cmd/print_staging_report/main.go
geoknee: This feels like a bit of distinct scope that we can pinch of to allow work to get unblocked and keep changesets small and easy to review. Filed an issue here #1017

ops/cmd/print_staging_report/main.go
geoknee: This would be part of this scope #1017

ops/cmd/import_devnet/main.go
geoknee: [6e560c2](https://github.com/ethereum-optimism/superchain-registry/pull/1011/commits/6e560c2b2846baa452917d379392e88d51dcae09)

ops/cmd/print_staging_report/main.go
geoknee: If we are descoping staging report from this PR, might be best to revert the changes to this file on this PR.

ops/cmd/import_devnet/main.go
geoknee: On the other hand, for any devnet being added without a manifest file on hand, could the user just create one with the minimal fields? Might actually be easier than threading everything through the CLI interface.

ops/internal/config/superchain.go
```diff
@@ -44,7 +39,7 @@ func FindValidL1URL(ctx context.Context, lgr log.Logger, urls []string, supercha
 		}
 
 		if err := validateL1ChainID(ctx, url, superchainId); err != nil {
-			lgr.Warn("l1-rpc-url has mismatched l1 chainId", "urlIndex", i)
+			lgr.Warn("l1-rpc-url has mismatched l1 chainId", "urlIndex", i, "err", err)
```
geoknee, line 42: Filed #1016

geoknee: (Because I was seeing 403s in CI)

ops/internal/generate/generate.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/ethereum-optimism/optimism/op-deployer/pkg/deployer/state"
+	"github.com/ethereum-optimism/superchain-registry/ops/internal/config"
+	"github.com/ethereum-optimism/superchain-registry/ops/internal/manage"
+	"github.com/ethereum-optimism/superchain-registry/ops/internal/output"
+	"github.com/ethereum-optimism/superchain-registry/ops/internal/paths"
+)
+
+// Generate creates a chain config and genesis file for the chain at index idx in the given state file
+// using the given shortName (and optionally, name and superchain identifier).
+// It writes these files to the staging directory.
```
geoknee, line 17: This is code that is being moved from `create_config` so that it can be shared with the new `import_devnet` command.

ops/internal/config/superchain.go
```diff
@@ -18,12 +18,7 @@ const (
 )
 
 func ParseSuperchain(in string) (Superchain, error) {
-	switch Superchain(in) {
-	case MainnetSuperchain, SepoliaSuperchain, SepoliaDev0Superchain:
-		return Superchain(in), nil
-	default:
-		return "", fmt.Errorf("unknown superchain: '%s'", in)
-	}
+	return Superchain(in), nil
```
geoknee, line 21: This is a bit smelly, having a string enum that we can just cast any string to. Might be better to just rip out the string enum.

bitwiseguy: Yes this will never error because the validation has been removed. So I would say we need to remove this function now and move towards using a `string` (instead of string alias) as the superchain type. Then if we use the `init()` function and populate a `var ValidSuperchains map[string]config.SuperchainDefinition` by reading from every `superchain/configs` dir, then we can query that map to determine if a given superchain is valid instead of calling the `ParseSuperchain` function

ops/internal/generate/generate.go
bitwiseguy: Instead of creating this new `generate` package could we just move this function inside of the `internal/manage` package? Also maybe rename from `Generate` to something more descriptive like `GenerateChainArtifacts`

ops/internal/manage/staging.go
```diff
@@ -172,27 +172,47 @@ func CopyDeployConfigHFTimes(src *genesis.UpgradeScheduleDeployConfig, dst *conf
...
 	}
+	return chainCfgs, nil
+}
 
-	cfgFilename := tomls[0]
-	chainCfg := new(config.StagedChain)
-	if err := paths.ReadTOMLFile(cfgFilename, chainCfg); err != nil {
-		return nil, fmt.Errorf("failed to read %s: %w", cfgFilename, err)
+// StagedSuperchainDefinition finds a superchain.toml file in the staging directory
+// (if it exits) and returns the parsed SuperchainDefinition struct.
```
bitwiseguy, line 203: typo: `exits` --> `exists`

ops/cmd/import_devnet/main.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/ethereum-optimism/superchain-registry/ops/internal/generate"
+	"github.com/ethereum-optimism/superchain-registry/ops/internal/output"
+	"github.com/ethereum-optimism/superchain-registry/ops/internal/paths"
+
+	"github.com/urfave/cli/v2"
+)
+
+var (
+	StatePath = &cli.StringFlag{
+		Name:      "state-filename",
```
bitwiseguy, line 19: Should probably change this to `state-path` for consistency

ops/cmd/sync_staging/main.go
```diff
@@ -57,8 +57,38 @@ func action(cliCtx *cli.Context) error {
 	}
 
 	stagingDir := paths.StagingDir(wd)
+	stagedSuperchainDefinition, err := manage.StagedSuperchainDefinition(wd)
+
+	l1RpcUrl := ""
+	if err == nil {
+		output.WriteOK("superchain definition found, finding L1 RPC URL...")
+		l1RpcUrl, err = config.FindValidL1URL(cliCtx.Context,
+			log.NewLogger(log.NewTerminalHandlerWithLevel(os.Stderr, log.LevelInfo, false)),
```
bitwiseguy, line 66: You're creating two separate loggers. One here and one down in the codegen block of code. Might be better to just create one at the beginning of this action function and use it in both places

ops/cmd/sync_staging/main.go
```diff
@@ -57,8 +57,38 @@ func action(cliCtx *cli.Context) error {
...
+			l1RpcUrls, stagedSuperchainDefinition.L1.ChainID)
+		if err != nil {
+			return fmt.Errorf("failed to find valid L1 RPC URL: %w", err)
+		}
+		stagedSuperchainDefinition.L1.PublicRPC = l1RpcUrl
+		err = manage.WriteSuperchainDefinition(
+			path.Join(wd, "superchain", "configs", stagedSuperchainDefinition.Name, "superchain.toml"),
```
bitwiseguy, line 73: There are some helper functions in the `interna/paths` package that would be could to use here instead of building the path in this way.

geoknee: [f58513e](https://github.com/ethereum-optimism/superchain-registry/pull/1011/commits/f58513eccde3c420cee0e89606b2eec0c425229f)

ops/cmd/sync_staging/main.go
geoknee: [42d9ab6](https://github.com/ethereum-optimism/superchain-registry/pull/1011/commits/42d9ab669b96dbb402697a2e4e839cc5675d8425)

ops/cmd/import_devnet/main.go
geoknee: [6e560c2](https://github.com/ethereum-optimism/superchain-registry/pull/1011/commits/6e560c2b2846baa452917d379392e88d51dcae09)

ops/internal/manage/staging.go
geoknee: [b984051](https://github.com/ethereum-optimism/superchain-registry/pull/1011/commits/b984051c672dc7fc6697027b6506f8477d5fb27d)

ops/internal/generate/generate.go
geoknee: [8e66ad6](https://github.com/ethereum-optimism/superchain-registry/pull/1011/commits/8e66ad6f6edbcd55d419e97a4944e5a08f6de022)

ops/internal/config/superchain.go
geoknee: `Superchain` is actually not an alias presently but a distinct type with string as its underlying type. In [c2f83c1](https://github.com/ethereum-optimism/superchain-registry/pull/1011/commits/c2f83c1e9815f16ff5ceb8855f73bc0c1973ee50) I changed it so it _is_ an alias (meaning, just another way of saying `string`). See this article https://go.dev/blog/alias-names. It makes the refactor much simpler to achieve.

geoknee: I didn't put in the `ValidSuperchains` mapping yet because we already have `paths.Superchains()` which returns a slice of strings with the existing superchains (computed at runtime). This is used in a bunch of places and I think does the job we need here.

ops/cmd/print_staging_report/main.go
geoknee: In the end I kept most of the changes (namely, looping over multiple chains) but added an early `continue` statement for any chain not in the mainnet or sepolia superchains.

ops/cmd/print_staging_report/main.go
```diff
@@ -93,82 +93,86 @@ func PrintStagingReport(cliCtx *cli.Context) error {
...
+		case config.MainnetSuperchain:
+			stdConfigs = validation.StandardConfigParamsMainnet
+			stdRoles = validation.StandardConfigRolesMainnet
+			l1RPCURL = cliCtx.String(MainnetRPCURLFlag.Name)
+		case config.SepoliaSuperchain:
+			stdConfigs = validation.StandardConfigParamsSepolia
+			stdRoles = validation.StandardConfigRolesSepolia
+			l1RPCURL = cliCtx.String(SepoliaRPCURLFlag.Name)
+		default:
+			output.WriteNotOK("skipping staging report for chain %s in unsupported superchain: %s",
```
geoknee, line 119:
```suggestion
			output.WriteOK("skipping staging report for chain %s in unsupported superchain: %s",
```

ops/cmd/import_devnet/main.go
bitwiseguy: I think this is ok for now. I don't think we will have a scenario where a devnet's L2s will join the registry in multiple steps. Rather, I think its safe to assume all of a devnet's L2s will join at the same time, from the same state.json

bitwiseguy (approved):
Looks good!

PR #1013, opened 2025-05-16
ops: add depset checker

bitwiseguy:
# Overview
Adds a new `DepsetChecker` object to perform the following interop validations:
1. Offchain: every chain in the depset has the exact same depset in their config toml
2. Offchain: `StaticConfigDependencySet.ActivationTime` must be ≥ to that chain’s `[hardforks.interop_time]`
3. Onchain: every chain in the depset has the same DisputeGameFactoryProxy and EthLockboxProxy address
    * Note: EthLockboxProxy check is currently disabled. Need to be able to update op-fetcher import (via monorepo) before we can make this check as currently designed. Blocked by https://github.com/ethereum-optimism/optimism/issues/15934

`DepsetChecker` is wrapped by the new `ops/cmd/check_depsets` program, which is invoked by a new circleci job `check-depsets` for every pr.

# Details
`DepsetChecker` takes three inputs:
1. logger
2. `AddressesJSON` mapping of chainId --> addresses: contains addresses for every chain it needs to validate. In production, this will be data from the codegen'ed `addresses.json` file that is populated by `op-fetcher` after it has retrieved all addresses from onchain. This means `DepsetChecker` MUST run after `codegen` in the production ci.
3. `[]DiskChainConfig` slice, which includes the data from each chains .toml config. In production, all chain configs from all superchains in the child dirs within `superchain/configs` will be provided.

# Related issues
* https://github.com/ethereum-optimism/optimism/issues/15870
* https://github.com/ethereum-optimism/optimism/issues/15872
* https://github.com/ethereum-optimism/optimism/issues/15871


ops/internal/manage/depsets.go
```diff
@@ -0,0 +1,230 @@
...
+		// collect all chains in the depset, then process them
+		depsetCfgs := []DiskChainConfig{}
+		depset := cfg.Config.Interop.Dependencies()
+
+		for chainId := range depset {
+			chainIdUint64, ok := chainId.Uint64()
+			if !ok {
+				return fmt.Errorf("chain ID is not a uint64: %v", chainId)
+			}
+			depsetCfgs = append(depsetCfgs, dc.diskChainCfgs[chainIdUint64])
```
ajsutton, line 68: Should we check the config exists in this map?

ops/internal/manage/depsets.go
```diff
@@ -0,0 +1,230 @@
...
+	errDepsetLengths         = errors.New("inconsistent depset lengths")
+	errInconsistentDepsets   = errors.New("inconsistent depset values")
+	errMissingAddress        = errors.New("missing address")
+)
+
+type DepsetChecker struct {
+	lgr             log.Logger
+	addrs           config.AddressesJSON
+	diskChainCfgs   map[uint64]DiskChainConfig
+	processedChains map[uint64]bool
```
ajsutton, line 25: Any reason not to just use `eth.ChainID` as the key and for all the chain ID types? Would save checking it is a uint64 and a bunch of conversions later on.  eth.ChainID is safe to use as a map key.

ajsutton (commented):
Logic looks good to me.

ops/internal/manage/depsets.go
bitwiseguy: The issue is that in the chain config tomls chainId is stored as a `uint64` currently ([ref](https://github.com/ethereum-optimism/superchain-registry/blob/main/ops/internal/config/chain.go#L72)). So I think we'll have to do roughly the same amount of conversions between types either way. It just moves the conversion if we change the type you pointed out to be `eth.ChainId`. Unless we change the chain config toml chainId type

ajsutton: That counts as reason enough to me. :) Its not a big deal.

ops/internal/manage/depsets.go
bitwiseguy: Done in https://github.com/ethereum-optimism/superchain-registry/pull/1013/commits/24f4a3f4afacca61103674b0ac63a0d253a6aa9e

ops/internal/manage/depsets.go
```diff
@@ -0,0 +1,243 @@
...
+		diskChainCfgs:   make(map[uint64]DiskChainConfig),
+		processedChains: make(map[uint64]bool),
+	}
+
+	for _, cfg := range cfgs {
+		dc.diskChainCfgs[cfg.Config.ChainID] = cfg
+	}
+
+	logger.Info("loaded depset checker instance with chain configs", "numDiskCfgs", len(cfgs))
+	return dc, nil
```
geoknee, line 43: Is there any point returning an error if it is always `nil`?

ops/internal/manage/depsets.go
```diff
@@ -0,0 +1,243 @@
...
+		for _, dep := range depsetCfgs {
+			dc.lgr.Info("processed chain", "chainID", dep.Config.ChainID)
+			dc.processedChains[dep.Config.ChainID] = true
+			dc.chainsProcessed++
+		}
+	}
+
+	numChains := len(dc.diskChainCfgs)
+	if dc.chainsProcessed != numChains {
+		return fmt.Errorf("chainsProcessed (%d) does not match totalChainCgs (%d)", dc.chainsProcessed, numChains)
```
geoknee, line 92:
```suggestion
		return fmt.Errorf("chainsProcessed (%d) does not match totalChainCfgs (%d)", dc.chainsProcessed, numChains)
```

ops/internal/manage/depsets.go
```diff
@@ -0,0 +1,243 @@
...
+		addrs, err := dc.getAndValidateAddresses(cfg.Config.ChainID)
+		if err != nil {
+			return err
+		}
+
+		if strings.ToLower((*addrs.DisputeGameFactoryProxy).String()) != firstDisputeGameFactoryProxy {
+			return fmt.Errorf("DisputeGameFactoryProxy address mismatch for chain %d, expected %s, got %s",
+				cfg.Config.ChainID, firstDisputeGameFactoryProxy, addrs.DisputeGameFactoryProxy)
+		}
+		// TODO: re-enable this once we can pull in the updated op-fetcher from monorepo
```
geoknee, line 212: is it worth making a tracking issue and recording it here?

ops/internal/manage/depsets.go
```diff
@@ -0,0 +1,243 @@
...
+	logger.Info("loaded depset checker instance with chain configs", "numDiskCfgs", len(cfgs))
+	return dc, nil
+}
+
+func (dc *DepsetChecker) Check() error {
+	for _, cfg := range dc.diskChainCfgs {
+		if dc.processedChains[cfg.Config.ChainID] {
+			// already processed this chain
+			continue
+		}
```
geoknee, line 51: Nit. This worried me when I fast saw it, because I thought it meant we wouldn't revisit a chain at all once we passed over it once. Digging in further, I think we are fine, but might be nice to explain why here. 
```suggestion
		if dc.processedChains[cfg.Config.ChainID] {
			// already processed this chain (it may be revisited anyway during checkOffChain)
			continue
		}
```

For reference, here is the test case I was worried about 98a1d0fda815f6e8173e57c751bf840d5c24022b. Happily the checker does report an error.

geoknee (approved):

ops/internal/manage/collect_test.go
```diff
@@ -24,13 +24,13 @@ func TestCollectChainConfigs(t *testing.T) {
 		{
 			ShortName:  "op",
 			Filepath:   paths.ChainConfig("testdata", "sepolia", "op"),
-			Superchain: config.SepoliaSuperchain,
+			Superchain: "sepolia",
```
mslipper, line 27: Why did you change this from the enum to strings?

ops/internal/manage/collect.go
```diff
@@ -19,7 +19,7 @@ const CollectorConcurrency = 8
 type DiskChainConfig struct {
 	ShortName  string
 	Filepath   string
-	Superchain config.Superchain
+	Superchain string
```
mslipper, line 22: Why did you change this?

mslipper (approved):
lgtm but i'm wondering about why you changed some of the types

ops/internal/manage/collect.go
bitwiseguy: In the pending pr to add a `import_devnet` command ([ref](https://github.com/ethereum-optimism/superchain-registry/pull/1011)), we need to be able to create new superchains. Therefore, we need to be able to have a dynamic list of acceptable superchains. Using an enum to define the list of acceptable superchains makes it difficult to add any new ones. So we need the change there. See [this file](https://github.com/ethereum-optimism/superchain-registry/blob/21e2492579dfa6bf99af95764892e8fa930cb413/ops/internal/config/superchain.go#L12-L18) for the relevant changes.

Additionally, within this pr I wanted to be able to pass a `testdata` dir as the input to `CollectChainConfigs` to fully test the end-to-end flow of the new `cmd/check_depsets`. Therefore, I need to be able to pass a value outside of the defined enum. I believe we should merge the `import_devnet` pr first, then rebase this one on top so the two modifications you've highlighted will change slightly (assuming you're good with the changes in that other pr)

ops/internal/manage/collect_test.go
bitwiseguy: See other comment. This change can be reverted after rebasing on top of `import_devnet` pr

ops/internal/manage/depsets.go
bitwiseguy: Removed in https://github.com/ethereum-optimism/superchain-registry/pull/1013/commits/bbb7f18834dfaa2f0183640e659280aad20e2ebb

ops/internal/manage/depsets.go
bitwiseguy: Done in https://github.com/ethereum-optimism/superchain-registry/pull/1013/commits/c03558c4a594cd6850615cebd73e0f6a8960536e

PR #1015, opened 2025-05-19
test importing a devnet

geoknee:
This PR is here to trigger the chain compliance report tool to run against #1011 with an example devnet in the staging area.

Has a few hacks in it at the moment, which will not want to commit. Exposing a few spots where we have baked in assumptions about which superchains are in play, etc.


ops/internal/report/l2.go
```diff
@@ -56,15 +58,35 @@ func DiffL2Genesis(
...
+	if err != nil {
+		return common.Hash{}, nil, fmt.Errorf("failed to get working directory: %w", err)
+	}
+	stagedSuperchainDefinition, err := manage.StagedSuperchainDefinition(wd)
+
+	if err != nil {
+		l1ChainID = stagedSuperchainDefinition.L1.ChainID
+	} else {
+		if chainCfg.Superchain == config.MainnetSuperchain {
+			l1ChainID = 1
```
geoknee: If there is no _staged_ superchain file, we could look up the superchain definition by `chainCfg.Superchain` key in a mapping that is populated at package init. See https://github.com/ethereum-optimism/superchain-registry/pull/1011#discussion_r2095738865

.staging/interop-rc-alpha-1.toml
```diff
@@ -0,0 +1,72 @@
+name = "interop-rc-alpha-1"
```
geoknee, line 1: We might want to choose a healthier example devnet to test this with.

PR #1020, opened 2025-05-20
test adding two chains post #15935

geoknee:
The compliance report still gets generated for one of the two chains in the staging dir. 

The report contains errors, as op-deployer is having trouble. I believe these are preexisting issues, but we should confirm. 

- [x] We can do that by rebasing the camp PR on latest `main` and seeing if we hit the same errors. And then fixing them somehow, we might need #1014 to do this. Did that here https://github.com/ethereum-optimism/superchain-registry/pull/1022


geoknee:
Although some CI checks are failing here, we believe they are not caused by the changes on #1011 .

PR #1022, opened 2025-05-20
test adding camp again but with an updated branch

geoknee:


geoknee:
This was just an experiment for local testing purposes.

PR #970, opened 2025-04-10
Add radius testnet chain

iujiban:
# Adding a New Chain

This PR adds [Radius-testnet] to the registry.

## Checklist

- [x] I have declared the chain at the appropriate [Superchain Level](../docs/glossary.md#superchain-level-and-rollup-stage).
- [x] I have run `just codegen` to ensure that all generated files are created.
- [x] I have run `just codegen` to ensure that the chainlist and other generated files are up-to-date and include my chain.
- [x] I have checked `Allow edits from maintainers` on this PR.

PR #1024, opened 2025-05-21
Add radius testnet

Wazabie:
This PR adds radius testnet to SCR

I ran just sync staging and codegen on #970 

NOTE: #970 has two zst in the staging directory. This PR has used the file originally called genesis.json.zst from PR #970


Wazabie:
Closing this as the wrong genesis.json.zst was used

PR #1028, opened 2025-05-21
add a test case for transitive dependencies in depset_test.go

geoknee:


ops/internal/manage/depsets_test.go
```diff
@@ -75,6 +75,17 @@ func TestDepsetChecker(t *testing.T) {
 		require.NoError(t, err)
 		require.Error(t, checker.Check())
 	})
+
+	t.Run("invalid depsets (transience)", func(t *testing.T) {
+		addrs := loadAddresses(t, "testdata/transience_invalid/addresses.json")
+		chains, err := CollectChainConfigs("testdata/transience_invalid")
```
bitwiseguy, line 81: Need to add `require.NoError(t, err)` on the line below

bitwiseguy: Also, since the `testdata/` dir is shared by the whole `manage` package, could we nest your new test files in a `testdata/depsets_invalid/transience` dir to make its purpose clearer?

PR #1029, opened 2025-05-22
ops/manage: inflate dependency set from op-deployer state file in InflateChainConfig

geoknee:
Closes [#15875](https://github.com/ethereum-optimism/optimism/issues/15875)

### DIFF
--- a/ops/internal/manage/staging.go
+++ b/ops/internal/manage/staging.go
@@ -7,6 +7,7 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/ethereum-optimism/optimism/devnet-sdk/proofs/prestate"
 	"github.com/ethereum-optimism/optimism/op-chain-ops/genesis"
 	"github.com/ethereum-optimism/optimism/op-deployer/pkg/deployer/inspect"
 	"github.com/ethereum-optimism/optimism/op-deployer/pkg/deployer/state"
@@ -116,6 +117,12 @@ func InflateChainConfig(st *state.State, idx int) (*config.StagedChain, error) {
 		PermissionedDisputeGame:           config.NewChecksummedAddress(chainState.PermissionedDisputeGameAddress),
 	}
 
+	// Dependency Set Inflation
+	if isChainInDependencySet(chainID, st.InteropDepSet) {
+		cfg.Interop = new(config.Interop)
+		cfg.Interop.Dependencies = convertToSCRDepset(st.InteropDepSet)
+	}
+
 	return cfg, nil
 }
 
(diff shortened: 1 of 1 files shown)


ops/internal/manage/staging.go
```diff
@@ -215,3 +222,39 @@ func StagedSuperchainDefinition(rootP string) (*config.SuperchainDefinition, err
...
+func convertToSCRDepset(depSet *prestate.DependencySet) map[string]config.Dependency {
+	if depSet == nil {
+		return nil
+	}
+	if depSet.Dependencies == nil {
+		return nil
+	}
+	ds := make(map[string]config.Dependency, len(depSet.Dependencies))
+	for k, v := range depSet.Dependencies {
+		decimalChainId, err := common.HexToHash(k).Big().MarshalText()
```
geoknee, line 249: Converting from hex to decimal here to be consistent with what we do elsewhere in the registry.

ops/internal/manage/staging.go
```diff
@@ -215,3 +222,39 @@ func StagedSuperchainDefinition(rootP string) (*config.SuperchainDefinition, err
...
 	return sM, err
 }
+
+// isChainInDependencySet checks if a chain ID is in the supplied dependency set
+func isChainInDependencySet(chainID common.Hash, depSet *prestate.DependencySet) bool {
+	if depSet == nil {
+		return false
+	}
+	for id := range depSet.Dependencies {
+		if id == chainID.Hex() {
```
geoknee, line 232: Potential for a footgun here since the op-deployer struct uses `common.Hash` for chain Ids in some places and `string` in others. So if we compare hex encoding with decimal encoding we could miss a chain id match.

geoknee:
Just realised that our monorepo reference is a stale and updating it requires some changes to the way we use op-depoyer types and utils. Might solve the chain id comparison issue above.

geoknee:
Updated the monorepo import and fixed a bunch of compiler errors. I think the changes to the `state.json` schema are not backwards compatible though, so the `import_devnet` or `create_config` commands are going to fail if they use files with the old format.

geoknee:
Before updating the monorepo import, I tested importing the `interop-rc-alpha` devnet but after modifying the `state.json` to add

```json
  "interopDepSet": {
    "dependencies": {
      "0x00000000000000000000000000000000000000000000000000000000190a85c3": {
        "chainIndex": 0,
        "activationTime": 99,
        "historyMinTime": 99
      },
      "0x00000000000000000000000000000000000000000000000000000000190a85c4": {
        "chainIndex": 1,
        "activationTime": 99,
        "historyMinTime": 99
      }
    }
  },
```
It seemed to work.

ops/internal/report/l2.go
```diff
@@ -147,14 +152,35 @@ func DiffL2Genesis(
...
+			// BaseFeeVaultMinimumWithdrawalAmount:      cfg.BaseFeeVaultMinimumWithdrawalAmount.ToInt(),
+			// BaseFeeVaultWithdrawalNetwork:            big.NewInt(int64(cfg.BaseFeeVaultWithdrawalNetwork.ToUint8())),
+			L1FeeVaultRecipient: common.Address(*&chainCfg.L1FeeVaultRecipient),
+			// L1FeeVaultMinimumWithdrawalAmount: chainCfg.L1FeeVaultMinimumWithdrawalAmount.ToInt(),
+			// L1FeeVaultWithdrawalNetwork:       big.NewInt(int64(chainCfg.L1FeeVaultWithdrawalNetwork.ToUint8())),
+			// GovernanceTokenOwner: common.Address(*&chainCfg.Addresses.GovernanceTokenOwner),
+			// Fork:                 big.NewInt(chainCfg.SolidityForkNumber(1)),
+			UseInterop:       true,
+			EnableGovernance: chainCfg.GovernedByOptimism,
+			// FundDevAccounts:      chainCfg.f,
```
geoknee, line 182: Highlighting that this section needs some attention.

PR #1030, opened 2025-05-22
ops/manage: inflate dependency set from op-deployer state file in InflateChainConfig

geoknee:
Closes [#15875](https://github.com/ethereum-optimism/optimism/issues/15875)

⚠️ It is a known issue that this code will need to change when we update the op-deployer structs which are imported from the monorepo. Doing that in a backwards-compatible way is not currently possible but is being worked towards, see #1014

### DIFF
--- a/ops/internal/manage/staging.go
+++ b/ops/internal/manage/staging.go
@@ -7,6 +7,7 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/ethereum-optimism/optimism/devnet-sdk/proofs/prestate"
 	"github.com/ethereum-optimism/optimism/op-chain-ops/genesis"
 	"github.com/ethereum-optimism/optimism/op-deployer/pkg/deployer/inspect"
 	"github.com/ethereum-optimism/optimism/op-deployer/pkg/deployer/state"
@@ -116,6 +117,12 @@ func InflateChainConfig(st *state.State, idx int) (*config.StagedChain, error) {
 		PermissionedDisputeGame:           config.NewChecksummedAddress(chainState.PermissionedDisputeGameAddress),
 	}
 
+	// Dependency Set Inflation
+	if isChainInDependencySet(chainID, st.InteropDepSet) {
+		cfg.Interop = new(config.Interop)
+		cfg.Interop.Dependencies = convertToSCRDepset(st.InteropDepSet)
+	}
+
 	return cfg, nil
 }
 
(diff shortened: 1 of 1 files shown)


ops/internal/manage/staging.go
```diff
@@ -215,3 +222,39 @@ func StagedSuperchainDefinition(rootP string) (*config.SuperchainDefinition, err
...
 	return sM, err
 }
+
+// isChainInDependencySet checks if a chain ID is in the supplied dependency set
+func isChainInDependencySet(chainID common.Hash, depSet *prestate.DependencySet) bool {
+	if depSet == nil {
+		return false
+	}
+	for id := range depSet.Dependencies {
+		if id == chainID.Hex() {
```
geoknee, line 232: We have to be careful here because we are comparing a `string` (we don't really know if this is a decimal or hex encoding, but it seems that it tends to be a hex encoded)  with a `common.Hash`.

ops/internal/manage/staging.go
```diff
@@ -215,3 +222,39 @@ func StagedSuperchainDefinition(rootP string) (*config.SuperchainDefinition, err
...
+func convertToSCRDepset(depSet *prestate.DependencySet) map[string]config.Dependency {
+	if depSet == nil {
+		return nil
+	}
+	if depSet.Dependencies == nil {
+		return nil
+	}
+	ds := make(map[string]config.Dependency, len(depSet.Dependencies))
+	for k, v := range depSet.Dependencies {
+		decimalChainId, err := common.HexToHash(k).Big().MarshalText()
```
geoknee, line 249: In the SCR we do not tend to work with hex string chain IDs but decimal ones, hence this conversion. But again the (current) type system doesn't really tell us what to do. That could be improved.

ops/internal/manage/staging.go
bitwiseguy: This will also be improved once we update the monorepo import since the incoming depset type uses eth.ChainId instead of string for chainIds

geoknee: Yep

ops/internal/manage/staging.go
```diff
@@ -215,3 +222,39 @@ func StagedSuperchainDefinition(rootP string) (*config.SuperchainDefinition, err
 
 	return sM, err
 }
+
+// isChainInDependencySet checks if a chain ID is in the supplied dependency set
+func isChainInDependencySet(chainID common.Hash, depSet *prestate.DependencySet) bool {
```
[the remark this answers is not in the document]
geoknee: Yeah another thing which will improve when we get to updating the monorepo import.

geoknee:
In favor of https://github.com/ethereum-optimism/superchain-registry/pull/1052

PR #1031, opened 2025-05-23
ops: support multiple op-deployer binaries

bitwiseguy:
# Overview
We need to be able to generate the genesis for multiple op-contracts releases. op-contracts are now mapped 1:1 to op-deployer tags. Therefore, this pr does the following:

1. Adds `versions.json` file mapping `op-contracts/vx.x.x` --> `op-deployer/vx.x.x`. All of these versions correspond to monorepo git tags.
2. Adds `build-deployer.sh` script that checks out a specified `op-deployer` tag of the monorepo and builds+caches the binary
3. Creates a go object `OpDeployer` which invokes the `build-deployer.sh` script, then can invoke the op-deployer binary to generate an l2 genesis.

Sequence to generate genesis:
```
deployer, _ := NewOpDeployer(lgr, l1ContractsRelease)
deployer.GenerateStandardGenesis(statePath)
```

# Related Issues
Closes https://github.com/ethereum-optimism/optimism/issues/15934


scripts/build-deployer.sh
```diff
@@ -0,0 +1,41 @@
...
+echo "Binary does not exist at $BINARY_PATH"
+echo "Building binary..."
+
+# Create temp directory
+TEMP_DIR=$(mktemp -d)
+trap "rm -rf $TEMP_DIR" EXIT
+
+# Efficient clone - only fetches the specific tag without history
+echo "Cloning monorepo tag $DEPLOYER_VERSION..."
+git clone --depth 1 --branch "$DEPLOYER_VERSION" https://github.com/ethereum-optimism/optimism.git "$TEMP_DIR"
```
geoknee, line 28: The [docs](https://devdocs.optimism.io/op-deployer/#portable) say that op-deployer is portable and easy to install.   I'm wondering if we can just pull a Docker image or something to make this a little more streamlined and looser coupled?

ops/internal/deployer/deployer.go
```diff
@@ -0,0 +1,161 @@
...
+type OpDeployer struct {
+	lgr                log.Logger
+	l1ContractsRelease string
+	inputStatePath     string
+	rootDir            string
+	deployerVersion    string
+}
+
+// NewOpDeployer creates a new OpDeployer instance.
+func NewOpDeployer(lgr log.Logger, l1ContractsRelease, stateFilepath, rootDir string) (*OpDeployer, error) {
```
geoknee, line 28: Could we infer `l1ContractsRelease` from the `stateFilePath`? As in, unmarshal the state.json as an opaque mapping and then lookup `l1ContractsLocator` property?

ops/internal/deployer/deployer.go
```diff
@@ -0,0 +1,161 @@
...
+	return nil
+}
+
+// GenerateGenesis runs op-deployer binary to generate a genesis
+func (d *OpDeployer) GenerateGenesis(workdir string) (*core.Genesis, error) {
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get home directory: %w", err)
+	}
+	deployerPath := filepath.Join(homeDir, ".cache/binaries", d.deployerVersion, "op-deployer")
```
geoknee, line 136: These few lines could be factored into a method `OpDeployer.getExePath()" so we can reuse them in other methods.

PR #1032, opened 2025-05-23
WIP: create_config uses deployer as a binary not as a library

geoknee:
This is very hacky right now, just trying to successfully use an OPD binary to get the genesis from a state.json file. 

The idea is to replace importing op-deployer structs and methods (for create_config and import_devnet) as library code with an extended version of what is being built here #1031 (where we shell out to a different op-deployer binary depending on what the chain was deployed with). We will need to rely on some parts of op-deployer being stable:

* the CLI interface, otherwise we need to track the invocations that we need to do over time as well, which will be painful.
* fields like `l1ContractsLocator` in the state.json file, which we need to use to infer which op-deployer binary to build and run. 

Other pain point are that the devnets we have deployed apparently don't use officially released op-deployer versions. Supporting arbitrary versions I think will be difficult so we should try and be stricter here in future. 

TODO

1. 
```
just create-config cupid ../devnets/betanets/cupid/op-deployer/state.json
```
Should work end to end, including:
- [ ] staging a compressed genesis file
- [ ] staging a chain config toml file

2. Same for `import_devnet` (although devnets generally do not launch off stable tags...)


ops/cmd/create_config/main.go
```diff
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/ethereum-optimism/optimism/op-deployer/pkg/deployer/state"
```
geoknee, line 7: RIP

ops/cmd/import_devnet/main.go
```diff
@@ -46,71 +40,72 @@ func main() {
 }
 
 func action(cliCtx *cli.Context) error {
-	wd, err := paths.FindRepoRoot()
-	if err != nil {
-		return fmt.Errorf("failed to get working directory: %w", err)
-	}
+	// 	wd, err := paths.FindRepoRoot()
```
geoknee, line 43: Just stubbed this entire thing out for now to prevent compiler errors.

ops/internal/deployer/deployer.go
```diff
@@ -45,8 +45,8 @@ func (d *OpDeployer) BuildBinary() (string, error) {
 	// 1. Remove tag:// prefix if present
 	// 2. Remove any -rc.X suffix for version matching
 	contractsKey := strings.TrimPrefix(d.l1ContractsRelease, "tag://")
-	rcSuffixRegex := regexp.MustCompile(`-rc\.[0-9]+$`)
-	contractsKey = rcSuffixRegex.ReplaceAllString(contractsKey, "")
+	// rcSuffixRegex := regexp.MustCompile(`-rc\.[0-9]+$`)
+	// contractsKey = rcSuffixRegex.ReplaceAllString(contractsKey, "")
```
geoknee, line 49: This is a hack I needed because some versions of op-deployer actually check for a tag which has the rc suffix.

ops/internal/deployer/state.go
```diff
@@ -260,3 +260,27 @@ func readL1ChainID(node *dasel.Node) (uint64, error) {
 	}
 	return uint64(l1ChainIDFloat), nil
 }
+
+func ReadL1ContractsRelease(node *dasel.Node) (string, error) {
+	l1ContractsReleaseNode, err := node.Query("appliedIntent.l1ContractsLocator")
```
geoknee, line 265: Ideally this is stable so we don't have to modify this function in future to keep up with OPD versions.

ops/internal/deployer/state.go
```diff
@@ -260,3 +260,27 @@ func readL1ChainID(node *dasel.Node) (uint64, error) {
...
+	}
+	l1ContractsRelease, ok := l1ContractsReleaseNode.InterfaceValue().(string)
+	if !ok {
+		return "", errors.New("failed to parse L1 contracts release")
+	}
+	return l1ContractsRelease, nil
+}
+
+func ReadL2ChainId(node *dasel.Node, idx int) (string, error) {
+	l2ChainIdNode, err := node.Query(fmt.Sprintf("appliedIntent.chains.[%d].id", idx))
```
geoknee, line 277: Again we want this to be stable. Not impossible to work around if e.g. `id` got changed to `chainId`... but still.

ops/internal/manage/generate.go
```diff
@@ -2,51 +2,84 @@ package manage
...
+	// }
+	// cfg.ShortName = shortName
+
+	// if name != nil {
+	// 	cfg.Name = *name
+	// }
+
+	// if superchain != nil {
+	// 	cfg.Superchain = config.Superchain(*superchain)
+	// }
```
geoknee, line 63: stubbed out for now, will circle back to this.

ops/internal/manage/generate.go
```diff
@@ -2,51 +2,84 @@ package manage
...
+
 	output.WriteOK("reading genesis")
-	genesis, _, err := inspect.GenesisAndRollup(&st, st.AppliedIntent.Chains[idx].ID)
+
+	genesis, err := opd.InspectGenesis(wd, chainId)
```
geoknee, line 67: Haven't yet got this fully working with any example state.json but I think I am close.

PR #1041, opened 2025-05-29
deployer: add OpDeployer for building, using binaries

bitwiseguy:
# Overview
Continues the buildout of `ops/internal/deployer` package. This pr splits off changes to that package from https://github.com/ethereum-optimism/superchain-registry/pull/1031, which then uses the `deployer` package for various superchain-registry actions.

Adds `OpDeployer` object, which can be instantiated such that it interfaces with an official `op-deployer` release. The appropriate `op-deployer` version is determined by a `versions.json` file, which maps `op-contracts` versions to the version of `op-deployer` that supports those contracts.

Usage:
```
opd, err := NewOpDeployer(lgr, l1ContractsRelease)
```

# Details
There is a new circleci job that builds all the op-deployer binaries in `versions.json`. The binaries are cached both within the workflow and across workflow runs. The binaries are then used by `OpDeployer` instances


ops/internal/deployer/deployer.go
```diff
@@ -0,0 +1,335 @@
...
+		d.lgr.Error(stderr.ReadString(0))
+		return nil, fmt.Errorf("failed to run op-deployer inspect genesis: %w", err)
+	}
+
+	var genesis OpaqueMapping
+	if err := json.Unmarshal(output, &genesis); err != nil {
+		return nil, fmt.Errorf("failed to parse op-deployer inspect genesis output: %w", err)
+	}
+
+	return &genesis, nil
```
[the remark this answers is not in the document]
bitwiseguy: Should we also return `OpaqueMapping` type for other methods (e.g. `InspectRollup`)? Or should we leave let them return their specific type since that's what we want in the other cases so that its easier to read nested fields

ops/internal/deployer/versions.json
```diff
@@ -0,0 +1,8 @@
+{
+  "op-contracts/v1.6.0": "op-deployer/v0.0.14",
+  "op-contracts/v1.7.0-beta.1+l2-contracts": "op-deployer/v0.0.14",
+  "op-contracts/v1.8.0": "op-deployer/v0.0.14",
+  "op-contracts/v2.0.0": "op-deployer/v0.2.2",
+  "op-contracts/v3.0.0": "op-deployer/v0.3.1",
+  "op-contracts/v4.0.0": "op-deployer/v0.4.0-rc.2"
+}
```
bitwiseguy, line 8: The `-rc.x` suffix for `op-contracts` is stripped before using this map so, e.g. any `op-contracts/v2.0.0*` will map to `op-deployer/v0.2.2`

ops/internal/deployer/deployer.go
```diff
@@ -0,0 +1,335 @@
...
+
+	deployerPath, err := d.getBinaryPath()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get deployer binary path: %w", err)
+	}
+
+	// We don't want to use a funded account here, because there should not be any txs sent.
+	// All contracts should have already been deployed, and the 'apply' command should skip
+	// those pipeline steps, then only generate the genesis.
+	privateKey, err := ecdsa.GenerateKey(crypto.S256(), rand.Reader)
```
mslipper, line 176: Just use the test... junk private key rather than a random one.

ops/internal/deployer/versions.json
```diff
@@ -0,0 +1,8 @@
+{
```
mslipper, line 1: [nit] This could just be a mapping in the Go code.

mslipper (commented):
last item is removing the op-deployer builds and using the binaries, and updating CI to download the binaries to a directory.

ops/internal/deployer/deployer.go
bitwiseguy: Done in https://github.com/ethereum-optimism/superchain-registry/pull/1041/commits/cd4a6eb6f4f47948ef16b0470750705218de75f7

ops/internal/deployer/scripts/build-binaries.sh
```diff
@@ -0,0 +1,220 @@
...
+  # Move to destination and make executable
+  echo "Installing binary to $binary_path"
+  mkdir -p "$(dirname "$binary_path")"
+  mv "$extracted_binary" "$binary_path"
+  chmod +x "$binary_path"
+
+  return 0
+}
+
+build_from_source() {
```
bitwiseguy, line 159: Needed this complexity because we won't always have a release to download from. For example, devnets will run from tags that aren't finalized releases. Therefore we need to build from source, which then adds some complexity using mise to install the correct go version before building the binary.

ops/internal/deployer/deployer.go
```diff
@@ -0,0 +1,308 @@
...
+	// 1. Remove tag:// prefix if present
+	// 2. Remove any -rc.X suffix for version matching
+	contractsKey := strings.TrimPrefix(d.l1ContractsRelease, "tag://")
+	rcSuffixRegex := regexp.MustCompile(`-rc\.[0-9]+$`)
+	contractsKey = rcSuffixRegex.ReplaceAllString(contractsKey, "")
+
+	// Read and parse versions.json
+	d.lgr.Info("Reading versions.json")
+	// Get the directory containing this file (deployer.go)
+	_, filename, _, _ := runtime.Caller(0)
```
mslipper, line 62: If we absolutely must use versions.json, then let's at least embed it into the binary using `goembed` and unmarshal it once rather than reading it at runtime.

ops/internal/deployer/deployer.go
```diff
@@ -0,0 +1,308 @@
...
+
+	deployerVersion, ok := versions[contractsKey]
+	if !ok {
+		return fmt.Errorf("no deployer version found for contracts: %s", contractsKey)
+	}
+	d.lgr.Info("Found deployer version", "version", deployerVersion)
+	d.DeployerVersion = deployerVersion
+
+	// Check if the op-deployer binary already exists
+	homeDir, err := os.UserHomeDir()
```
mslipper, line 84: Can we pass in the binary directory here as discussed rather than reading the user's home directory? This creates an implied API that we can make much clearer by passing in the directory on the command line.

ops/internal/deployer/deployer.go
```diff
@@ -0,0 +1,308 @@
...
+
+	// We don't want to use a funded account here, because there should not be any txs sent.
+	// All contracts should have already been deployed, and the 'apply' command should skip
+	// those pipeline steps, then only generate the genesis.
+	seed := []byte("seed phrase")
+	hash := crypto.Keccak256(seed)
+	privateKey, _ := crypto.ToECDSA(hash[:32]) // Use first 32 bytes as private key
+
+	privateKeyBytes := crypto.FromECDSA(privateKey)
+	privateKeyHex := "0x" + hex.EncodeToString(privateKeyBytes)
```
mslipper, line 169: Can we also use the private key from the`test... junk` mnemonic as discussed? The private key for that is `0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80`.

ops/internal/deployer/deployer.go
```diff
@@ -0,0 +1,308 @@
...
+	d.lgr.Info("Running `op-deployer inspect rollup`")
+
+	workdir, err := d.copyStateFileToTempDir(statePath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to copy state file to temporary directory: %w", err)
+	}
+	defer os.RemoveAll(workdir)
+
+	cmd := exec.Command(d.binaryPath, "inspect", "rollup", "--workdir", workdir, chainId)
+	stderr := bytes.Buffer{}
```
mslipper, line 267: [nit] for these buffers it's safer to do new(bytes.Buffer), since it's bad practice to copy the buffer by value

ops/internal/deployer/state.go
```diff
@@ -27,24 +34,66 @@ var standardV3State []byte
...
 		return nil, fmt.Errorf("failed to unmarshal JSON: %w", err)
 	}
 	return out, nil
 }
 
-func MergeStateV2(userState OpaqueMapping) (OpaqueMapping, OpaqueMapping, error) {
-	l1ChainID, err := readL1ChainID(dasel.New(userState))
+func getMergeStateFunc(version string) (func(state OpaqueState) (OpaqueMap, OpaqueState, error), error) {
+	// Extract the version number using regex
+	re := regexp.MustCompile(`op-deployer/v\d+\.(\d+)\.\d+`)
```
mslipper, line 53: Can we make this regexp reusable? We're doing version logic in two places.

ops/internal/deployer/state.go
```diff
@@ -43,8 +48,52 @@ func ReadOpaqueMappingFile(p string) (OpaqueMapping, error) {
 	return out, nil
 }
 
+func getMergeStateFunc(version string) (func(state OpaqueMapping) (OpaqueMapping, OpaqueMapping, error), error) {
```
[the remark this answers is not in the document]
mslipper: Did you actually do this?

ops/internal/deployer/state.go
```diff
@@ -27,24 +34,66 @@ var standardV3State []byte
...
+	var out OpaqueState
 	if err := json.NewDecoder(f).Decode(&out); err != nil {
 		return nil, fmt.Errorf("failed to unmarshal JSON: %w", err)
 	}
 	return out, nil
 }
 
-func MergeStateV2(userState OpaqueMapping) (OpaqueMapping, OpaqueMapping, error) {
-	l1ChainID, err := readL1ChainID(dasel.New(userState))
+func getMergeStateFunc(version string) (func(state OpaqueState) (OpaqueMap, OpaqueState, error), error) {
```
mslipper, line 51: Still need the statemerger type alias here.

ops/internal/deployer/opaque_map.go
```diff
@@ -0,0 +1,140 @@
...
+					val[i] = int64(fItem)
+				} else if mapItem, ok := item.(map[string]any); ok {
+					useInts(mapItem)
+				}
+			}
+		}
+	}
+}
+
+func (om OpaqueState) ReadL1ChainID() (uint64, error) {
```
mslipper, line 42: Can we make type aliases for these like we discussed?

ops/internal/deployer/opaque_map.go
```diff
@@ -0,0 +1,140 @@
...
+		return common.HexToAddress(systemConfigProxy), nil
+	}
+
+	// Fallback to the legacy field name
+	systemConfigProxyNode, err = node.Query(fmt.Sprintf("appliedIntent.opChainDeployments.[%d].systemConfigProxyAddress", idx))
+	if err != nil {
+		return common.Address{}, fmt.Errorf("failed to read SystemConfigProxy: %w", err)
+	}
+
+	systemConfigProxy, ok := systemConfigProxyNode.InterfaceValue().(string)
```
mslipper, line 111: [nit] we may want a utility method that gets a string from a node, since we're duplicating a lot of code here to get strings out

ops/internal/deployer/deployer_test.go
```diff
@@ -0,0 +1,76 @@
...
+				require.NoError(t, err)
+				require.NotNil(t, deployer)
+				require.Equal(t, tt.l1ContractsRelease, deployer.l1ContractsRelease)
+				require.NotEmpty(t, deployer.DeployerVersion)
+
+				// Verify the binary exists
+				homeDir, err := os.UserHomeDir()
+				require.NoError(t, err)
+
+				binaryPath := filepath.Join(homeDir, ".cache", deployer.DeployerVersion, "op-deployer")
```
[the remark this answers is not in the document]
bitwiseguy: Yes the `deployer/scripts/build-binaries.sh` script is expected to build all of the binaries in ci prior to any of the go tests. You should be able to run the same script locally if you want to run these tests locally too

geoknee:
I made a PR with responses to @mslipper's review here https://github.com/ethereum-optimism/superchain-registry/pull/1042

ops/internal/deployer/deployer.go
bitwiseguy: Done in https://github.com/ethereum-optimism/superchain-registry/pull/1041/commits/1fba7e052bf9101c629c66b62fe5ee4e66b0c02d

ops/internal/deployer/deployer.go
bitwiseguy: Done in https://github.com/ethereum-optimism/superchain-registry/pull/1041/commits/1fba7e052bf9101c629c66b62fe5ee4e66b0c02d

ops/internal/deployer/state.go
bitwiseguy: Done in https://github.com/ethereum-optimism/superchain-registry/pull/1041/commits/1fba7e052bf9101c629c66b62fe5ee4e66b0c02d

ops/internal/deployer/opaque_map.go
bitwiseguy: Done in https://github.com/ethereum-optimism/superchain-registry/pull/1041/commits/1fba7e052bf9101c629c66b62fe5ee4e66b0c02d

ops/internal/deployer/deployer.go
bitwiseguy: Done in https://github.com/ethereum-optimism/superchain-registry/pull/1041/commits/1fba7e052bf9101c629c66b62fe5ee4e66b0c02d

ops/internal/deployer/deployer.go
mslipper: It will eventually be CLI invoked by other tooling, so what I'm suggesting we do here is update the deployer library to consume the binary directory as an arg so that we can pass it in via CLI later. The CI job that runs the report would specify the command line.

mslipper (approved):

PR #1043, opened 2025-05-30
ops/import-devnet: use op-deployer binaries instead of library

geoknee:
Devnets for which I have had this work:
- [x] eris (needed to allow the user to override the op-deployer version)
- [ ] interop-rc-alpha (deployed using op-deployer built from a custom commit, not a tag, ~we should allow to build arbitrary versions from source~ we will not support that)
- [x] cupid

---
TODO 
- [x] resolve TODOs
- [ ] tests. I think we can go through a bunch of devnets (and maybe a single chain as well) and check we can run the import and sync commands end to end with no errors. Easy enough to do manually but can also be automated. Marking this PR ready for initial review while we work on that.


ops/internal/deployer/versions.json
```diff
@@ -4,5 +4,6 @@
   "op-contracts/v1.8.0": "op-deployer/v0.0.14",
   "op-contracts/v2.0.0": "op-deployer/v0.2.2",
   "op-contracts/v3.0.0": "op-deployer/v0.3.1",
-  "op-contracts/v4.0.0": "op-deployer/v0.4.0-rc.2"
+  "op-contracts/v4.0.0": "op-deployer/v0.4.0-rc.2",
+  "interop-rc-alpha": "f79ed8b9c9cbdbf8bb492074f3f98da7f072e21a"
```
geoknee, line 8: This is abusing the semantic meaning of this file somewhat. 

Other options:
1. Accept a list of additional git references (e.g. commits) to the `build-binaries.sh` script instead of declaring them here
2. Require devnets to always be deployed from a tagged version of op-deployer

geoknee:
Question:

Can we simplify `queryInt` by assuming there are no floats (that is no numbers with a nonzero bit after the decimal).

ops/internal/manage/generate.go
```diff
@@ -32,21 +51,55 @@ func GenerateChainArtifacts(st state.State, wd string, shortName string, name *s
...
+	}
+
+	// Step 2: Unmarshal the JSON data into a core.Genesis struct
+	var genesis core.Genesis
+	if err := json.Unmarshal(jsonData, &genesis); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal JSON to Genesis: %w", err)
+	}
+
+	return &genesis, nil
+}
```
geoknee, line 105: This is not always safe: it is possible to get an `opaqueMap` from a _future_ version of op-deployer (say with a new hardfork time in it), and then we would lose information when converting to the `core.Genesis` if we are pinned to an older version of op-geth in the superchain registry.

If possible, we should work entirely with `opaqueMap`.

ops/internal/manage/staging.go
```diff
@@ -91,29 +108,27 @@ func InflateChainConfig(st *state.State, idx int) (*config.StagedChain, error) {
...
-		UnsafeBlockSigner: config.NewChecksummedAddress(chainIntent.Roles.UnsafeBlockSigner),
-		BatchSubmitter:    config.NewChecksummedAddress(chainIntent.Roles.Batcher),
-		Challenger:        config.NewChecksummedAddress(chainIntent.Roles.Challenger),
+		SystemConfigOwner: config.NewChecksummedAddress(dc.FinalSystemOwner),
+		ProxyAdminOwner:   config.NewChecksummedAddress(dc.ProxyAdminOwner),
+		Guardian:          config.NewChecksummedAddress(dc.SuperchainConfigGuardian),
+		Proposer:          config.NewChecksummedAddress(dc.L2OutputOracleProposer),
+		UnsafeBlockSigner: config.NewChecksummedAddress(dc.P2PSequencerAddress),
+		BatchSubmitter:    config.NewChecksummedAddress(dc.BatchSenderAddress),
+		// Challenger:        config.NewChecksummedAddress(chainIntent.Roles.Challenger), // TODO
```
geoknee, line 117: Let's get these with opaque mapping queries / getters.

ops/internal/deployer/versions.json
bitwiseguy: I think if we need to support any commit, we should change the overall format of this file or add an additional file. The mixing of `op-contracts version` and `devnet name` --> `op-deployer version` is awkward

geoknee: Have reverted this change.

ops/internal/manage/generate.go
geoknee: Added the logic to do this, plus a test.

ops/internal/manage/staging.go
bitwiseguy: Done in https://github.com/ethereum-optimism/superchain-registry/pull/1043/commits/0064ce7a22a3c400a7fb6db1f97db15eb53c1db3

ops/cmd/import_devnet/main.go
```diff
@@ -72,36 +73,42 @@ func action(cliCtx *cli.Context) error {
...
+	st, err := deployer.ReadOpaqueStateFile(statePath)
+	if err != nil {
+		return fmt.Errorf("failed to read opaque state file: %w", err)
+	}
+
+	numChains, err := st.GetNumChains()
+	if err != nil {
+		return fmt.Errorf("failed to read number of chains: %w", err)
+	}
+	if numChains != len(m.L2.Chains) {
```
mslipper, line 85: there should only ever be one chain in the manifest file or state file - can we chack for that here?

ops/cmd/sync_staging/main.go
```diff
@@ -138,14 +154,20 @@ func action(cliCtx *cli.Context) error {
...
+		if !preserveInput {
 			cfgFilename := path.Join(stagingDir, chainCfg.ShortName+".toml")
 			if err := os.Remove(cfgFilename); err != nil {
 				output.WriteNotOK("failed to remove %s: %v", cfgFilename, err)
 			}
 			if err := os.Remove(genesisFilename); err != nil {
 				output.WriteNotOK("failed to remove %s: %v", genesisFilename, err)
 			}
+			stateFilename := path.Join(stagingDir, "state.json")
+			if _, err := os.Stat(stateFilename); err == nil {
```
mslipper, line 166: why do you need this stat, and to remove the state?

ops/internal/deployer/deployer.go
```diff
@@ -72,31 +72,35 @@ func NewOpDeployer(lgr log.Logger, l1ContractsRelease string, binDir string) (*O
...
-	rcSuffixRegex := regexp.MustCompile(`-rc\.[0-9]+$`)
-	contractsKey = rcSuffixRegex.ReplaceAllString(contractsKey, "")
-
-	// Look up deployer version in the embedded map
-	deployerVersion, ok := contractVersions[contractsKey]
-	if !ok {
-		return fmt.Errorf("no deployer version found for contracts: %s", contractsKey)
+func (d *OpDeployer) checkBinary(binDir string, opDeployerVersion string) error {
+	if opDeployerVersion != "" {
+		d.DeployerVersion = opDeployerVersion
```
mslipper, line 77: early return would be simpler here and reduce nesting

ops/internal/deployer/deployer.go
```diff
@@ -63,7 +63,7 @@ func NewOpDeployer(lgr log.Logger, l1ContractsRelease string, binDir string) (*O
 		l1ContractsRelease: l1ContractsRelease,
 	}
 
-	err := opd.checkBinary(binDir)
+	err := opd.checkBinary(binDir, opDeployerVersion)
```
mslipper, line 66: This seems like kind of weird to me. Why can't the l1 contracts release be auto-detected in this case? Why are we passing in the deployer version like this? If we need an override or something it'd be cleaner to extract the deployer bin location out of the opdeployer struct and into some standalone util that gets passed into the struct's constructor.

ops/cmd/import_devnet/main.go
bitwiseguy: In the case of interop devnets, there are always at least two chains in the state file (e.g. [interop-rc-alpha](https://github.com/ethereum-optimism/devnets/blob/main/alphanets/interop-rc-alpha/op-deployer/state.json)). We know the incoming devnet we want to add to the registry will have interop enabled, so we're trying to optimize for that known use-case

ops/cmd/sync_staging/main.go
bitwiseguy: The workflow as implemented is:
1. `import_devnet` or `create_config` command copies external state.json file to the staging directory (.staging/state.json). These same commands also generate chain config (.staging/<shortname>.toml) and genesis files (.staging/<shortname>.json.zst) based on the state.json
3. `print_staging_report` reads from the state.json file to generate validation reports for new chains in pull requests
4. `sync_staging` processes the files by:
   a. validating them
   b. moving them to their final locations in the superchain registry
   c. cleaning up the original files from the staging directory (unless --preserve-input is specified)
5. ci checks verify that the staging directory is empty after processing

So [4] cleans up this file if it exists, as well as other artifacts from previous commands.

ops/internal/deployer/deployer.go
bitwiseguy: We're trying to cover some edge cases for `import_devnet` to handle like [this](https://github.com/ethereum-optimism/devnets/blob/main/alphanets/interop-rc-alpha/op-deployer/state.json#L14-L15) where the contracts locators are long urls, which we don't currently know how to encode and map to op-deployer versions

ops/cmd/import_devnet/main.go
mslipper: ah good point

mslipper (approved):
This unblocks us, so I’m going to go merge this keep momentum. I’ll do a quick follow-up pass tonight to clean up some of the structure and naming to make it easier to build on. I'm concerned about a few of the architectural choices here. Will provide some examples in the follow up PRs.

ops/internal/manage/generate.go
mslipper: we should always use opaquemap

PR #1048, opened 2025-05-30
ops/deployer: add `TestBinaryInvocation` and run in CI

### LINKED ISSUE #1044: ops: add a test to check that we can invoke `op-deployer` as a binary
Make the test pass in CI by:
* creating a temporary directory for the op-deployer binary
* have the `deployer` package read that in as an env var
* have the test invoke the binary and check that it responds as expected

geoknee:
Closes #1044

### DIFF
--- a/ops/internal/deployer/deployer.go
+++ b/ops/internal/deployer/deployer.go
@@ -23,10 +23,15 @@ var versionsJSON []byte
 // contractVersions maps contract versions to deployer versions
 var contractVersions map[string]string
 
+// cacheDirEnvVar is the environment variable that contains the path to the op-deployer cache directory
+var cacheDirEnvVar = "DEPLOYER_CACHE_DIR"
+var CACHE_DIR string
+
 func init() {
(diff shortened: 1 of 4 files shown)


ops/internal/deployer/deployer.go
```diff
@@ -23,10 +23,17 @@ var versionsJSON []byte
 // contractVersions maps contract versions to deployer versions
 var contractVersions map[string]string
 
+// cacheDirEnvVar is the environment variable that contains the path to the op-deployer cache directory
+var (
+	cacheDirEnvVar = "DEPLOYER_CACHE_DIR"
+	CACHE_DIR      string
```
mslipper, line 29: [nit] since this isn't actually a constant, can we make this `StudCaps`?

mslipper (approved):
This looks good to me. When we integrate this with the reporting function, we should refactor the deployer.go implementation to move the env var/cache directory stuff into the CLI flags for the report tool.

PR #1053, opened 2025-06-03
Clean up deployer lib implementation

mslipper:
Perform clean ups as described in https://github.com/ethereum-optimism/superchain-registry/pull/1043#pullrequestreview-2890420471. Specific refactors are called out as code review comments.

### DIFF
--- a/ops/cmd/create_config/main.go
+++ b/ops/cmd/create_config/main.go
@@ -40,6 +40,7 @@ func main() {
 		Flags: []cli.Flag{
 			StateFilename,
 			Shortname,
+			OpDeployerBinDir,
 		},
 		Action: action,
 	}
--- a/ops/cmd/create_config/main.go
+++ b/ops/cmd/create_config/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 
 	"github.com/ethereum-optimism/superchain-registry/ops/internal/manage"
 	"github.com/ethereum-optimism/superchain-registry/ops/internal/output"
@@ -22,6 +23,13 @@ var (
 		Usage:    "Shortname of the chain.",
 		Required: true,
 		Value:    "newchain",
+	}
+	OpDeployerBinDir = &cli.StringFlag{
+		Name:     "op-deployer-bin-dir",
+		Usage:    "Path to the directory containing op-deployer binaries.",
+		Required: true,
+		EnvVars:  []string{"DEPLOYER_CACHE_DIR"},
+		Value:    defaultBinDir(),
 	}
 )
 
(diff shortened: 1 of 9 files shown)


ops/internal/deployer/deployer.go
```diff
@@ -23,88 +23,85 @@ var versionsJSON []byte
 // contractVersions maps contract versions to deployer versions
 var contractVersions map[string]string
 
-// cacheDirEnvVar is the environment variable that contains the path to the op-deployer cache directory
-var (
-	cacheDirEnvVar = "DEPLOYER_CACHE_DIR"
-	CacheDir       string
-)
+type BinaryPicker interface {
```
mslipper, line 26: For an op deployer binary to work, you need to provide both the path to the binary as well as how to merge state. (We should probably separate these concerns in the future, but this will do for now.) Passing them in as separate arguments to the op deployer lib as we were before was very confusing. For example, if you passed in an empty string for the deployer version, it would auto detect the right binary and merge function from the l1 contracts release. However, if you specified a deployer version it would set the binary path directly and ignore any value provided for the l1 contracts release.

Instead, I have created an interface to represent the info the deployer lib needs to pick a binary and a state merger. Then, I pulled the logic that autodetects these based on the release out into its own utility method that returns the interface. Logic that branches based on the value of the op-deployer version is further extracted out of this package altogether.

ops/internal/manage/generate.go
```diff
@@ -21,22 +21,51 @@ import (
...
 	st, err := deployer.ReadOpaqueStateFile(statePath)
 	if err != nil {
 		return fmt.Errorf("failed to read opaque state file: %w", err)
 	}
-	l1contractsrelease, err := st.ReadL1ContractsLocator()
-	if err != nil {
-		return fmt.Errorf("failed to read L1 contracts release: %w", err)
+
+	var picker deployer.BinaryPicker
+	if opDeployerVersion == "" {
```
mslipper, line 40: The logic around what to do if the op deployer version is blank is very much a concern of the manage package, not the deployer package. As such I've pulled that logic in here.

ops/cmd/create_config/main.go
```diff
@@ -48,12 +56,22 @@ func action(cliCtx *cli.Context) error {
...
+	err = manage.GenerateChainArtifacts(statePath, wd, cliCtx.String(Shortname.Name), nil, nil, 0, "", opDeployerBinDir)
 	if err != nil {
 		return fmt.Errorf("failed to generate chain config: %w", err)
 	}
 
 	output.WriteOK("done")
 	return nil
 }
+
+func defaultBinDir() string {
```
mslipper, line 70: There's some code duplication here with the import_devnet package, but I'm OK with it for the time being.

ops/internal/manage/generate.go
```diff
@@ -21,22 +21,51 @@ import (
...
+		// If no op-deployer version is specified, use the version specified in the state file.
+		// The library will autodetect the correct binary.
```
geoknee, line 42:
```suggestion
		// If no op-deployer version is specified, an appropriate version will be
		// inferred from the state file.
```

bitwiseguy (approved):

geoknee (approved):

ops/internal/deployer/state.go
```diff
@@ -48,9 +48,9 @@ func ReadOpaqueStateFile(p string) (OpaqueState, error) {
 	return out, nil
 }
 
-type stateMerger = func(state OpaqueState) (OpaqueMap, OpaqueState, error)
+type StateMerger = func(state OpaqueState) (OpaqueMap, OpaqueState, error)
 
-func getMergeStateFunc(version string) (stateMerger, error) {
+func GetStateMerger(version string) (StateMerger, error) {
```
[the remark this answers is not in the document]
bitwiseguy: For now we expect to only use tagged version of op-deployer and op-contracts. But it would be nice to add this flexibility in a follow-up because its likely we will have cases where we want to add non tagged versions

PR #1065, opened 2025-06-04
depsets: onchain checks dependent on interop_time

bitwiseguy:
Modifies the depset checker based on recent tests of the devnet rehearsal:
1. Allows chain configs to have a depset even if they don't have an `interop_time` set yet
2. Only performs onchain checks for chains whose `interop_time` < the current time

So the new validations will be:
1. offchain: performed for all chains who have a depset in their config
2. onchain: only performed if the chain also has an `interop_time` < current time

### DIFF
--- a/ops/internal/manage/depsets.go
+++ b/ops/internal/manage/depsets.go
@@ -154,23 +154,23 @@ func (dc *DepsetChecker) checkOnchain(cfgs []DiskChainConfig) error {
 
 	// Find chains that have already activated interop
 	now := time.Now().Unix()
-	var validChains []DiskChainConfig
+	var activatedChains []DiskChainConfig
 	for _, cfg := range cfgs {
 		interopTime := cfg.Config.Hardforks.InteropTime
 		if interopTime == nil || *interopTime.U64Ptr() > uint64(now) {
 			continue
 		}
-		validChains = append(validChains, cfg)
+		activatedChains = append(activatedChains, cfg)
 	}
 
 	// If we have fewer than 2 activated chains, no comparison needed
-	if len(validChains) < 2 {
+	if len(activatedChains) < 2 {
 		dc.lgr.Info("skipping onchain checks for depset")
 		return nil
 	}
 
 	// Get and validate the first valid chain's addresses
-	firstAddrs, err := dc.getAndValidateAddresses(validChains[0].Config.ChainID)
+	firstAddrs, err := dc.getAndValidateAddresses(activatedChains[0].Config.ChainID)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


ops/internal/manage/depsets.go
```diff
@@ -146,7 +141,8 @@ func (dc *DepsetChecker) checkOffchain(cfgs []DiskChainConfig) error {
 }
 
 // checkOnchain ensures that DisputeGameFactoryProxy and EthLockboxProxy addresses read from onchain
-// are the same for all chains in a depset
+// are the same for all chains in a depset. These checks are only performed for chains who have
+// already activated interop (i.e. interop_time < current unix timestamp).
```
geoknee, line 145: iiuc there is a period of time (I'm not sure for how long) where we would expect chains in a depset to share these proxy addresses _before_ the hf activation. 

I just want to check that we are happy to not be able to detect if the proxy addresses aren't in fact shared during that time?

It feels like this is not a check on the depset per se, but a check on some L1 contracts' property of chains in a given depset. I am wondering if it should be split out from the consistency check on the depset itself?

ops/internal/manage/depsets.go
```diff
@@ -156,8 +152,25 @@ func (dc *DepsetChecker) checkOnchain(cfgs []DiskChainConfig) error {
 		return nil
 	}
 
-	// Get and validate the first chain's addresses
-	firstAddrs, err := dc.getAndValidateAddresses(cfgs[0].Config.ChainID)
+	// Find chains that have already activated interop
+	now := time.Now().Unix()
+	var validChains []DiskChainConfig
```
geoknee, line 157: Is there a better name than `validChains`? I don't think it is invalid to not have interop activated yet.

ops/internal/manage/depsets.go
```diff
@@ -167,8 +180,9 @@ func (dc *DepsetChecker) checkOnchain(cfgs []DiskChainConfig) error {
 	// issue: https://github.com/ethereum-optimism/optimism/issues/16058
```
geoknee, line 180: Just a reminder that we are unblocked on this now.

geoknee (approved):

ops/internal/manage/depsets.go
bitwiseguy: Yes I'm planning to address that in a follow-up. We will need to update the monorepo import to pull in the latest op-fetcher (which now queries for the EthLockboxProxy address), then we can re-enabled this code

ops/internal/manage/depsets.go
bitwiseguy: This is the onchain param we can query [want](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/L1/OptimismPortal2.sol#L126) to determine if `opcm.migrate` has been called for a specific chain.

Note that opcm.migrate may be called twice (may change later); when an already migrated 1-interop chain is being added to another interop set. But as soon as that boolean is set, you can be sure that the chain is using super roots.

So we should implement something like this for the onchain checks:
```
if optimismPortal.superRootsActive && len(depset) > 1 {
  // loop through depset and validate all members have same DisputeGameFactoryProxy and same EthLockboxProxy
}
```

bitwiseguy: Added an issue to capture this work https://github.com/ethereum-optimism/superchain-registry/issues/1069

PR #1070, opened 2025-06-05
devnet: import rehearsal-0-bn

bitwiseguy:
Adds the `rehearsal-0-bn` superchain and its two chains to the registry. Ran the following commands locally to create these artifacts:
```
just import-devnet $HOME/testfiles/state.json $HOME/testfiles/manifest.yaml
just sync-staging --l1-rpc-urls=$SEPOLIA_RPC_URL
```

### DIFF
--- a/superchain/configs/rehearsal-0-bn/rehearsal-0-bn-0.toml
+++ b/superchain/configs/rehearsal-0-bn/rehearsal-0-bn-0.toml
@@ -20,6 +20,7 @@ max_sequencer_drift = 600
   granite_time = 0 # Thu 1 Jan 1970 00:00:00 UTC
   holocene_time = 0 # Thu 1 Jan 1970 00:00:00 UTC
   isthmus_time = 0 # Thu 1 Jan 1970 00:00:00 UTC
+  interop_time = 1749150000 # Thu 5 Jun 2025 19:00:00 UTC
 
 [interop]
   [interop.dependencies]
--- a/superchain/configs/rehearsal-0-bn/rehearsal-0-bn-1.toml
+++ b/superchain/configs/rehearsal-0-bn/rehearsal-0-bn-1.toml
@@ -20,6 +20,7 @@ max_sequencer_drift = 600
   granite_time = 0 # Thu 1 Jan 1970 00:00:00 UTC
   holocene_time = 0 # Thu 1 Jan 1970 00:00:00 UTC
   isthmus_time = 0 # Thu 1 Jan 1970 00:00:00 UTC
+  interop_time = 1749150000 # Thu 5 Jun 2025 19:00:00 UTC
 
 [interop]
   [interop.dependencies]
--- a/superchain/configs/rehearsal-0-bn/superchain.toml
+++ b/superchain/configs/rehearsal-0-bn/superchain.toml
@@ -4,6 +4,7 @@ superchain_config_addr = "0x672F11f34b7bE67C13A6FCba819c339bc3B0a585"
 op_contracts_manager_addr = "0x0fcD165e2531C934a186e3a412D58Daa241E0Ad0"
 
 [hardforks]
+  interop_time = 1749150000 # Thu 5 Jun 2025 19:00:00 UTC
 
 [l1]
   chain_id = 11155111


superchain/configs/rehearsal-0-bn/rehearsal-0-bn-0.toml
```diff
@@ -0,0 +1,70 @@
...
+max_sequencer_drift = 600
+
+[hardforks]
+  canyon_time = 0 # Thu 1 Jan 1970 00:00:00 UTC
+  delta_time = 0 # Thu 1 Jan 1970 00:00:00 UTC
+  ecotone_time = 0 # Thu 1 Jan 1970 00:00:00 UTC
+  fjord_time = 0 # Thu 1 Jan 1970 00:00:00 UTC
+  granite_time = 0 # Thu 1 Jan 1970 00:00:00 UTC
+  holocene_time = 0 # Thu 1 Jan 1970 00:00:00 UTC
+  isthmus_time = 0 # Thu 1 Jan 1970 00:00:00 UTC
```
bitwiseguy, line 22: Should we manually add the `interop_time` here?

superchain/configs/rehearsal-0-bn/superchain.toml
```diff
@@ -0,0 +1,11 @@
+name = "rehearsal-0-bn"
+protocol_versions_addr = "0x72B9c5a159B0dEE8331a268791aa432619693c06"
+superchain_config_addr = "0x672F11f34b7bE67C13A6FCba819c339bc3B0a585"
+op_contracts_manager_addr = "0x0fcD165e2531C934a186e3a412D58Daa241E0Ad0"
+
+[hardforks]
+
+[l1]
+  chain_id = 11155111
+  public_rpc = "https://ci-sepolia-l1-archive.optimism.io"
```
bitwiseguy, line 10: This just uses the rpc url provided to the following command:
```
just sync-staging --l1-rpc-url=$SEPOLIA_RPC_URL
```
Do we want a different one here?

superchain/configs/rehearsal-0-bn/rehearsal-0-bn-1.toml
```diff
@@ -0,0 +1,70 @@
...
+  L1StandardBridgeProxy = "0xA2221fDE8C41b4b0802F5eF991CbF4feb75E8081"
+  OptimismMintableERC20FactoryProxy = "0x28fBc3FaDb3862F7c149A20E09c19FF8EbD0C8Ad"
+  OptimismPortalProxy = "0x1FD346D7e2394fbcC31501cbE88Df89D6Ddb0a0e"
+  SystemConfigProxy = "0x08712404111E4b5B4504e67BEF33205608791Ef7"
+  ProxyAdmin = "0xC302B8Fe8970578ca8E656800Dc302EE90558Fb9"
+  SuperchainConfig = "0x672F11f34b7bE67C13A6FCba819c339bc3B0a585"
+  AnchorStateRegistryProxy = "0x2F4C2F691d2A128CA4238b921DF802C8f02422A1"
+  DelayedWETHProxy = "0x08b5aBC848C8C641eB7a3927ddae11295d2914cd"
+  DisputeGameFactoryProxy = "0xB76D4C3acD6c750A1986120aAE1379Db8bf46B89"
+  PermissionedDisputeGame = "0xe517036541aDa792FDC47edd638028489e9A6222"
```
bitwiseguy, line 70: These are the addresses as set in the state.json. Some are already stale (e.g. DisputeGameFactoryProxy) due to reconfigurations made after deployment. The latest addresses pulled from onchain are in `addresses.json`

superchain/configs/rehearsal-0-bn/superchain.toml
geoknee: I think this url is validated to have the correct chain id, at least. Doesn't seem problematic to me for devnets to have this. wdyt?

superchain/configs/rehearsal-0-bn/rehearsal-0-bn-1.toml
geoknee: Let's get on and rip these addresses out of the individual chain cofigs soon. Do we have an issue tracking that?

superchain/configs/rehearsal-0-bn/superchain.toml
bitwiseguy: Yes it is validated in that way. Seems fine to me as well so will leave

superchain/configs/rehearsal-0-bn/rehearsal-0-bn-0.toml
bitwiseguy: I went ahead and added the timestamp to superchain.toml and then ran `just apply-hardforks`. I think we want that timestamp in the configs so that it will be included when we update op-geth's import of superchain-registry to include these configs. Then we can test op-nodes running with this timestamp

geoknee: We may want to cover this in the readme instructions for `import_devnet`.

geoknee (approved):

PR #1071, opened 2025-06-05
depsets: re-enable EthLockboxProxy consistency validation

bitwiseguy:
# Overview
This pr updates the monorepo import so that we pull in the latest op-fetcher which queries for the EthLockboxProxy address. op-fetcher is then used to update the codegen files. This allows us to re-enable the depset validation for EthLockboxProxy consistency.

# Related Issues
Closes https://github.com/ethereum-optimism/optimism/issues/16058


chainList.json
```diff
@@ -903,7 +903,7 @@
       "chain": "sepolia"
     },
     "faultProofs": {
-      "status": "none"
```
geoknee, line 906: Is this an expected change? What explains it?

geoknee (approved):
Code changes look good, but it would be nice to record the reason for some of the address updates on this PR (for posterity).

PR #1077, opened 2025-06-13
Update addresses after Metal Mode FP upgrade execution

Wazabie:
Ran just codegen to update addresses in the SCR


sbvegan (approved):
lgtm and til that boba sepolia is also running the fp system and has rotated their keys

PR #1127, opened 2025-10-24
jovian: add bounds for minimum base fee in standard config

geoknee:
This change forms part of a [proposal put to Optimism Governance](https://gov.optimism.io/t/upgrade-17-proposal-jovian-hardfork/10374), so should only be merged after it is approved by gov.

Summary:

Minimum base fee: no minimum, but standard chains must not set it higher than 10 GWei.
DA Footprint Gas Scalar: no restrictions


bitwiseguy (commented):
Seems like we should add these same bounds to the sepolia standard-config-params files too

PR #1143, opened 2025-11-04
chore(daily-codegen): update PR message.

scharissis:
Add information about what files we should check.


falcorocks (approved):

scharissis:
> A link to the runbook is already included in the slack message. The runbook has additional context, including links to all codegen files. I would prefer to update the runbook if needed instead of adding redundant info to the slack message

This PR doesn't update the Slack message. It updates the PR messages. Specifically it adds:
1. What files to review
2. A link to the runbook
As an (oncall) reviewer, I felt this information would be a useful addition.

Do you still think this is not useful?

bitwiseguy (approved):

PR #1155, opened 2025-11-14
chore: removing snax chain

sbvegan:
## description

Snax Chain has been sunset, see emiliano's comment here: https://github.com/ethereum-optimism/superchain-registry/pull/1151


bitwiseguy (commented):
To remove a chain from the registry I believe we should do the following:

1. remove chain config toml ([here](https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/configs/mainnet/snax.toml))
2. removed compressed genesis file ([here](https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/extra/genesis/mainnet/snax.json.zst))
3. run codegen, which should then remove entries from the following codegen files:
    * chainList.json
    * chainList.toml
    * CHAINS.md
    * addresses.json

Also not sure why some addresses.json values are changing as part of your current pr. Might need to rebase on top of latest main commit

sbvegan:
ah man something funky happened when I merged main, I'll take a look

PR #1198, opened 2026-02-12
[Celo Sepolia] add isthmus activation

palango:
Adds Celo Sepolia Isthmus activation to the registry


bitwiseguy:
@palango - I don't think you want to retroactively activate jovian at genesis. That will cause issues for celo sepolia

palango:
@bitwiseguy Not adding it was an oversight from our side. It's in the official genesis files as well: https://storage.googleapis.com/cel2-rollup-files/celo-sepolia/genesis.json

PR #1201, opened 2026-03-13
Expand check-codegen to catch Go code/dep changes

stevennevins:
## Problem

The daily-codegen CI broke for 5+ weeks undetected because check-codegen only validates when .toml config files change. When Go code or dependencies change (e.g., stale op-fetcher), the job skips validation entirely.

## Solution

Expand check-codegen to run a smoke test when Go files in ops/ change. If no .toml files changed but Go code did, run codegen for a single chain to verify the binary works.

## Changes

- Detect Go code/dependency changes in the check-codegen job
- Run a smoke test for OP Mainnet (chain 10) when Go changes are detected
- Skip validation only when neither .toml nor Go files changed

This prevents silent failures when dependencies become stale without config updates.


bitwiseguy:
Replaced by https://github.com/ethereum-optimism/superchain-registry/pull/1211

PR #1212, opened 2026-04-27
superchain: remove base-mainnet|sepolia|sepolia-dev-0

bitwiseguy:
Used `just remove-chain <chain-id>` 3x, once for each chain


sbvegan (approved):
lgtm

PR #1259, opened 2026-06-18
ops: add keep_karst_upgrade_gas hardfork flag

sebastianst:
## Summary

Adds a per-chain config flag `keep_karst_upgrade_gas` (default `false`) and sets it to `true` for the **sepolia** and **mainnet** superchains.

## Background

With Karst the OP Stack introduced "upgrade gas": extra gas added to a NUT-bundle fork's **activation block** gas limit so the network-upgrade transactions don't have to fit within the normal system-tx budget. It is meant to apply **only** to the activation block.

A bug in op-node and kona made that upgrade gas **stick** to every block after the activation block — the inflated gas limit is reconstructed from the parent header and re-applied indefinitely. The fix (ethereum-optimism/optimism#21441) subtracts the upgrade gas again at the block right after the activation block, reverting to the steady-state gas limit.

Chains that **already activated Karst with the leak baked into their history** can't take that subtraction retroactively — it would change their historical gas limits and break validation. For those chains, `keep_karst_upgrade_gas = true` opts out of the **Karst** subtraction only: the inflated limit is kept (history still validates), and the operator clears it themselves with a `setGasLimit` whenever they choose. Later NUT-bundle forks have no opt-out (they aren't live yet, so the fix applies cleanly from the start).

The sepolia and mainnet superchains have Karst scheduled, so they get `keep_karst_upgrade_gas = true`.

## Changes

- `ops/internal/config/chain.go`: add `KeepKarstUpgradeGas bool` to the `Hardforks` struct, ordered between Karst and Lagoon to match the canonical fork sequence.
- `ops/internal/config/hardforks.go`: `CopyHardforks` propagates the flag from the superchain definition down to each chain. The reflection-based activation-time copy loop now skips non-pointer fields, so the new bool no longer trips it.
- `superchain/configs/{sepolia,mainnet}/superchain.toml`: set `keep_karst_upgrade_gas = true`.
- Regenerated the affected per-chain configs via `just apply-hardforks`.

## Testing

- `go test ./ops/internal/config/...` — green; new test covers `CopyHardforks` propagating the bool to chains.

---

Paired with ethereum-optimism/optimism#21441 (the op-node + kona fix that reads this flag).

🤖 Generated by Claude Code

### DIFF
--- a/superchain/README.md
+++ b/superchain/README.md
@@ -1,3 +1,30 @@
+## Hardforks
+
+Each chain config (`superchain/configs/<superchain>/<chain>.toml`) carries a `[hardforks]` table with the activation
+times of the OP Stack network upgrades. Every `*_time` field is the L2 block timestamp (Unix seconds, UTC) at which that
+fork activates; an absent field means it is not scheduled, and `0` means active from genesis.
+
+Activation times can be inherited superchain-wide from the superchain's `superchain.toml` when a chain sets
+`superchain_time`, and are then written into each chain's own config. See
+[Hardfork activation inheritance](../docs/hardfork-activation-inheritance.md) for the exact rules.
+
+Two optional `[hardforks]` fields are not forks themselves but toggle a corrective behavior tied to one:
+
+- **`pectra_blob_schedule_time`** (timestamp, optional) — the Pectra blob-schedule fix: the L1-origin timestamp until
+  which (exclusive) the L1 block info's blob base-fee calculation keeps using the pre-Prague (Cancun) blob parameters,
+  switching to the Prague (Pectra) parameters from then on. It is needed by chains that were running buggy node software
+  when Prague activated on their L1: it records the L1 timestamp at which they actually switched, so re-derivation
+  reproduces their canonical history. As a timestamp field it participates in superchain inheritance. See the
+  [Pectra blob schedule spec](https://specs.optimism.io/protocol/pectra-blob-schedule/derivation.html) for details.
+
+- **`keep_karst_upgrade_gas`** (bool, optional, default `false`) — opts a chain out of the Karst upgrade-gas fix.
+  Karst's activation block adds one-time "upgrade gas" to the block gas limit so the network-upgrade transactions fit,
+  and the consensus-layer software is meant to subtract it again in the post-activation block so the limit reverts to
+  its steady-state value. Some chains activated Karst with a bug in that software that missed the subtraction, leaving
+  them with a permanently inflated gas limit; they set this to `true` so the fixed software keeps reproducing that
+  inflated limit and their history stays valid (the operator can lower it later with a `setGasLimit` transaction).
+  Unlike the timestamp fields, this flag is set per-chain and is never inherited from the superchain.
+
 ## Compression
 
 OP Stack genesis files are typically around 9MB, which quickly bloats the binary size of applications that use the
--- a/superchain/configs/sepolia/lisk.toml
+++ b/superchain/configs/sepolia/lisk.toml
@@ -23,6 +23,7 @@ max_sequencer_drift = 600
   isthmus_time = 1744905600 # Thu 17 Apr 2025 16:00:00 UTC
   jovian_time = 1763568001 # Wed 19 Nov 2025 16:00:01 UTC
   karst_time = 1781712001 # Wed 17 Jun 2026 16:00:01 UTC
+  keep_karst_upgrade_gas = true
 
 [optimism]
   eip1559_elasticity = 10
--- a/ops/internal/config/chain.go
+++ b/ops/internal/config/chain.go
@@ -116,9 +116,10 @@ type Hardforks struct {
 	HoloceneTime           *HardforkTime `toml:"holocene_time"`
 	PectraBlobScheduleTime *HardforkTime `toml:"pectra_blob_schedule_time,omitempty"`
 	IsthmusTime            *HardforkTime `toml:"isthmus_time"`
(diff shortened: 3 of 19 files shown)


superchain/configs/mainnet/superchain.toml
```diff
@@ -13,6 +13,7 @@ holocene_time = 1736445601 # Thu 9 Jan 2025 18:00:01 UTC
 isthmus_time = 1746806401 # Fri 9 May 2025 16:00:01 UTC
 jovian_time = 1764691201 # Tue 2 Dec 2025 16:00:01 UTC
 karst_time = 1783526401 # Wed 8 Jul 2026 16:00:01 UTC
+keep_karst_upgrade_gas = true
```
geoknee: If we add a new chain in the future, will it inherit this by default? Wondering if it would be better to just insert manually into each affected chain (the sepolia ones that already activated, plus any mainnet ones we want to have that behaviour).

sebastianst: that makes sense, will remove from `superchain.toml` and instead add to the relevant individual chains: OP, Ink, Soneium, Mode, Metal, Zora, and Unichain (both, sepolia and mainnet).

sebastianst: done

geoknee (approved):
Approving with one nit: would be nice to leave some explanation or paper trail explaining what this bool is about. The PR description is good, but not an obvious place to look to someone in the future trying to understand this. 

Can be addressed in a followup.

karlfloersch (approved):

superchain/configs/sepolia/creator-chain-testnet.toml
```diff
@@ -23,6 +23,7 @@ max_sequencer_drift = 600
   isthmus_time = 1744905600 # Thu 17 Apr 2025 16:00:00 UTC
   jovian_time = 1763568001 # Wed 19 Nov 2025 16:00:01 UTC
   karst_time = 1781712001 # Wed 17 Jun 2026 16:00:01 UTC
+  keep_karst_upgrade_gas = true
```
sbvegan: the security council does not hold the creator-chain-testnet keys, so we should omit this one

superchain/configs/sepolia/ethernity.toml
```diff
@@ -23,6 +23,7 @@ max_sequencer_drift = 600
   isthmus_time = 1744905600 # Thu 17 Apr 2025 16:00:00 UTC
   jovian_time = 1763568001 # Wed 19 Nov 2025 16:00:01 UTC
   karst_time = 1781712001 # Wed 17 Jun 2026 16:00:01 UTC
+  keep_karst_upgrade_gas = true
```
sbvegan: the security council does not hold these keys, so we should omit this one

superchain/configs/sepolia/lisk.toml
```diff
@@ -23,6 +23,7 @@ max_sequencer_drift = 600
   isthmus_time = 1744905600 # Thu 17 Apr 2025 16:00:00 UTC
   jovian_time = 1763568001 # Wed 19 Nov 2025 16:00:01 UTC
   karst_time = 1781712001 # Wed 17 Jun 2026 16:00:01 UTC
+  keep_karst_upgrade_gas = true
```
sbvegan, line 26: the security council does not hold these keys, so we should omit this one

superchain/README.md
```diff
@@ -1,3 +1,30 @@
+## Hardforks
+
+Each chain config (`superchain/configs/<superchain>/<chain>.toml`) carries a `[hardforks]` table with the activation
+times of the OP Stack network upgrades. Every `*_time` field is the L2 block timestamp (Unix seconds, UTC) at which that
+fork activates; an absent field means it is not scheduled, and `0` means active from genesis.
+
+Activation times can be inherited superchain-wide from the superchain's `superchain.toml` when a chain sets
+`superchain_time`, and are then written into each chain's own config. See
+[Hardfork activation inheritance](../docs/hardfork-activation-inheritance.md) for the exact rules.
```
sbvegan, line 9: can we also update the hardfork activation inheritance behavior document? Specifically adding the following condition:

- The chain must have its ProxyAdmin owner roles managed by Optimism to ensure any necessary smart contract upgrades that need to be applied prior to the hard fork are applied. 

Then we should remove the superchain_time from chains that are not: OP, Ink, Unichain, Soneium, Mode, Metal, or Zora

sbvegan:
Specifically we should only set the hard fork times for OP, Ink, Unichain, Soneium, Mode, Metal, and Zora because those are chains where their ProxyAdmin owner is managed by Optimism. Other chains we don't have any control over upgrading so I don't want a hard fork to activate prior to a necessary smart contract upgrade

superchain/README.md
sebastianst: This seems out of scope of this PR.

superchain/configs/sepolia/ethernity.toml
sebastianst: Removed this chain with https://github.com/ethereum-optimism/superchain-registry/pull/1263

superchain/configs/sepolia/lisk.toml
sebastianst: We decided to keep the flag here because Lisk did evidently activate the bug.

superchain/configs/sepolia/creator-chain-testnet.toml
sebastianst: Removed this chain with https://github.com/ethereum-optimism/superchain-registry/pull/1263

PR #1265, opened 2026-06-24
feat: document and enforce config field lifecycle (immutable/append-only/mutable)

geoknee:
## What

Chain config TOMLs mix values that are **fixed at genesis and describe the chain itself** (e.g. `chain_id`, `block_time`, the `[genesis]` block) with values that **evolve over time** (hardfork activations, role rotations, contract upgrades). Nothing told a reader — or CI — which is which. This PR makes the distinction explicit and adds an **advisory** CI check.

## How

Three layers, all driven by one source of truth:

1. **Struct tags + doc comment** (`ops/internal/config/chain.go`) — every `Chain` field carries a `lifecycle:"immutable|append-only|mutable"` tag, plus guidance on classifying new fields. This is the source of truth.
2. **Human-readable contract** (`superchain/configs/README.md`) — a new "Field lifecycle" section with a table and examples.
3. **Advisory CI check** — `config.CheckImmutableFields` compares old vs new configs based on the tags. `ops/cmd/check_immutable_fields` runs it against every changed `superchain/configs/*.toml` (diffing against the merge-base, mirroring the codegen check) and **warns**, via the `check-immutable-fields` CircleCI job.

## The lifecycle rules

| Lifecycle | Rule | Examples |
| --- | --- | --- |
| **immutable** | Identifies the chain, OR is a consensus param whose source of truth is L1 (registry only mirrors it; no auto-sync or change process today, so an edit is almost always accidental) | `chain_id`, `batch_inbox_addr`, `block_time`, `seq_window_size`, `max_sequencer_drift`, `gas_paying_token`, `data_availability_type`, `[optimism]` (EIP-1559 params), `[interop]` deps, `[genesis]` |
| **append-only** | New entries may be added; a **not-yet-active** activation may be re-scheduled, but one **already in the past** is frozen | `[hardforks]` |
| **mutable** | May change freely | `[roles]`, `[addresses]`, RPCs, `superchain_level`, … |

## The check is advisory, not blocking

The `check-immutable-fields` job is **intentionally not a required status check**. A change to a frozen field surfaces a **warning** (a red X a reviewer can see and acknowledge), not a hard failure — so legitimate clean-ups are never blocked. This resolves the tension raised in review: some "immutable" fields do occasionally need a genuine correction (a value committed wrong), and that shouldn't require reclassifying the field as mutable. Classify by what a field *represents*, not by whether it has ever been edited.

> Note: making this a non-required check requires dropping `ci/circleci: check-immutable-fields` from the branch-protection required-checks list — a repo-settings change, not something in this PR.

The append-only rule is **time-aware**: hardfork activations are routinely pushed out while still in the future (e.g. `jovian_time` was rescheduled twice via merged PRs); a naive "freeze any set value" rule would have falsely rejected those. Past activations are on-chain history and stay frozen. The check also handles the two non-standard `[hardforks]` entries: the optional `pectra_blob_schedule_time` (a timestamp, aged like any activation) and `keep_karst_upgrade_gas` (a bool that records how Karst activated — governed by `karst_time`, so adjustable until Karst is in the past and frozen thereafter). `TestHardforksFieldsClassified` fails if a new `[hardforks]` field is added without classifying how it ages.

## Testing

- Unit tests cover each lifecycle, the past/future hardfork distinction, `batch_inbox_addr`/`data_availability_type` immutability, the `keep_karst_upgrade_gas` modifier gated by `karst_time`, the optional `pectra_blob_schedule_time`, and guards that fail if a new `Chain` or `[hardforks]` field is added without classification.
- The `check_immutable_fields` tool was verified end-to-end: changing a past hardfork or an immutable field prints a clear warning and exits non-zero; a clean tree exits 0.
- `go build ./...`, `go vet ./...`, `go test ./internal/config/... ./internal/manage/...`, `gofmt` all clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ops/internal/config/chain.go
```diff
@@ -65,28 +65,44 @@ type Interop struct {
...
-	Roles                Roles               `toml:"roles"`
-	Addresses            Addresses           `toml:"addresses"`
+	Name                 string              `toml:"name" lifecycle:"mutable"`
+	PublicRPC            string              `toml:"public_rpc" lifecycle:"mutable"`
+	SequencerRPC         string              `toml:"sequencer_rpc" lifecycle:"mutable"`
+	Explorer             string              `toml:"explorer" lifecycle:"mutable"`
+	SuperchainLevel      SuperchainLevel     `toml:"superchain_level" lifecycle:"mutable"`
+	GovernedByOptimism   bool                `toml:"governed_by_optimism" lifecycle:"mutable"`
+	SuperchainTime       *uint64             `toml:"superchain_time" lifecycle:"mutable"`
+	DataAvailabilityType string              `toml:"data_availability_type" lifecycle:"mutable"`
```
sebastianst: is this really mutable?

ops/internal/config/chain.go
```diff
@@ -65,28 +65,44 @@ type Interop struct {
...
+	ChainID              uint64              `toml:"chain_id" lifecycle:"immutable"`
+	BatchInboxAddr       *ChecksummedAddress `toml:"batch_inbox_addr" lifecycle:"mutable"`
```
sebastianst: Is it really mutable? I think only the _batcher sender_ is mutable.

ops/internal/config/lifecycle.go
```diff
@@ -0,0 +1,140 @@
...
+// checkAppendOnly verifies the append-only contract for each entry of a struct field
+// (e.g. the activation times in [Hardforks]). For each entry that was already set in
+// the old version:
+//   - a hardfork activation whose time is at or before now is frozen and must be
+//     unchanged (it has activated on-chain and is historical fact);
+//   - a hardfork activation still in the future may be adjusted or removed;
+//   - any other already-set entry type is frozen (strict append-only).
+//
+// Entries that were unset (nil/zero) in the old version may be freely added.
+func checkAppendOnly(parent string, oldV, newV reflect.Value, now uint64) []string {
```
sebastianst, line 100: need to make sure this properly handles the optional fields for the blob schedule fix (timestamp) and `keep_karst_upgrade_gas` (bool).

ops/internal/config/chain.go
geoknee: I believe so. `data_availability_type` has legitimately changed in repo history: PR #646 ("Fix alt da chains Lyra and Orderly") flipped lyra from `eth-da` → `alt-da`. If we classified it immutable, that PR would have been rejected by this check. A chain can also migrate its DA layer operationally without becoming a different chain, so `mutable` is the safe (permissive) classification here. Keeping it as-is — but happy to reconsider if you think DA migrations should be gated some other way.

ops/internal/config/chain.go
geoknee: Good catch — agreed. Changed `batch_inbox_addr` to `immutable` in 0af2ad5. It has never changed for any chain across the registry's full history, and it isn't part of the on-chain `SystemConfig` (only the batcher *sender* / `BatchSubmitter` role is mutable, which is in `[roles]` and stays `mutable`). Updated the README table and added a `batch_inbox_addr is immutable` test case.

ops/internal/config/lifecycle.go
geoknee: Handled both in 0af2ad5 (and rebased onto main so `keep_karst_upgrade_gas` is actually present):

- **`pectra_blob_schedule_time` (timestamp):** it's a `*HardforkTime`, so it already flows through the same time-aware activation logic — unset may be added, a future value may be re-scheduled, a past value is frozen. Added `TestFreezePastBlobSchedule` + a table case to prove it.
- **`keep_karst_upgrade_gas` (bool):** this one needed real work. A bool has no "unset" state distinct from `false`, so freezing on the field's own value was wrong. It records *how Karst activated*, so `checkAppendOnly` now gates it on `karst_time`: freely settable/clearable while Karst is still in the future (or unset), frozen once Karst is in the past. See `hardforkModifiers` (maps the flag → its governing hardfork) and `TestKeepKarstUpgradeGas`.

Also added `TestHardforksFieldsClassified`, which fails if a future `[hardforks]` field is neither a `*HardforkTime` nor a registered modifier — so we can't silently add another optional field without deciding how it ages. Unclassified *changes* also fail closed at runtime.

ops/internal/config/chain.go
geoknee: That was Claude by the way. 

I reckon that PR was more "we committed the wrong information and fixed it" rather than reflecting a change the chain actually went through. We'll need to figure out how to deal with that kind of flow.

And yes this should probably be immutable.

geoknee: We should probably make the check added on this PR informational. So either a CI check which is not required, or a post to the PR thread warning of a change. That way, clean ups are not blocked and we just get a warning.

geoknee: Both done in ed2b459:

1. **`data_availability_type` → `immutable`.** Agreed the lyra change was "committed wrong info, then fixed" rather than a real DA migration. Now that the check is advisory (below), we classify by true semantics and such a correction just surfaces a warning.

2. **The check is now advisory, not blocking.** Went with the non-required-CI-check option you suggested: the git-aware enforcement moved out of the go test suite into a dedicated `ops/cmd/check_immutable_fields` tool, run by a new `check-immutable-fields` CircleCI job. It warns and exits non-zero (visible red X listing the fields), but is intentionally **not** a required status check, so clean-ups aren't blocked — a reviewer just acknowledges the warning.

One thing that needs doing outside this PR: dropping `ci/circleci: check-immutable-fields` from the branch-protection required-checks list, so the red X doesn't gate merges. Flagged that in the PR description too.

ops/internal/config/chain.go
```diff
@@ -65,28 +65,47 @@ type Interop struct {
...
+	DataAvailabilityType string              `toml:"data_availability_type" lifecycle:"immutable"`
+	ChainID              uint64              `toml:"chain_id" lifecycle:"immutable"`
+	BatchInboxAddr       *ChecksummedAddress `toml:"batch_inbox_addr" lifecycle:"immutable"`
+	BlockTime            uint64              `toml:"block_time" lifecycle:"immutable"`
+	SeqWindowSize        uint64              `toml:"seq_window_size" lifecycle:"immutable"`
+	MaxSequencerDrift    uint64              `toml:"max_sequencer_drift" lifecycle:"immutable"`
+	GasPayingToken       *ChecksummedAddress `toml:"gas_paying_token,omitempty" lifecycle:"immutable"`
+	Hardforks            Hardforks           `toml:"hardforks" lifecycle:"append-only"`
+	Interop              *Interop            `toml:"interop,omitempty" lifecycle:"mutable"`
+	Optimism             Optimism            `toml:"optimism" lifecycle:"mutable"`
```
geoknee: Does `Interop` belong here? And isn't `Optimism` actually referring to the genesis parameters which cannot change? I'm pretty sure these params are modified via the system config and _not_ by the superchain registry. All changes to consensus parameters or rules need to be synchronised across the network, so either an L1 event triggers the change or it is activated at a set timestamp (like a hardfork).

geoknee: Good question — I checked, and the answer is **no**, which sharpens the model:

- The only writer of the source `superchain/configs/*.toml` files is `WriteChainConfig`, and it's called **only** by the add-chain staging flow (`sync_staging`) — not by any periodic job.
- The daily codegen job writes only **derived** artifacts (`addresses.json`, `chainList.json`/`.toml`, `CHAINS.md`). It does **not** sync `[optimism]` — nor `[addresses]`/`[roles]` — back into the source configs.

So today `[optimism]` changes in the source only via a hand edit. That's exactly the accidental-edit case we want to warn on → `immutable` for now.

Your distinction is spot on and I've written it into the docs: for these fields the SCR is **downstream** of L1 — L1 (the SystemConfig) is the source of truth and the registry only mirrors it — **unlike hardfork times, where the SCR itself is the source of truth** and drives the activation (hence those are append-only, not immutable). If we later add a sync job or an intentional-change process, these would become "mutable by that job," downstream of the real consensus change — but we're not there yet. (85a6182)

